PR #42, opened 2015-01-05
ENH : Added quotes for __str__ method on EpicsMotor

stuwilkins:
Added quotes round string in __str__ method so this can be used to create the object.

### DIFF
--- a/ophyd/controls/positioner.py
+++ b/ophyd/controls/positioner.py
@@ -620,21 +620,21 @@ class PVPositioner(Positioner):
         return tuple(self._limits)
 
     def __repr__(self):
-        repr = 'setpoint={0._setpoint.read_pvname!r}'.format(self)
+        repr = ['setpoint={0._setpoint.read_pvname!r}'.format(self)]
         if self._readback:
-            repr = '{0}, readback={1._readback.read_pvname!r}'.format(repr, self)
+            repr.append('readback={0._readback.read_pvname!r}'.format(self))
         if self._actuate:
-            repr = '{0}, act={1._actuate.read_pvname!r}, ' \
-                'act_val={1._act_val!r}'.format(repr, self)
+            repr.append('act={0._actuate.read_pvname!r}'.format(self))
+            repr.append('act_val={0._act_val!r}'.format(self))
         if self._stop:
-            repr = '{0}, stop={1._stop.read_pvname!r}, ' \
-                'stop_val={1._stop_val!r}'.format(repr, self)
+            repr.append('stop={0._stop.read_pvname!r}'.format(self))
+            repr.append('stop_val={0._stop_val!r}'.format(self))
         if self._done:
-            repr = '{0}, done={1._done.read_pvname!r}, ' \
-                'done_val={1._done_val!r}'.format(repr, self)
-        repr = '{0}, put_complete={1._put_complete!r}'.format(repr, self)
-        repr = '{0}, settle_time={1._settle_time!r}'.format(repr, self)
-        repr = '{0}, limits={1._limits!r}'.format(repr, self)
-        repr = '{0}, name={1.name!r}'.format(repr, self)
+            repr.append('done={0._done.read_pvname!r}'.format(self))
+            repr.append('done_val={0._done_val!r}'.format(self))
+        repr.append('put_complete={0._put_complete!r}'.format(self))
+        repr.append('settle_time={0._settle_time!r}'.format(self))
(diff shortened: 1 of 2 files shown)


stuwilkins:
I just found something else which I think comes from pyepics. The EpicsSignal comes back as:

<pre>
EpicsSignal(value=296.68, read_pv=<PV 'XF:23ID1-ES{TCtrl:1-Chan:A}T-I', count=1, type=time_double, access=read/write>, write_pv=None)
</pre>

the <PV ...> is not a real repr. should we subclass PV from pyepics and overwrite that?

stuwilkins:
Sorry problem with html parsing

`EpicsSignal(value=296.68, read_pv=<PV 'XF:23ID1-ES{TCtrl:1-Chan:A}T-I', count=1, type=time_double, access=read/write>, write_pv=None)`

klauer:
We could subclass it, or just change `epics.PV._fmtsca` (scalar format) and `_fmtarr` (array format) on library initialization. I assume your preference would be something like `PV(pvname='XF:23ID1-ES{TCtrl:1-Chan:A}T-I', ...)`? 

To be fair, our `repr`s aren't correct either since we can't recreate the object from the string. 

My issue with doing it properly is that real `repr`s can get terribly verbose and not show what you would really want as an ophyd user. Say you have an EpicsSignal, `sig`, and type just that in IPython session, you'll see its `repr`, unless you force it into a string. A proper `repr` implementation would only show the PV names that it uses, while `str` would show whatever information we deem useful (value, etc.). The real question is then: does the utility of recreating objects from `repr` strings outweigh this?

stuwilkins:
Hi Ken,

yes my preference would be `PV(pvname = …)` for the `__repr__`.

I though that if you have a `__str__` method then that is shown when the user prints or even just types it in IPython. So we can have verbose `__repr__` strings and nice `__str__` strings


S


On Jan 6, 2015, at 10:22 AM, K Lauer <notifications@github.com<mailto:notifications@github.com>> wrote:


We could subclass it, or just change epics.PV._fmtsca (scalar format) and _fmtarr (array format) on library initialization. I assume your preference would be something like PV(pvname='XF:23ID1-ES{TCtrl:1-Chan:A}T-I', ...)?

To be fair, our reprs aren't correct either since we can't recreate the object from the string.

My issue with doing it properly is that real reprs can get terribly verbose and not show what you would really want as an ophyd user. Say you have an EpicsSignal, sig, and type just that in IPython session, you'll see its repr, unless you force it into a string. A proper repr implementation would only show the PV names that it uses, while str would show whatever information we deem useful (value, etc.). The real question is then: does the utility of recreating objects from repr strings outweigh this?

klauer:
If we can find another reason to subclass `epics.PV`, let's go that route. As it is, I think just replacing those strings should suffice.

A quick check shows IPython and Python behave similarly with `repr`/`str`:

```python
In [1]: class A(object):
   ...:     def __repr__(self): return 'repr'
   ...:     def __str__(self): return 'str'
   ...:

In [2]: a = A()

In [3]: a
Out[3]: repr

In [4]: str(a)
Out[4]: 'str'
```

stuwilkins:
Mmm that is unfortunate it would be much better if the __str__ was returned at the ipython prompt. Do I have the understanding of `__str__` and `__repr__` backwards?

Perhaps we should just put in the `__repr__` for the PV the actual name as PV(‘pv_name’).

S

On Jan 6, 2015, at 10:42 AM, K Lauer <notifications@github.com<mailto:notifications@github.com>> wrote:


If we can find another reason to subclass epics.PV, let's go that route. As it is, I think just replacing those strings should suffice.

A quick check shows IPython and Python behave similarly with repr/str:

In [1]: class A(object):
   ...:     def __repr__(self): return 'repr'
   ...:     def __str__(self): return 'str'
   ...:

In [2]: a = A()

In [3]: a
Out[3]: repr

In [4]: str(a)
Out[4]: 'str'

klauer:
> Please do not pollute the object model by subclassing from pyepics classes. I still think composition is the right approach here…

No argument there.

> Mmm that is unfortunate it would be much better if the __str__ was returned at the ipython prompt. Do I have the understanding of `__str__` and `__repr__` backwards?

Check [here](https://docs.python.org/2/reference/datamodel.html#object.__repr__) and maybe even [here](http://satyajit.ranjeev.in/2012/03/14/python-repr-str.html)

tacaswell:
attn @danielballan regarding IPython display hooks.

PR #50, opened 2015-01-13
ENH: Added functionality to format signal groups and trigger scaler

stuwilkins:
@dchabot, @klauer 

This is a suggested enhancement to run_engine and signal groups. Here we set the signal names in the group to have the base signal name and a postfix for the type of signal (for example `'_s1'` for S1 of the scaler). 

This then also adds signal groups to the event descriptors and events such that the signals are all added. I think this makes more intuitive sense to "bundle" up signals into groups which are then added. For example all the signals for a temperature controller could be bundled in a signal group.

Thoughts.... I tested this and it works OK with the scaler. 

I also added the `sclr.count` as the trigger so this can be triggered in the scan.

### DIFF
--- a/ophyd/controls/scaler.py
+++ b/ophyd/controls/scaler.py
@@ -32,13 +32,17 @@ class EpicsScaler(SignalGroup):
         Eventually need to provide PR1..16 -- preset counts too.
         '''
         signals = [EpicsSignal(self.field_pv('CNT'),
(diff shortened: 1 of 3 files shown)


stuwilkins:
Closed this after seeing @dchabot's comment, I will use this for testing at the moment in my branch then nuke the changes.

PR #57, opened 2015-02-10
Mds integration

ericdill:


ophyd/runengine/runengine.py
```diff
@@ -8,20 +10,49 @@
...
+        List of ophyd detectors, optional
+    data : dict
+        Dictionary of actual data
+    """
+    pv_info = {pos.name: {'source': pos.report['pv'],
+                          'dtype': 'number',
+                          'shape': None}
+               for pos in positioners}
+
+    def get_det_info(detector):
```
tacaswell: why do this as a nested function?

ophyd/runengine/runengine.py
```diff
@@ -8,20 +10,49 @@
...
+        """Internal function to grab info from a detector
+        """
+        val = data[detector.name]
+        try:
+            shape = val.shape
+        except AttributeError:
+            # val is probably a float...
+            shape = None
+        source = "PV:{}".format(detector.pvname)
+        return {detector.name: {'source': source, 'dtype': 'number',
```
tacaswell: this should be 'array' if `shape` is not `None`

ophyd/runengine/runengine.py
tacaswell: nm, see the closure.

ophyd/runengine/runengine.py
```diff
@@ -120,26 +150,33 @@ def _move_positioners(self, positioners=None, settle_time=None, **kwargs):
...
             time.sleep(settle_time)
-        # return {pos.name: pos.position for pos in positioners}
-        ret = {}
-        [ret.update({pos.name: pos.position}) for pos in positioners]
 
-        return ret
+        # use metadatastore to format the events so that ophyd is insulated from
+        # metadatastore spec changes
+        return {
+            pos.name: {
```
tacaswell: do this and all the other like dictionaries need to be turned into tuples?

tacaswell: or do I not really understand the comment above this?

ophyd/runengine/runengine.py
```diff
@@ -158,22 +195,51 @@ def _start_scan(self, **kwargs):
...
+                        detvals.update({sig.name: {
+                            'timestamp': sig.timestamp[sig.pvname.index(sig.report['pv'])],
+                            'value': sig.value}})
                 else:
-                    detvals.update({det.name: det.value})
+                    detvals.update({
+                        det.name: {'timestamp': det.timestamp,
+                                   'value': det.value}})
             detvals.update(posvals)
+            detvals = mds.format_events(detvals)
```
tacaswell: Ah, I assume this is what the above comment was alluding to.

ophyd/runengine/runengine.py
```diff
@@ -158,22 +195,51 @@ def _start_scan(self, **kwargs):
...
-                                                 seq_no=seqno,
-                                                 data=detvals)
-            create_event(event)
-            seqno += 1
+            print('datapoint[{}]: {}'.format(seq_num, detvals))
+            # grab the current time as a timestamp that describes when the
+            # event data was bundled together
+            bundle_time = time.time()
+            # actually insert the event into metadataStore
+            try:
```
tacaswell: Wouldn't this be simpler to be

```python
if event_descritpor is None:
    event_descritpor = mds.whatever()

event = mds.insert_event(..)
```
so that the insertion code only is in there once?

ophyd/runengine/runengine.py
```diff
@@ -191,28 +257,46 @@ def _get_data_keys(self, **kwargs):
...
+        end_args
+        scan_args
+
+        Returns
+        -------
+        data : dict
+            {data_name: []}
+        """
+        # format the begin run event information
+        beamline_id = scan_args.get('beamline_id', None)
```
tacaswell: This will raise an exception in the cause of the default `scan_args` value.  I suggest either making it mandatory (my relatively uninformed preference) or adding the 

```python
if scan_args is None:
    scan_args = dict()
```

ophyd/runengine/runengine.py
```diff
@@ -191,28 +257,46 @@ def _get_data_keys(self, **kwargs):
...
+            beamline_config=beamline_config, scan_id=runid, custom=custom)
+
         keys = self._get_data_keys(**scan_args)
         data = {k: [] for k in keys}
-        # print('keys = %s'%keys)
-        # event_descriptor = {'a': 1, 'b':2}
-        event_descriptor = data_collection.create_event_descriptor(
-            run_header=header, event_type_id=1, data_keys=keys,
-            descriptor_name='Scan Foo')
         if scan_args is not None:
```
tacaswell: If this check is redundent, if it is `None` an exception was raised well above.

ericdill:
ahhhhhh damn... you merged some config stuff into master :(

PR #65, opened 2015-02-15
API: Fix mds api

ericdill:
- Required by NSLS-II/metadataStore#86


ericdill:
NOTE: DO NOT MERGE THIS BEFORE mds#86

ericdill:
@arkilic Why is that advantageous?

PR #66, opened 2015-02-16
API : change begin/end event names

tacaswell:
insert_end_run -> insert_run_end
insert_begin_run -> insert_run_start

sundry variable/arg name changes


tacaswell:
I am in the process of doing so.  This is the easiest way to get the code to a controls network computer from home ;)

tacaswell:
It seems to be working

PR #72, opened 2015-02-21
BUG : Fixed name of logbook in count

stuwilkins:
I should really learn to call things words I can spell.........


tacaswell:
As a said earlier, with git your typos are forever.

tacaswell:
Do any of the existing log books need to be migrated to the new name in conjunction with this merge?

stuwilkins:
No …. There is already a logbook with the correct name for scans … it is only “counts” that go to the incorrect logbook

On Feb 21, 2015, at 8:39 PM, Thomas A Caswell <notifications@github.com<mailto:notifications@github.com>> wrote:


Do any of the existing log books need to be migrated to the new name in conjunction with this merge?

Issue #68, opened 2015-02-17
Sort out scan_ids


ericdill:
The mechanism for producing scan_ids needs to be investigated.  As of now, the scan_id seems to reset to an old number on CSX.  The behavior should be that the scan_id defaults to something like 

```
   session_manager.scan_id = int(DataBroker[-1].scan_id) + 1
```

On ophyd startup go grab the last scan_id from `DataBroker`.   This scan_id can be overridden by the user at run time.

Reported by Claudio at CSX

arkilic:
I think we could do this within metadatastore via mongo daemon. Mongo provides bunch of aggregation methods that would do this much faster than us. We could use $max. This would return the maximum scan_id value given an owner or w/e query we are interested in. Much simpler than getting bunch of headers and unpacking to get to desired info.

dchabot:
> On Feb 17, 2015, at 4:45 PM, Eric Dill <notifications@github.com> wrote:
> 
> The mechanism for producing scan_ids needs to be investigated. As of now, the scan_id seems to reset to an old number on CSX. The behavior should be that the scan_id defaults to something like
> 
>    session_manager.scan_id = int(DataBroker[-1].scan_id) + 1
> 
> On ophyd startup go grab the last scan_id from DataBroker. This scan_id can be overridden by the user at run time.
> 
> Reported by Claudio at CSX

I don’t want that degree of coupling between ophyd and DataBroker.
Let’s investigate why ophyd is producing incorrect numbers and simply fix that…

People can and will change the scan_id as they see fit, regardless of where it ends up downstream…

stuwilkins:
+1 Daron. 

Just to be clear again: 

1) `scan_id` are not unique and there is no guarantee that they are. They are a convenience method to assign a number to scans and it is a *convinience* that they are incremented by `ophyd`. They are just a human readable concept. 
2) The use should be able to define the behavior, or even reset or advance the number depending on their needs.
3) I 100% Agree with Daron that we don't want this polluting between MDS and `ophyd`
4) We should fix the problem at source.
5) To my knowledge this has happened once, lets focus on things that really need fixing. 
6) This may be a non-issue, at one point we had the test stand putting log entries into the CSX logbook. That caused some confusion as the number is therefore out of sequence. 

This does have some implications for the new broker interface, there should probably be some way of defining a default search terms before using the indexing notation such as (in pseudo code):
```
db = DataBroker()
[...]

PR #75, opened 2015-02-24
BUG: better scan_id persistence.

dchabot:
Patch against #68.
Please test.

### DIFF
--- a/ophyd/session/sessionmgr.py
+++ b/ophyd/session/sessionmgr.py
@@ -243,10 +243,12 @@ class SessionManager(object):
            Then, persists the scan_id using IPython's "%store" magics.
         '''
         self['_scan_id'] += 1
-        return self['_scan_id']
+        return self.persist_var('_scan_id', value=self['_scan_id'])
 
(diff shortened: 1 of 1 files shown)


ericdill:
Seems pretty reasonable. Just to be annoying, what happens when you load up two ipython sessions on the same computer?  As it is currently implemented at CSX, the `scan_id` is read from the disk at IPython startup and written when IPython is closed.  With this model, two IPython ophyd sessions can never be in sync with their `scan_ids`.  Does this IPython `%magic` stuff allow those two sessions to be in sync with regards to the `scan_id`?

danielballan:
It's possible to refresh stored variables. That would avoid collisions from two IPython sessions.

http://ipython.org/ipython-doc/rel-0.12/config/extensions/storemagic.html

@dchabot if we can finalize this PR today we can include it in tomorrow's deployment and update of other beamlines.

stuwilkins:
What about hooking into the ipython prompt so that each time the prompt processes it does a store and refresh?

Best wishes,
Stuart

Issue #78, opened 2015-02-28
BUG: order of operations in DScan. 


dchabot:
`DScan.pre_scan( )` method adds the relative positions to the absolute positions *after* `Scan.check_paths( )` is called. Thus, `check_paths( )` is actually trying to verify if the *deltas* are within the `(low_limit, high_limit)` of a `Positioner`.

Clear as mud...?

Demonstrated (safely!!!) at xf23id1 with:

```
 dscan(pgm_energy, -15, 30, 40)
```

stuwilkins:
Whoops ..... Good catch!

Best wishes,
Stuart


On Feb 28, 2015, at 16:38, dchabot <notifications@github.com<mailto:notifications@github.com>> wrote:


DScan.pre_scan( ) method adds the relative positions to the absolute positions after Scan.check_paths( ) is called. Thus, check_paths( ) is actually trying to verify if the deltas are within the (low_limit, high_limit) of a Positioner.

Clear as mud...?

Demonstrated (safely!!!) at xf23id1 with:

 dscan(pgm_energy, -15, 30, 40)

dchabot:
> On Feb 28, 2015, at 4:39 PM, Stuart Wilkins <notifications@github.com> wrote:
> 
> Whoops ..... Good catch!

We can look at it as verification that the limit-checking logic is sound ;-)

>

dchabot:
Fixed with PR#80

PR #80, opened 2015-03-03
BUG : Fixed problem with limits in DScan (Issue #78)

stuwilkins:


stuwilkins:
Yes, just tested on the add_detetctors branch to the best I can.

Issue #74, opened 2015-02-24
BUG : Debug using print must go away


stuwilkins:
Looking through the current master branch there is a lot of debug using ``print()`` I think this should all be removed. We should use either the logging (after all that is what it is for!) or when a PR is created to master this should be stripped.

ericdill:
Are you talking about print statements in the `runengine`, the `userapi` or both?

stuwilkins:
All ….. I don’t think we should be using ``print()`` in production systems.

S
On Feb 24, 2015, at 10:12 AM, Eric Dill <notifications@github.com<mailto:notifications@github.com>> wrote:


Are you talking about print statements in the runengine, the userapi or both?

ericdill:
Are you suggesting that you dont want **any** print statements at all?  See this one for example: https://github.com/NSLS-II/ophyd/blob/master/ophyd/userapi/scan_api.py#L630

stuwilkins:
No as per the title … no debug prints. Statements to inform the user about what is going on (moving motors etc.) is OK.

S

On Feb 24, 2015, at 11:28 AM, Eric Dill <notifications@github.com<mailto:notifications@github.com>> wrote:


Are you suggesting that you dont want any print statements at all? See this one for example: https://github.com/NSLS-II/ophyd/blob/master/ophyd/userapi/scan_api.py#L630

gjwillms:
Can we discuss this further?  I'd like to see print() disappear, to be replaced by a logger.  The user should have the option to display minimal formatted information, e.g., [positioners] [user_detectors] [default_detectors] (time), at the INFO level and the current output would be provided at DEBUG.  As a terminal guy, there's way too much stuff getting dumped to my screen and I'd like to have the ability to control this, maybe even by providing additional formatting at custom logging levels.

tacaswell:
:+1: on proper use of logging

stuwilkins:
Agreed. The mechanism is in place, and we should use it and make sure it propagates up to the top layers. I agree more informative print is also needed on the beginning of a run…. scan_id etc.

Best wishes,
Stuart

gjwillms:
I'm going to work on this.  I'd apologize in advance, but I don't think I'll actually regret this action, or its consequences, and I hate to be dishonest.  Nevertheless, I'm sure it's a milestone of some sort...

PR #87, opened 2015-03-18
switch from 'print()' to logging in scanapi and runengine

gjwillms:
addresses concerns raised around issue #74

### DIFF
--- a/ophyd/runengine/runengine.py
+++ b/ophyd/runengine/runengine.py
@@ -7,10 +7,12 @@ import time
 from threading import Thread
 from Queue import Queue
 import numpy as np
+import collections
+
 from ..session import register_object
 from ..controls.signal import SignalGroup
-
 from metadatastore import api as mds
+
 
 # Data formatting helper function
 def _get_info(positioners=None, detectors=None, data=None):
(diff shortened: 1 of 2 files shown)


ophyd/runengine/runengine.py
```diff
@@ -218,19 +235,22 @@ def _start_scan(self, **kwargs):
...
+            self._sessionmgr._logger.info(
+                '{}'.format(self._demunge_values(detvals, names)))
             # grab the current time as a timestamp that describes when the
             # event data was bundled together
             bundle_time = time.time()
             # actually insert the event into metadataStore
             try:
-                print('\n\ninserting event {}\n------------------'.format(seq_num))
+                self._sessionmgr._logger.debug(
+                    'inserting event {}------------------'.format(seq_num))
```
tacaswell: Loggers are one case where using the old-style formatting still makes sense.  If you do

```python
logger.debug('%s foo bar %d", 'str', 5)
```

The formatting will only be done if the message is actually propagated out to a handler.

gjwillms:
A print() statement remains in runengine.Demuxer because I can't figure out what is done with a Demuxer inside a RunEngine object...

danielballan:
Superceded by #103

PR #97, opened 2015-04-02
ENH: Pass scan kws to run start document

ericdill:


ophyd/userapi/scan_api.py
```diff
@@ -17,7 +17,7 @@
 session_manager = get_session_manager()
 logger = session_manager._logger
 
-__all__ = ['AScan', 'DScan', 'Scan', 'Data', 'Count']
+__all__ = ['AScan', 'DScan', 'Data', 'Count']
```
tacaswell: Was this change intentional?

ericdill: Yes.

ericdill: sitting with @dchabot and @danielballan at XPD. dan and daron discussed it and I simply added the change to this PR

tacaswell: Can you add a note about the API change?

stuwilkins: Why are we making this change?

Best wishes,
Stuart

danielballan:
@stuwilkins Scan does not have interaction with Olog, and the only thing is does with trajectories is check that they are within the bounds of the positions. Kevin used it by mistake, and obviously wasted some time, so @dchabot, @ericdill, and I think it muddies the waters. But if it is serving a purpose we don't know about, it should stay.

@tacaswell Yes.

PR #107, opened 2015-04-08
Fix AD errors

danielballan:
Some overly aggressive changes by yours truly during the XPD deployment broke CSX. Those changes are here reverted. CSX detectors work now. I will need to verify that XPD works before this is merged / released.

In a separate commit, Stuart cleaned up and simplified his stats code.

As an incidental result of these changes, the error-level log messages issued by ophydobj.py have been resolved.

### DIFF
--- a/ophyd/controls/area_detector.py
+++ b/ophyd/controls/area_detector.py
@@ -88,7 +88,8 @@ class AreaDetector(SignalDetector):
                                       rw=False, recordable=True)
 
                 self.add_signal(sig, add_property=True)
-                sig.subscribe(self._stats_changed)
+
+                #sig.subscribe(self._stats_changed)
 
         self._shutter_val = shutter
         self._shutter_rb_val = shutter_rb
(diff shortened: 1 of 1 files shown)


ophyd/controls/area_detector.py
```diff
@@ -126,21 +127,10 @@ def _acquire_changed(self, value=None, old_value=None, **kwargs):
...
-            nstats = len(self._stats) * self._acq_num
-            if ((self._stats_counter == nstats) and
-                 (self._acq_count == self._acq_num)):
-                self._run_subs(sub_type=self._SUB_DONE)
-                self._reset_sub(self._SUB_DONE)
-        else:
-            if self._acq_count == self._acq_num:
-                self._run_subs(sub_type=self._SUB_DONE)
-                self._reset_sub(self._SUB_DONE)
+        if self._acq_count == self._acq_num:
```
tacaswell: This still uses `self._acq_num`.  I still think that we should ensure that it is created in `__init__`.  Just because you _can_ dynamically add attributes to an object does not mean you _should_.

PR #111, opened 2015-04-09
FIX: TIFF should use multiple mode.

danielballan:
* Acquisition mode should have been multiple. This was a bug.
* To be safe, ophyd enables the TIFF plugin on configure and disables on deconfigure.
* AutoSave must be on.


tacaswell:
@stuwilkins Looking at this again, shouldn't the enable/disable logic live a lot higher in the class tree?

stuwilkins:
No, this can be quite specific. I would prefer to keep it as a per plugin need.

S

PR #116, opened 2015-04-16
Experimental: ophyd callbacks for built-in or user-defined plotting

danielballan:
This PR adds callbacks to the scan, allowing the user to register functions at the start and stop and after the insertion of each event and event descriptor. I think this is more or less in line with @dchabot's general plans for the flow of ophyd.

There are two callback registries here, a private one that is processed by the Scan thread and a public one that is processed by the main thread. The private callbacks add new data to a Queue, and that Queue is processed by the public callbacks in the main thread.

Here, I use those callbacks to generate live-updating matplotlib figures: one figure with a subplot for each scalar detector, and a separate figure for each image detector. The registries are set up on a per-scan basis, so beamlines can set up special live plotting feedback for their special scans.

*This simple implementation is not a good general solution.* Although matplotlib and the scan are running on separate threads, they are on the same process. If matplotlib cannot do its work in whatever time the scan spends sleeping, it will slow down the scan. In the future, these callbacks could be used to communicate with a separate plotting *process*. All of the matplotlib code is isolate can be neatly excised later.

### DIFF
--- a/ophyd/userapi/scan_api.py
+++ b/ophyd/userapi/scan_api.py
@@ -4,6 +4,7 @@ import matplotlib.pyplot as plt
 from time import sleep
 import sys
 import collections
+from Queue import Queue
 import itertools
 import string
 import traceback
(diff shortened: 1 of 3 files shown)


ophyd/userapi/scan_api.py
```diff
@@ -151,6 +158,16 @@ def __init__(self, *args, **kwargs):
...
+        self._scan_cb_registry = CallbackRegistry()  # process on scan thread
+        self.cb_registry = CallbackRegistry()  # process on main thread
+        self._plot_mgr = PlotManager()
+        self._autoplot = None
+        self.autoplot = True
+
+        self.ev_queue = Queue()
+        self.desc_queue = Queue()
+        self._register_scan_callback('event', self._push_to_ev_queue)
+        self._register_scan_callback('descriptor', self._push_to_desc_queue)
```
dchabot: Don't forget to push to start and stop queues.

Issue #109, opened 2015-04-09
scans should die more gracefully with no detectors or positioners


tacaswell:
running `ct` with no detectors or positioners defined results in an exception from deep in the opyhd stack/metadatastore on a different thread which requires ctrl-c to kill.

dchabot:
Please provide a backtrace for this.

gjwillms:
I think that this is what Tom is referring to:
Ophyd [20]: ct.detectors
   Out[20]: []

Ophyd [21]: ct
----------> ct()
Stored '_scan_id' (int)
2015-04-16 12:55:12,530 [ophyd_session:INFO] Scan ID: 382
2015-04-16 12:55:12,530 [ophyd_session:INFO] Time: 2015-04-16T12:55:12.529094
2015-04-16 12:55:12,531 [ophyd_session:INFO] uid: 13509b88-010b-482c-beb8-f92e7eebf1cf
2015-04-16 12:55:12,531 [ophyd_session:INFO] Begin Run...
2015-04-16 12:55:12,531 [ophyd_session:INFO] 
2015-04-16 12:55:12,631 [ophyd_session:INFO] 
Exception in thread Scanner:
Traceback (most recent call last):
  File "/nfs/xf05id1/mc/envs/ophyd_ng/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/nfs/xf05id1/mc/envs/ophyd_ng/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "build/bdist.linux-x86_64/egg/ophyd/runengine/runengine.py", line 234, in _start_scan
    data_keys=mds.format_data_keys(data_key_info))
  File "build/bdist.linux-x86_64/egg/metadatastore/commands.py", line 28, in inner
    return func(*args, **kwargs)
  File "build/bdist.linux-x86_64/egg/metadatastore/commands.py", line 286, in insert_event_descriptor
    event_descriptor.save(validate=True, write_concern={"w": 1})
  File "build/bdist.linux-x86_64/egg/mongoengine/document.py", line 224, in save
    self.validate(clean=clean)
  File "build/bdist.linux-x86_64/egg/mongoengine/base/document.py", line 323, in validate
    raise ValidationError(message, errors=errors)
ValidationError: ValidationError (EventDescriptor:None) (Field is required and cannot be empty: ['data_keys'])

<*ophyd hangs here until a sigint is sent*>

[...]

PR #117, opened 2015-04-16
Patch against #109

dchabot:


ophyd/userapi/scan_api.py
```diff
@@ -229,11 +229,17 @@ def run(self, *args, **kwargs):
         The main loop of the scan. This routine runs the scan and calls the
         ophyd runengine.
         """
+
+        # Raise if no detectors are associated with a scan instance,
+        # and do it prior to context manager __entry__
+        if not len(self.detectors) > 0:
```
tacaswell: python nitpick

```python
if not self.detectors:
    raise ...
```

Has the same behaviour.  I think it is slightly more readable, but not a huge issue either way.

dchabot: Thanks for the tip.

PR #119, opened 2015-04-22
ENH: RunEngine mockup with Threads and Events

ericdill:
This should allow me/anyone_else to fool around with the RunEngine and callbacks and stuff without needing the full epics stack and a running IOC


ericdill:
This was a useful exercise in that I learned more about python threading, but this is ultimately not helpful for Ophyd moving forward.

PR #122, opened 2015-05-13
Documentation

danielballan:
Since there was nothing at [nsls-ii.github.io/ophyd](http://nsls-ii.github.io/ophyd/index.html), I went ahead and published them there, where they will be easier to review and read.

I am sure @klauer, @dchabot, and @stuwilkins  could have explained better. Don't hold back!


klauer:
We should revisit this and get automatic documentation generation going again. The new `Device` model also makes having class inheritance diagrams much more important.

PR #124, opened 2015-05-28
WIP: Dispatcher, a framework for callbacks that consume Documents

danielballan:
As per a discussion with @dchabot, this PR removes the metadatastore dependency from ophyd, making it a generic "consumer" of Documents created during the scan. There are two kinds of consumers:

1. Consumers that consume documents on the Scan Thread, blocking the scan's progress until it has returned successfully.
2. Consumers that consume documents on the Main Thread, reading them out of a queue. This is not as  reliable, but it does not lock up the scan.

Only one consumer is hard-coded: a consumer that consumes Documents on the Scan Thread and pushes them into the queue, where they can be read out by consumers on the Main Thread.

Metadatastore should be registered a Scan Thread consumer in the config file, but it no longer a hard dependency. User code (e.g., live analysis or visualization) should be registered as a Main Thread consumer. Notably, this user code does not have to read the data out of Data Broker; it can get it directly from ophyd.

This PR is feature-complete, but needs to be tested and of course reviewed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,9 @@ target/
 # PyCharm
 .idea/
 .ropeproject
+
+# notebook checkpoints
+.ipynb_checkpoints
+
+# generated docs
+doc/source/generated
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -36,7 +36,6 @@ requirements:
     - python
     - pyepics
     - pcaspy
-    - metadatastore
     - six
     - ipython
     - pyolog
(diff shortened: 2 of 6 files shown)


tacaswell:
Can you nuke the ipython checkpoints and add that to the ignores? 

:+1: on force pushing to expunge

danielballan:
@tacaswell Done.

danielballan:
@dchabot I tested this on a couple invocations of ascan() and ct(). I'm sure we'll find *something* wrong, but I think this is good to merge from my end.

danielballan:
And, notably, this works just in time for *happy hour*. :beers:

PR #127, opened 2015-06-11
BUG: scans using positioners failed

klauer:
`positioners` was in `kwargs`, which is passed to `_move_positioners` directly. Removing it caused `positioners` to be `None` in `_move_positioners`.

(ref 4fbbecf6f3b991787dbd0bb7aa87abdd922c7f09)

### DIFF
--- a/ophyd/runengine/runengine.py
+++ b/ophyd/runengine/runengine.py
@@ -186,7 +186,7 @@ class RunEngine(object):
         while self._scan_state is True:
             self.logger.debug(
                 'self._scan_state is True in self._start_scan')
-            posvals = self._move_positioners(**kwargs)
+            posvals = self._move_positioners(positioners=positioners, **kwargs)
             self.logger.debug('moved positioners')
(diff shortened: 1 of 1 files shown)


ophyd/runengine/runengine.py
```diff
@@ -275,7 +275,7 @@ def _demunge_names(self, names):
         msg = ''.join('{}\t'.format(name) for name in unique_names)
         return msg
 
-    def _get_data_keys(self, positioners=None, detectors=None):
+    def _get_data_keys(self, positioners=None, detectors=None, **kwargs):
```
tacaswell: Why is kwargs back? it is not used.

klauer: Perhaps the HXN configuration is incorrect, but things like `settle_time` were getting passed along and causing that to break:

```python
TypeError: _get_data_keys() got an unexpected keyword argument 'settle_time'
```

A better solution, I suppose, would be to just give `_get_data_keys` only the arguments it expects.

PR #128, opened 2015-06-18
Allow truncation of EpicsSignal read() values

ericdill:
This PR adds two kwargs to the instantiation of the EpicsSignal
class: `dtype` and `num_decimals`.
- `dtype` can be a float, int or string
- `num_decimals` is the decimal position to round the return value
  to. Negative numbers will round to that many digits before the 
  '.' and positive numbers will round to that many digits after
  the '.'.

Use case: The encoder for a motor (e.g., theta) is known
to be sensitive to the millidegree (0.001). As such, having the
read() function return a number like 1.031239867892734 is 
unnecessary, but also misleading. Instead, allow the beamline
scientist to denote that the read value should return a float with
three decimal points. For example:
```python
eta = EpicsMotor('XF:23ID1-ES{Diag:1-Ax:Eta}Mtr', name='eta',
                 dtype='float', num_decimals=3)
```
will return a float with 3 decimals when eta.read() is called.
eta.describe() will contain the extra field `num_decimals` that
shows how many decimals the number is being rounded to.


ericdill:
wow thought i rebased this.
derp

PR #129, opened 2015-06-18
Allow truncation of EpicsSignal read() values

ericdill:
This PR adds two kwargs to the instantiation of the EpicsSignal
class: `dtype` and `num_decimals`.
- `dtype` can be a float, int or string
- `num_decimals` is the decimal position to round the return value
  to. Negative numbers will round to that many digits before the 
  `.` and positive numbers will round to that many digits after
  the `.`.

Use case: The encoder for a motor (e.g., theta) is known
to be sensitive to the millidegree (0.001). As such, having the
read() function return a number like 1.031239867892734 is 
unnecessary, but also misleading. Instead, allow the beamline
scientist to denote that the read value should return a float with
three decimal points. For example:
```python
eta = EpicsMotor('XF:23ID1-ES{Diag:1-Ax:Eta}Mtr', name='eta',
                 dtype='float', num_decimals=3)
```
will return a float with 3 decimals when eta.read() is called.
eta.describe() will contain the extra field `num_decimals` that
shows how many decimals the number is being rounded to.

### DIFF
--- a/ophyd/controls/signal.py
+++ b/ophyd/controls/signal.py
@@ -177,29 +177,17 @@ class EpicsSignal(Signal):
         Check limits prior to writing value
     auto_monitor : bool, optional
         Use automonitor with epics.PV
-    dtype : type, optional
+    dtype : {float, int, string}
         Defaults to float.
-        This is the datatype that the value from epics will be returned as.
-        Note that this says nothing about the precision of the value, simply
-        what its type should be.
-        Can be any valid python or numpy type. Some examples to get you
-        started...
-        `complex`, `float`, `int`, `uint`, `byte`, `bool`, `string`
-        {'complex': [numpy.complex64, numpy.complex128, numpy.complex256],
-         'float': [numpy.float16, numpy.float32, numpy.float64, numpy.float128],
-         'int': [numpy.int8, numpy.int16, numpy.int32, numpy.int64],
-         'others': [bool, object, str, str, numpy.void],
-         'uint': [numpy.uint8, numpy.uint16, numpy.uint32, numpy.uint64]}
+        This is the type that read() will be return.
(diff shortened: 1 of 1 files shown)


ericdill:
requested by @cmazzoli

stuwilkins:
There is a more EPICS way to do this. There is a PREC field of most epics PVs which gives the precision set by the user. That gives the number of decimal places. That could be read by the EpicsSignal on creation and set the number of decimals appropriately. This may even be in there somewhere, I remember adding it for the display.

S

On Jun 18, 2015, at 9:27 AM, Eric Dill <notifications@github.com<mailto:notifications@github.com>> wrote:


requested by @cmazzoli<https://github.com/cmazzoli>

klauer:
Right - EpicsSignal has the precision property, which is the readback PV's reported precision.

ophyd/controls/signal.py
```diff
@@ -195,6 +207,17 @@ def __init__(self, read_pv, write_pv=None,
         self._pv_kw = pv_kw
         self._auto_monitor = auto_monitor
 
+        if dtype is None:
+            dtype = float
+        valid_dtypes = ['float', 'int', 'str']
```
tacaswell: This is the wrong list of valid entries for dtypes.  See http://nsls-ii.github.io/arch/metadatastore-format.html#schema

tacaswell:
The `EventDescriptor` documents are specified as having a fixed set of _required_ keys and then as many other keys as you want so if we can extract more information out of epics we probably should (iirc, spec will also tell you the engineering untis?).  Down-stream software can use the extra keys or not, but cannot require that those keys be there.

If we want to require extra keys we need to version bump the spec and should probably put that off for a while if possible.

PR #131, opened 2015-06-24
Enh squash stack

tacaswell:


ophyd/controls/area_detector.py
```diff
@@ -166,6 +166,9 @@ def configure(self, **kwargs):
 
         # If using the stats, configure the proc plugin
 
+        # Enable images to be saved
+        self._write_plugin('EnableCallbacks', 1, self._file_plugin)
```
tacaswell: This should move down to the `*Filestore` child class

ophyd/controls/area_detector.py
```diff
@@ -228,7 +231,7 @@ def _set_shutter(self, value):
     def _start_acquire(self, **kwargs):
         """Do an actual acquisiiton"""
         if self._acq_count < self._acq_num:
-            self._set_shutter(self._acq_count % 2)
+            self._set_shutter(self._acq_count % 2 + 1)
```
tacaswell: make sure this works at XPD and sort out a way to make this configurable so we do not break other beamlines.

ophyd/controls/area_detector.py
```diff
@@ -667,10 +669,6 @@ def __init__(self, *args, **kwargs):
         self.add_signal(self._ad_signal('FilePathExists', '_filepath_exists',
                                         self._file_plugin,
                                         recordable=False))
-
-        # Acquisition mode (single image)
-        self._image_acq_mode = 0
```
tacaswell: Make sure this does not break any other beamlines.

tacaswell: @stuwilkins what does this do?

dchabot: That may have been an unintentional removal?
At any rate, removing that line means an areadetector instance will be left in whatever mode it was in prior to the object's instantiation (single, multiple, continuous = (0,1,2)). IOW, not instantiated into single-image mode.

danielballan:
Related to #158?

danielballan:
Superceded by #163

PR #134, opened 2015-06-25
MNT: More updates because mds breaks all the things

### LINKED ISSUE #132: metadatastore master breaks all the things (or maybe, one logging statement)
For example: `https://github.com/NSLS-II/ophyd/blob/master/ophyd/runengine/runengine.py#L333` will be broken because `https://github.com/NSLS-II/ophyd/blob/master/ophyd/runengine/runengine.py#L326` returns a uid!

ericdill:
Closes #132


ericdill:
see #143

PR #143, opened 2015-09-10
Turn on/off file plugin after before/scan

ericdill:


danielballan:
@klauer Will this cause you any trouble?

ericdill:
closes https://github.com/NSLS-II/ophyd/issues/133

klauer:
Probably not. You might have seen from some of my recent changes to the area_detector module that I like the ability to at least toggle functionality like this.

Sometimes non-optional features/settings/whatever are buried so deeply into the class hierarchy you have no choice but to re-implement everything surrounding it just to skip it in your subclass.

ericdill:
This needs to go in before the deployment to CHX

danielballan:
@klauer Yeah I agree that's a major flaw with the current design. I think we can do better in the medium term.

ericdill:
@danielballan should this go in before too much longer?

danielballan:
Merged so that we don't break CSX. A more customizable AreaDetector is high on the to-do list, but it's not going to make the cutoff for September 30.

Issue #136, opened 2015-07-02
Do not accept 'None' as a number


danielballan:
Raise or at least re-read when the ioc returns `None` for a value that is expected to be numeric.

Real-life example:

```
In [17]: theta.read()
Out[17]: {'theta': {'timestamp': None, 'value': None}}
```

klauer:
To clarify, EPICS IOCs don't have a concept of ```None``` - this is unfortunately how pyepics indicates an error/timeout. `epics.caget` works the same. For example:

```python
In [1]: pv = epics.PV('_INVALID_PV_NAME_')
In [2]: value = pv.get()
In [3]: print value
None
```

danielballan:
Right, this could be addressed at the pyepics level instead. For now, it seems easier to catch it in ophyd and raise an exception instead of saying "The motor position is None."

tacaswell:
:+1: on ophyd raising on a failed read.

tacaswell:
@dchabot Should this be marked as done due to #144 ?

dchabot:
Yes, #144 covers this.

Issue #90, opened 2015-03-20
When ophyd fails to connect and times out, it should raise, not return None.


danielballan:

tacaswell:
closed by #144

PR #144, opened 2015-09-18
[FIX] CA (dis)connection management

dchabot:
Addresses #136 and #90

Basically, avoids the assumption in a couple of `__init__`-ializers that a connection to the controls layer has been made.

This means reads/writes to disconnected channels will `raise`, and loading of startup configuration files will not stall if channels cannot be reached initially. Need to have a think about how best to inform users that only M out of N of their devices made a connection by the time the __In [1]:__ prompt lands. Suggestions welcome.

NOTE: this does _not_ include more graceful handling of swarms of (dis)connects.

### DIFF
--- a/ophyd/controls/positioner.py
+++ b/ophyd/controls/positioner.py
@@ -17,7 +17,7 @@ from epics.pv import fmt_time
 
 from .signal import (EpicsSignal, SignalGroup)
 from ..utils import TimeoutError
-from ..utils.epics_pvs import record_field
+from ..utils.epics_pvs import record_field, raise_if_disconnected
 from .ophydobj import MoveStatus
 
 logger = logging.getLogger(__name__)
@@ -176,6 +176,7 @@ class Positioner(SignalGroup):
         self._reset_sub(self._SUB_REQ_DONE)
 
     @property
+    @raise_if_disconnected
     def position(self):
         '''The current position of the motor in its engineering units
 
@@ -263,20 +264,24 @@ class EpicsMotor(Positioner):
         self._user_readback.subscribe(self._pos_changed)
 
     @property
+    @raise_if_disconnected
     def precision(self):
         '''The precision of the readback PV, as reported by EPICS'''
         return self._user_readback.precision
 
     @property
+    @raise_if_disconnected
     def egu(self):
         '''Engineering units'''
         return self._egu.value
 
     @property
+    @raise_if_disconnected
     def limits(self):
         return self._user_setpoint.limits
 
     @property
+    @raise_if_disconnected
     def moving(self):
         '''Whether or not the motor is moving
 
@@ -286,6 +291,7 @@ class EpicsMotor(Positioner):
         '''
         return bool(self._is_moving.get(use_monitor=False))
 
+    @raise_if_disconnected
     def stop(self):
         self._stop.put(1, wait=False)
         Positioner.stop(self)
@@ -299,6 +305,7 @@ class EpicsMotor(Positioner):
         '''Return a full PV from the field name'''
         return record_field(self._record, field)
 
+    @raise_if_disconnected
     def move(self, position, wait=True,
              **kwargs):
 
@@ -346,6 +353,7 @@ class EpicsMotor(Positioner):
             self._done_moving(timestamp=timestamp, value=value)
 
     @property
+    @raise_if_disconnected
     def report(self):
         return {self._name: self.position,
                 'pv': self._user_readback.pvname}
@@ -453,6 +461,7 @@ class PVPositioner(Positioner):
         self._setpoint.check_value(pos)
 
     @property
+    @raise_if_disconnected
     def moving(self):
         '''Whether or not the motor is moving
 
@@ -545,6 +554,7 @@ class PVPositioner(Positioner):
             self._setpoint.put(position, wait=False,
                                callback=done_moving)
 
+    @raise_if_disconnected
     def move(self, position, wait=True, **kwargs):
         if wait:
             try:
--- a/ophyd/controls/positioner.py
+++ b/ophyd/controls/positioner.py
@@ -259,11 +259,8 @@ class EpicsMotor(Positioner):
         for signal in signals:
             self.add_signal(signal)
 
-        self._moving = bool(self._is_moving.value)
         self._done_move.subscribe(self._move_changed)
         self._user_readback.subscribe(self._pos_changed)
-
-        self._set_position(self._user_readback.value)
 
     @property
     def precision(self):
@@ -291,7 +288,6 @@ class EpicsMotor(Positioner):
 
     def stop(self):
         self._stop.put(1, wait=False)
-
         Positioner.stop(self)
 
     @property
@@ -425,8 +421,6 @@ class PVPositioner(Positioner):
                                         name=self.name))
 
             self._readback.subscribe(self._pos_changed)
-
-            self._set_position(self._readback.value)
         else:
             self._setpoint.subscribe(self._pos_changed)
 
--- a/ophyd/controls/signal.py
+++ b/ophyd/controls/signal.py
@@ -15,7 +15,8 @@ import time
 import epics
 
 from ..utils import (ReadOnlyError, TimeoutError, LimitError)
-from ..utils.epics_pvs import (get_pv_form, waveform_to_string)
+from ..utils.epics_pvs import (get_pv_form,
+                               waveform_to_string, raise_if_disconnected)
 from .ophydobj import OphydObject, DetectorStatus
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


ericdill:
This is probably me just aggressively pattern matching, but it seem that you would also want to get rid of the other references to `self.moving` in the code block at lines 331-346 and replace some of them with `self.moving`?

ophyd/controls/signal.py
```diff
@@ -285,6 +285,13 @@ def _connected(self, pvname=None, conn=None, pv=None, **kwargs):
             self._ses_logger.debug(msg)
 
     @property
+    def connected(self):
+        if self._write_pv is None:
+            return self._read_pv.connected
+        else:
+            return self._read_pv.connected and self._write_pv.connected
```
ericdill: @tacaswell  has me well trained to replace if/else blocks. Consider replacing this if/else block with:

```python
if self._write_pv:
    return self._read_pv.connected and self._write_pv.connected
return self._read_pv.connected
```

ophyd/controls/signal.py
```diff
@@ -326,6 +333,11 @@ def get(self, as_string=None, **kwargs):
         if as_string is None:
             as_string = self._string
 
+        if not self._read_pv.connected:
+            if not self._read_pv.wait_for_connection():
```
ericdill: perhaps combine these two if statements into a single if statement

ophyd/session/sessionmgr.py
```diff
@@ -228,7 +228,7 @@ def stop_all(self):
             self._run_engine.stop()
 
         for pos in self._registry['positioners'].values():
-            if pos.moving is True:
+            if pos.connected and pos.moving is True:
```
ericdill: `if pos.connected and pos.moving is True:` is equivalent to `if pos.connected and pos.moving`, perhaps use the latter

ericdill:
looks good to me other than a few minor line comments, thanks @dchabot !

ophyd/session/sessionmgr.py
dchabot: Oops, that was sloppy. Will fix this (and others) later today. Busy this AM.

ophyd/controls/positioner.py
```diff
@@ -425,8 +421,6 @@ def __init__(self, setpoint, readback=None,
                                         name=self.name))
 
             self._readback.subscribe(self._pos_changed)
-
-            self._set_position(self._readback.value)
```
klauer: So in the case that the read-back PV has yet to connect, `motor.position` will return `None`. You should probably do a connectivity check in the `position` property.

klauer:
Nice work overall - this will significantly decrease @danielballan's frustration level :+1:

ophyd/controls/positioner.py
dchabot: Good catch. Do you think a plain `Exception` will do here?

klauer: If unconnected, I think it should try an explicit `readback.get()` to wait on the connection. That way it has a chance of getting the position as requested -- or it'll throw a `TimeoutError` with your new changes.

dchabot: Hmmm. A bit of subtlety here - in order to cover subclasses of `Positioner`, the solution should reside in `Positioner` Otherwise, `EpicsMotor`, `PVPositioner`, and `PseudoPositioner` will each require their own `position` property. That smells a bit funny to me.

ophyd/controls/signal.py
dchabot: Ok. I'll bite - why?

danielballan:
Hahaha, it's true.

ophyd/controls/signal.py
danielballan: I personally prefer it the way @dchabot wrote it. I find that more readable. Giant (multi-case) if/elif/elif/.../else blocks are an anti-pattern in Python -- maybe that is what @tacaswell was talking about?

ophyd/controls/signal.py
```diff
@@ -526,6 +538,10 @@ def add_signal(self, signal, prop_name=None):
             if prop_name:
                 setattr(self, prop_name, signal)
 
+    @property
+    def connected(self):
```
danielballan: I like this!

ophyd/controls/positioner.py
```diff
@@ -259,12 +259,9 @@ def __init__(self, record, settle_time=0.05, **kwargs):
         for signal in signals:
             self.add_signal(signal)
 
-        self._moving = bool(self._is_moving.value)
```
danielballan: @tacaswell Is this attribute involved in the asyncio status obj?

ophyd/controls/positioner.py
danielballan: Re: @klauer's initial comment about `motor.position` doing a connectivity check, is this a good application for a `ensure_connection` decorator, as @dchabot originally suggested in conversation?

klauer: @danielballan Does `ensure_connection` check to see if it's disconnected and raise an exception, or try to connect?

@dchabot I can see where that would seem a bit odd, but in reality the base class shouldn't know anything about the signals implemented in layers above it. So I don't think it's so strange to have to customize it for all subclasses.

danielballan: I'll let @dchabot clarify what he had in mind.

ophyd/controls/signal.py
danielballan: Apropo of code style, https://twitter.com/devsecops/status/644790108852498432

ophyd/controls/positioner.py
dchabot: I was planning to have `ensure_connection` simply `raise`. 

If a disconnect has occurred, then something is happening beyond the CA client's realm of control anyway. I feel that it's enough to bring the condition to the attention of the user as timely as possible, but if others lean strongly the other way then we explore fault tolerance options.

BTW - be in around lunch time. I'm on airport shuttle service this morning.

klauer: My concern is more for the case where the `EpicsMotor` has just been created but has yet to connect. We got around that previously by blocking on creation. This at least ensured the initial state.

A (more complex) solution for `ensure_connection` might check to see if the initial connection attempt has finished yet. If not, it blocks. In the case of a random disconnection, it raises immediately.

ophyd/controls/positioner.py
tacaswell: I don't think so, it as all done via callbacks now.

ophyd/controls/signal.py
tacaswell: It is mostly a style thing and matters more if you have patterns like

```python
if thing:
    return
else:
    lots
    of
    code
```

If there is a return in and `if` block, then anything under it is the `else` block

tacaswell:
? why does this need a rebase? gh thinks it merges cleanly (the non-existent tests just don't pass)

ophyd/utils/epics_pvs.py
```diff
@@ -325,3 +326,13 @@ def records_from_db(fn):
...
     return ret
+
+def raise_if_disconnected(fcn):
+    '''Decorator to catch attempted access to disconnected EPICS channels.'''
+    @functools.wraps(fcn)
+    def wrapper(self, *args, **kwargs):
+        if self.connected:
+            return fcn(self, *args, **kwargs)
+        else:
+            raise Exception('{} is not connected'.format(self.name))
```
klauer: Maybe `NotConnectedError` or `TimeoutError` (as it's trying to connect but failing?)

dchabot: Not a `TimeoutError`. 
I am not aware of a native exception that was good match, and hesitant to make something new for this.

Perhaps I should though...?

klauer: Anything's better than `Exception` - as the caller, you have no idea what happened if something so generic is raised.

Personally, I like when packages which define their own exceptions use a common base class. That's what I was going for in `ophyd/utils/errors.py`.

dchabot: Doesn't the exception message count as info?

Anyway, point taken. Will adjust PR content.

klauer: Consider the following:
```python
try:
    do_something()
except VerySpecificError:
    pass
except Exception:
    pass
```

vs

```python
try:
    do_something()
except Exception as ex:
    if 'is not connected' in ex.message:
        pass
    elif 'something else' in ex.message:
        pass
    else:
        pass
```

dchabot: Yep, that's ugly. Have you ever seen code in the wild written like that???

BTW - I meant "info for human beings to ponder", not programmatic digestion.

klauer: I see what you mean. But yeah, unfortunately with just the exception class `Exception` you're left with no real choice but such ugliness. I've both seen and (regrettably) had to write code that checks the message like the above :expressionless:

tacaswell: Patterns like this can be very helpful, we use it heavily in bluesky https://github.com/NSLS-II/bluesky/blob/master/bluesky/run_engine.py#L618 to deal with cleaning up with a variety of different failure modes.

PR #145, opened 2015-09-22
ENH: Add Eiger AreaDetector class.

tacaswell:


danielballan:
Note: I am merging this because it has been in activate use at CHX for more than a month. We just forgot to push it upstream.

PR #148, opened 2015-09-29
REF: exchange print statements for logger.debug

dchabot:


ophyd/controls/area_detector.py
```diff
@@ -181,7 +184,7 @@ def configure(self, **kwargs):
         self._acquire.put(0)
         while self._acquire.value:
             time.sleep(0.5)
-            print('[!!] Waiting for camera to stop acquiring ....')
+            logger.debug'[!!] Waiting for camera to stop acquiring ....')
```
tacaswell: These all lost the opening `(`

klauer:
Despite how horribly annoying those `'waiting for ...'` messages can be, I'd rather see them than wonder why the scan is hanging. Would you agree with raising some of those to `logger.warning`?

danielballan:
New area detector uses logging, not print.

klauer:
Looking back on this issue, it may be nice to have some convenience functions to set logging levels on the appropriate logger for areaDetector (and related) debugging.

PR #149, opened 2015-09-30
Fix signatures of configure/deconfigure; add 'state' property.

danielballan:
The signature changes are noncontroversial. Someone read them to make sure I didn't miss something?

I called the new property 'state'. I am not attached to that name.@klauer expressed some minor reservations about 'report' because it could be a verb or a noun -- hard to guess if it's a method or a property.

Whatever we choose let's get this merged soon so the framework is in place and tested in time for beam. We can add actual *content* to state pretty safely as we go.

### DIFF
--- a/ophyd/controls/ophydobj.py
+++ b/ophyd/controls/ophydobj.py
@@ -323,6 +323,13 @@ class OphydObject(object):
     def state(self):
         return {}
 
+    def configure(self, state=None):
+        # does nothing; subclasses can override if configuration is possible
+        return self.state, self.state
+
+    def deconfigure(self):
+        return self.state
+
     def __repr__(self):
         return self._get_repr()
 
--- a/ophyd/controls/ophydobj.py
+++ b/ophyd/controls/ophydobj.py
@@ -319,6 +319,10 @@ class OphydObject(object):
         '''
         pass
 
+    @property
+    def state(self):
+        return {}
+
     def __repr__(self):
         return self._get_repr()
 
--- a/ophyd/controls/area_detector.py
+++ b/ophyd/controls/area_detector.py
@@ -173,8 +173,10 @@ class AreaDetector(SignalDetector):
 
         return self._get_repr(repr)
 
-    def configure(self, **kwargs):
+    def configure(self, state=None):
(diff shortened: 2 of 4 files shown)


ophyd/controls/ophydobj.py
```diff
@@ -319,6 +319,17 @@ def check_value(self, value, **kwargs):
...
+    def state(self):
+        return {}
+
+    def configure(self, state=None):
+        # does nothing; subclasses can override if configuration is possible
+        return self.state, self.state
+
+    def deconfigure(self):
+        return self.state
+
```
dchabot: I don't understand how this is the proper place in the object hierarchy.
Perhaps this change fits better in `SignalGroup`?

tacaswell: In any case, this needs to properly deal with the MI from `Detector` (which coming back to it, all of the methods it provides as a mix in now live in `ophydobj` anyway so it can probably go away).

If it goes on `SignalGroup` then we will need to re-implement it for the `Signal`/`EpicsSignal` chain.

danielballan: I was thinking of `OphydObj` as akin to our `Base`, implementing the minimal interface of "readable" object. Maybe you were thinking of `Detector` in that role. As Tom suggests, maybe those two could be merged. Is there a reason to have a base class above the base class that all readables inherit from?

danielballan: I meant to link `Base`: https://github.com/NSLS-II/bluesky/blob/master/bluesky/examples.py#L10

dchabot: It seems meaningless (to me) to offer `configure/deconfigure` methods on objects that represent a subscription registry (`OphydObject`), or on objects that represent a single process variable ('Signal`).

I realize `configure/deconfigure` could simply be a pass-through on those classes, but this just feels wrong.

tacaswell: The model in bluesky is that if an object can be read, then it also needs
{trigger, describe, configure, deconfigure, state}.

danielballan:
@dchabot If you are still OK with this, as discussed this morning, please merge. Thanks.

PR #150, opened 2015-09-30
ENH: Allow dynamic PV names in ADSignals

klauer:
Simple enhancement I've been using at the HXN - allow ADSignal descriptor PV names to be filled in with information from the detector/plugin object. Useful for devices with not-so-consistent PV prefixes/suffixes:

```python
class EpicsROI(ADBase):
    '''A configurable Xspress3 EPICS ROI'''

    bin_low = ADSignal('C{self.channel}_MCA_ROI{self.roi_num}_LLM')
    bin_high = ADSignal('C{self.channel}_MCA_ROI{self.roi_num}_HLM')
    enabled = ADSignal('C{self.channel}_ROI{self.roi_num}:EnableCallbacks',
                       has_rbv=True)
```

### DIFF
--- a/ophyd/controls/areadetector/detectors.py
+++ b/ophyd/controls/areadetector/detectors.py
@@ -119,13 +119,17 @@ def lookup_doc(cls_, pv):
 class ADSignal(object):
     '''A property-like descriptor
 
-    Don't create an EpicsSignal instance until it's accessed (i.e., lazy
-    evaluation)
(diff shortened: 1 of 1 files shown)


ericdill:
Definitely seems like a step in the right direction! :+1:

tacaswell:
This needs some documentation

klauer:
Added a bit of docs. Look ok?

PR #151, opened 2015-10-02
FIX: wh_pos with DisconnectedError

klauer:
Looks like this now:

```python
[[OPHYD_DEV]] In [1]: wh_pos()

+------------+---------------+--------------+--------------+
| Positioner |         Value |    Low Limit |   High Limit |
+------------+---------------+--------------+--------------+
| vx         |     -98.52000 |   -140.00000 |   -140.00000 |
| vy         |    2896.12300 |   2800.00000 |   3000.00000 |
| vz         |      11.41000 |      0.00000 |      0.00000 |
| zpssx      |       INVALID | Disconnected | Disconnected |
| zpssy      |       INVALID | Disconnected | Disconnected |
| zpssz      |       INVALID | Disconnected | Disconnected |
+------------+---------------+--------------+--------------+
```

### DIFF
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -572,9 +572,11 @@ def logbook_add_objects(objects, extra_pvs=None):
     msg += '{:-^120}\n'.format('')
 
     # Make a list of all PVs and positioners
-    pvs = [o.report['pv'] for o in objects]
+    reports = [o.report for o in objects]
+    pvs = [report['pv'] for report in reports]
     names = [o.name for o in objects]
-    values = [str(o.value) for o in objects]
+    values = [str(v) for report in reports for k, v in report.items() if k != 'pv']
+    #values = [str(report[o.value) for o in objects]
     if extra_pvs is not None:
         pvs += extra_pvs
         names += ['None' for e in extra_pvs]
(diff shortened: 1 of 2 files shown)


danielballan:
Looks good to me. I'll let one more pairs of eyes take a look before it's merged.

danielballan:
@klauer Could you verify that this fixes `log_pos` also?

danielballan:
@dchabot Can you make unit tests a high priority? It's a bummer that we were able to render these two functions unusable without noticing it.

klauer:
Added some untested fixes for `log_pos_diff`/`log_pos_mov` - which we don't use at the HXN. If you agree with the implementation, mind giving it a go @dchabot?

klauer:
`log_pos` fixes from @ericdill added - reported working at the beamline. Look good enough to merge yet?

ophyd/commands.py
```diff
@@ -538,9 +572,12 @@ def logbook_add_objects(objects, extra_pvs=None):
     msg += '{:-^120}\n'.format('')
 
     # Make a list of all PVs and positioners
-    pvs = [o.report['pv'] for o in objects]
+    reports = [o.report for o in objects]
+    pvs = [report['pv'] for report in reports]
     names = [o.name for o in objects]
-    values = [str(o.value) for o in objects]
+    values = [str(v) for report in reports
+              for k, v in report.items() if k != 'pv']
```
ericdill: @klauer : @dchabot wanted to have a bit more of a think about this line as it is pretty awful.  `o.value` was raising if a motor was disconnected, so not sure that this is the right place to handle that.  If @dchabot signs off on this, then this PR  looks good to me

danielballan:
@dchabot?

PR #156, opened 2015-10-14
ENH: scaler channel start number configurable

tacaswell:


danielballan:
This works at XPD. It is probably worth merging, but there has to be a more general way to structure these things.

danielballan:
attn @dchabot can this go in? XPD is currently running source-installed ophyd to have this patch.

PR #163, opened 2015-10-30
Enh squash stack

tacaswell:
The replaces #131.  ~~It is the same commits, just rebased to deal with a merge conflict.~~

@dchabot Please review this, it is high priority for XPD

### DIFF
--- a/ophyd/controls/area_detector.py
+++ b/ophyd/controls/area_detector.py
@@ -851,21 +851,71 @@ class AreaDetectorFileStoreTIFF(AreaDetectorFSIterativeWrite):
 
 class AreaDetectorFileStoreTIFFSquashing(AreaDetectorFileStoreTIFF):
 
+    def __init__(self, *args, **kwargs):
+        super(AreaDetectorFileStoreTIFFSquashing, self).__init__(
+                *args, **kwargs)
+
+        self.add_signal(self._ad_signal('Proc1:NumFilter', '_num_filter',
+                                        recordable=False),
+                        add_property=True)
+
+    def set_images(self, num_images, num_filter):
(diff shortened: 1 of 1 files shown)


ophyd/controls/area_detector.py
```diff
@@ -196,6 +196,9 @@ def configure(self, state=None):
 
         # If using the stats, configure the proc plugin
 
+        # Enable images to be saved
+        self._write_plugin('EnableCallbacks', 1, self._file_plugin)
```
tacaswell: On the other PR I left the comment "This should move down to the *Filestore child class"

ophyd/controls/area_detector.py
```diff
@@ -638,7 +638,6 @@ def configure(self, state=None):
         self._write_plugin('AutoSave', 1, self._file_plugin)
         self._write_plugin('NumCapture', 0, self._file_plugin)
         self._write_plugin('FileWriteMode', 2, self._file_plugin)
-        self._write_plugin('EnableCallbacks', 1, self._file_plugin)
```
stuwilkins: Out of curiosity why is this taken out?

tacaswell: it is taken care of in the base class `configure`

stuwilkins: Ok 10-4.

tacaswell: This bit of code has had a tremendous about of random hacking without any
tests so great skepticism is called for.

dchabot: Is this actually being used at 28id? Successfully?

tacaswell: Not with this exact code path because things changed under it.

PR #164, opened 2015-11-05
ENH: add top-level bluesky-related callbacks module (with uid publisher)

klauer:
See [https://github.com/NSLS-II/bluesky/issues/220] for details

Made a bit more generic from my HXN implementation as `EpicsSignal` shouldn't be necessary. Added in some EPICS database notes.

### DIFF
--- a/ophyd/callbacks.py
+++ b/ophyd/callbacks.py
@@ -1,0 +1,76 @@
+import logging
+
+from .utils import (DisconnectedError, TimeoutError)
+
+logger = logging.getLogger(__name__)
+
+
+class UidPublish:
+    '''Publishes current run start document UID to a given signal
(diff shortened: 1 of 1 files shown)


klauer:
Some analysis-related and realtime scan viewer things - it's the start of our data pipeline until we come up with a standardized solution.

I use it in conjunction with [this](https://github.com/NSLS-II-HXN/hxntools/blob/bs/hxntools/scan_monitor.py) as I mentioned in the other issue. The base class monitors the uid signal and notifies those subscribed; the sub-classes of it grab info from mds or distill the run start document into a few parameters we want to know about for each scan at the HXN.

tacaswell:
:beers:

PR #168, opened 2015-11-24
FIX: interpret badly-formed version from pyepics

klauer:
Not sure how this got buried. This is a fix from @tacaswell, but is really important for ophyd as with the current installation we're getting all local timestamps (i.e., when the monitor update was received).

Alternatively we should just not set 'form' manually anymore. After a lengthy discussion (and some of my fixes) with pyepics-lead @newville, he changed the default to using server-side PV timestamps.

### DIFF
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -233,6 +233,8 @@ def get_pv_form():
     -------
     {'native', 'time'}
     '''
+    def _fix_git_versioning(in_str):
+        return in_str.replace('-g', '+g')
 
     def _naive_parse_version(version):
         try:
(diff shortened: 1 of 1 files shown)


danielballan:
The code I see doesn't seem to match what @klauer is writing about. Did some commits get missed? This PR just fixed the format of the `__version__` string.

klauer:
Not sure if this will make it any clearer, but here goes:
Without the fix in the PR, the version parsing fails with @tacaswell's pyepics/versioneer package. When version parsing fails, it chooses form='native' - i.e., local timestamps.
With the fix in the code, the version parsing succeeds and it recognizes the version is > 3.2.3. It then chooses form='time' - server-side timestamps.

<= 3.2.3 had a segfault bug when using server-side timestamps which is the reason for this in the first place.

danielballan:
I see. Thanks for explaining.

PR #183, opened 2015-12-16
FIX: Remove attrs that do not belong to plugins.

danielballan:
This is after length discussion with and assistance from @dchabot and @arkilic. We are pretty sure that these fields do not *belong* to the plugins; they are just used by them. They should not be exposed per-plugin.

Also see:

```python
In [42]: for attr in dir(det.tiff1):
    try:
            getattr(det.tiff1, attr)
                except ophyd.utils.TimeoutError:
                        print(attr)
                           ....:
pool_alloc_buffers
pool_free_buffers
pool_max_buffers
pool_max_mem
pool_used_buffers
pool_used_mem
queue_free
queue_free_low
queue_size
queue_use
queue_use_high
queue_use_hihi
write_message
write_status
```

Those last two actually *do* belong to the fileplugin -- they are just failing because of an issue with the test IOC.

### DIFF
--- a/ophyd/controls/areadetector/plugins.py
+++ b/ophyd/controls/areadetector/plugins.py
@@ -66,13 +66,6 @@ class PluginBase(ADBase):
     asyn_io = C(EpicsSignal, 'AsynIO')
 
     nd_attributes_file = C(EpicsSignal, 'NDAttributesFile', string=True)
-    pool_alloc_buffers = C(EpicsSignalRO, 'PoolAllocBuffers')
-    pool_free_buffers = C(EpicsSignalRO, 'PoolFreeBuffers')
-    pool_max_buffers = C(EpicsSignalRO, 'PoolMaxBuffers')
-    pool_max_mem = C(EpicsSignalRO, 'PoolMaxMem')
-    pool_used_buffers = C(EpicsSignalRO, 'PoolUsedBuffers')
-    pool_used_mem = C(EpicsSignalRO, 'PoolUsedMem')
-    port_name = C(EpicsSignalRO, 'PortName_RBV', string=True)
 
     @property
     def array_pixels(self):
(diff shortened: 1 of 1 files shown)


arkilic:
Come back and merge this @klauer

ophyd/controls/areadetector/plugins.py
```diff
@@ -66,13 +66,6 @@ class PluginBase(ADBase):
     asyn_io = C(EpicsSignal, 'AsynIO')
 
     nd_attributes_file = C(EpicsSignal, 'NDAttributesFile', string=True)
-    pool_alloc_buffers = C(EpicsSignalRO, 'PoolAllocBuffers')
-    pool_free_buffers = C(EpicsSignalRO, 'PoolFreeBuffers')
-    pool_max_buffers = C(EpicsSignalRO, 'PoolMaxBuffers')
-    pool_max_mem = C(EpicsSignalRO, 'PoolMaxMem')
-    pool_used_buffers = C(EpicsSignalRO, 'PoolUsedBuffers')
-    pool_used_mem = C(EpicsSignalRO, 'PoolUsedMem')
-    port_name = C(EpicsSignalRO, 'PortName_RBV', string=True)
```
klauer: `port_name` does not belong in this PR, it is very useful

klauer:
I'm OK with removing these from the plugins, but I'd like to still have them accessible in the future for debugging purposes.

Perhaps create a `ophyd.controls.areadetector.debug` module and have a set of `Pool`/`Queue` Devices or something along those lines.

danielballan:
Apparently the version of AD on @klauer's test stand *has* these attrs, whereas the one @arkilic, @dchabot and I were testing on does not.

Any opinions on which version of AD we should standardize the facility on? Any reason not to use "latest"? cc @stuwilkins

klauer:
For reference, I am using 1-9-1 from the Debian packages and @dchabot's example IOC startup script.

klauer:
Further reference, current (2-x) ADCore database templates:

`NDPluginBase` has the `Queue` stuff: https://github.com/areaDetector/ADCore/blob/master/ADApp/Db/NDPluginBase.template#L143

NDPluginBase includes `NDArrayBase`, which has the `Pool` stuff: https://github.com/areaDetector/ADCore/blob/master/ADApp/Db/NDArrayBase.template#L352

This leads me to believe (as I thought I had verified in the past) that the PVs for plugins have not changed... ?

Can you please look further into this @dchabot?

tacaswell:
Closing as not helpful

PR #185, opened 2015-12-17
WIP: Flesh out DetectorBase with trigger, stage, unstage

danielballan:


danielballan:
This is my attempt to reproduce the acquisition logic in @stuwilkins' AreaDetector implementation in the new ophyd. The [original implementation of AreaDetector](https://github.com/NSLS-II/ophyd/blob/82d82feb07d274884d3ada565bd8123f18ec48ec/ophyd/controls/area_detector.py#L13) used more private flags and more subscriptions -- in particular [this method](https://github.com/NSLS-II/ophyd/blob/82d82feb07d274884d3ada565bd8123f18ec48ec/ophyd/controls/area_detector.py#L149) and associated logic -- but I think this PR achieves the same functionality with less complexity. Please review carefully, @stuwilkins and @dchabot, and let me know if I am wrong about that.

I have not yet included the darkframe functionality, but did I leave in the original implementation's mechanism for this. I will add it in future commits. There is still some debate about how best to make it general: through a special method (`trigger_dark()`) or a stateful button (`make_next_call_to_trigger_a_dark_one()`).

I will also add the plugin setup code and the filestore integration in future commits.

ophyd/controls/areadetector/detectors.py
```diff
@@ -39,9 +41,138 @@
            'URLDetector',
            ]
 
+def set_and_wait(signal, val):
```
danielballan: @stuwilkins This function encapsulates a pattern you used a couple times. But I'm wondering if there is any reason why wouldn't replace `set_and_wait(signal, val)` with `signal.put(val, wait=True)`. @dchabot says that `wait` is only as reliable as the underlying driver. Can we trust area detector to wait correctly until the value has been set?

ophyd/controls/areadetector/detectors.py
```diff
@@ -39,9 +41,138 @@
...
+        self._stage_sigs = {self.cam.acquire: 0,  # If acquiring, stop.
+                            self.cam.image_mode: 1  # 'Multiple' mode
+                           }
+
+        self._staged = False
+
+        self._acquisition_signal = self.cam.acquire  # for generality's sake
+        self._acquisition_signal.subscribe(self._acquire_changed)
+
+    def stage(self):
```
danielballan: Except for the last line, this code (and all the code for `unstage`) is completely generic. Maybe it should move into `OphydDevice`.

ophyd/controls/areadetector/detectors.py
```diff
@@ -39,9 +41,138 @@
...
+
+        # Call unstage() on child devices (including, notably, plugins).
+        for signal_name in self.signal_names:
+            signal = getattr(self, signal_name)
+            if hasattr(signal, 'stage'):
+                signal.stage()
+
+        self._staged = False
+
+    def trigger(self):
```
danielballan: The lion's share of the complexity in this code is due to letting one `trigger` (formerly known as `acquire`) invoke multiple acquisitions (formerly `_start_acquire`). This allows the detector to be in charge of dark frames and to isolate its activities from the scan logic. Is it worth it?

tacaswell: I think so.  The detector already has enough control to manage taking a dark frame (from the C layer) no point in leaving that on the table.

The advantage of making the detectors in charge of this is that you can use the same `Plan` with a wide range dark frame stratgies (as the DF logic is selected by setting state on the AD object).  The big selling point here is that BS becomes blissfully unaware of the ccd/IP vs detris/eiger/pilatus differences. 

Putting the logic in the AD opens up the possibility of it re-taking dark frames in response to external signals passed on some global/bigger than single run start/end scope (ex ever X minutes, every N frames) or on change of  (a sub-set?) of the configuration.

Having 'last dark frame' be state that the detector holds would also make the work flow where the user explicitly takes dark frames 'when they feel like it' possible.

I think we should have a `take_dark_frame` method, but not neccessarily expose it through the BS interface.

danielballan:
After in-person discussion with @dchabot, @tacaswell and I are convinced that #186 is the better approach.

ophyd/controls/areadetector/detectors.py
klauer: @dchabot is correct. This sort of function is necessary.

danielballan: Noted and applied to the `ad-alt` branch. Thanks.

PR #186, opened 2015-12-18
Alternate approach to AD: split out functionality that does multiple acq. per trigger

danielballan:
This is an alternate to #185.

There are two tricky problems that the original area detector implementation solved:
1. Monitoring the acquisition signal for an acquisition to be complete
2. Initiating multiple acquisitions per trigger.

In this PR, `DetectorBase` handles (1) and `MultiAcqDetectorBase` handles (2).

Also, the universal logic of `stage` and `unstage` in now in `OphydDevice`. It looks for a dict `_staged_sigs`, mapping signals to values, which can be populated by subclasses.


danielballan:
This needs more tests, but the pieces are in place and ready for review. The test script I have been using to work with the HDF5 plugin is [this gist](https://gist.github.com/danielballan/b01eaa3fccd804517381).

danielballan:
Tests pass! This gets AreaDetector and the file output plugins working and minimally tested, including filestore integration.

In future PRs, we will:
* implement the darkframe behavior, which should now be straightforward
* implement a filestore mixin that reads the image array directly through EPICS (mainly for testing purposes)
* implement a trigger mixin for Simon's free-running mode at XPD
* implement a variety of ways of using HDF5plugin
* test the actual contents of the output from the HDF5 and TIFF plugins

ophyd/controls/areadetector/filestore_mixins.py
```diff
@@ -0,0 +1,167 @@
...
+            (self.file_number, 0),
+            (self.auto_save, 1),
+            (self.num_capture, 0),
+            (self.file_write_mode, 1),  # 'capture' mode -- for AD 1.x
+            (self.file_path, full_write_path),
+            (self.file_name, self._filename),
+            (self.capture, 1),
+        ]
+        self.stage_sigs.extend(ssigs)
+        print('staging fsbase')
```
tacaswell: rouge print!

klauer: I was about to nitpick about your spelling, but it is a "rouge rogue print" with the syntax colorscheme.

ophyd/controls/device.py
```diff
@@ -233,7 +240,77 @@ def __new__(cls, name, bases, clsdict):
         return clsobj
 
 
-class OphydDevice(OphydObject, metaclass=ComponentMeta):
+
+# These stub 'Interface' classes are the apex of the mro heirarchy for
+# their respective methods. They make it safe to use multiple inheritance
+# with the base classes given in any order.
```
tacaswell: Is this comment actually true?

ophyd/controls/device.py
```diff
@@ -233,7 +240,77 @@ def __new__(cls, name, bases, clsdict):
...
+        for sig, val in list(self._original_vals.items())[::-1]:
+            set_and_wait(sig, val)
+
+        # Call unstage() on child devices (including, notably, plugins).
+        for signal_name in self.signal_names:
+            signal = getattr(self, signal_name)
+            if hasattr(signal, 'unstage'):
+                signal.unstage()
+
+        self._staged = False
```
tacaswell: Do we want to clear the old original values?

ophyd/controls/device.py
```diff
@@ -233,7 +240,77 @@ def __new__(cls, name, bases, clsdict):
...
+
+
+class GenerateDatumInterface:
+    """Classes that inherit from this can safely customize the
+    `generate_datum` method without breaking mro. If used along with the
+    BlueskyInterface, inherit from this second."""
```
tacaswell: Doesn't this need a stub function?

ophyd/utils/epics_pvs.py
```diff
@@ -336,3 +341,16 @@ def wrapper(self, *args, **kwargs):
...
+
+def set_and_wait(signal, val):
+    """
+    Set a signal to a value and wait until it reads correctly.
+
+    There are cases where this would not work well, so it should be revisited.
+    """
+    signal.put(val)
+    while signal.get() != val:
+        ttime.sleep(0.1)
```
tacaswell: The wait time should probably be an input

ophyd/controls/ophydobj.py
```diff
@@ -36,11 +38,12 @@ def _finished(self, *args, **kwargs):
         #     print("this should be empty: {}".format(args))
         # if kwargs:
         #     print("this should be empty: {}".format(kwargs))
-        self.done = True
+        with self._lock:
```
klauer: Consider a `locked` decorator as this is used in a few places here

ophyd/controls/device.py
```diff
@@ -303,12 +380,12 @@ def wait_for_connection(self, all_signals=False, timeout=2.0):
         # Instantiate first to kickoff connection process
         signals = [getattr(self, name) for name in names]
 
-        t0 = time.time()
-        while timeout is None or (time.time() - t0) < timeout:
+        t0 = ttime.time()
```
klauer: You guys and your `ttime`...

ophyd/controls/device.py
```diff
@@ -233,7 +240,77 @@ def __new__(cls, name, bases, clsdict):
...
+        """
+        Restore the device to 'standby'.
+
+        Multiple calls (without a new call to 'stage') have no effect.
+        """
+        if not self._staged:
+            return
+
+        # Restore original values.
+        for sig, val in list(self._original_vals.items())[::-1]:
```
klauer: `reversed()`

ophyd/controls/device.py
```diff
@@ -233,7 +240,77 @@ def __new__(cls, name, bases, clsdict):
...
+            set_and_wait(sig, val)
+
+        # Call unstage() on child devices (including, notably, plugins).
+        for signal_name in self.signal_names:
+            signal = getattr(self, signal_name)
```
klauer: This will instantiate _all_ signals, even lazy ones - is this intended?

ophyd/controls/device.py
klauer: Maybe, but it might also be useful for debugging to be able to poke around

ophyd/controls/device.py
```diff
@@ -386,6 +467,13 @@ def trigger_signals(self):
 
         return [getattr(self, name) for name in names]
 
+    def _done_acquiring(self, **kwargs):
```
klauer: Great change. I would have put it around line 500 though. :)

ophyd/utils/epics_pvs.py
klauer: Agreed

ophyd/utils/epics_pvs.py
```diff
@@ -336,3 +341,16 @@ def wrapper(self, *args, **kwargs):
...
+        logger.info("Waiting for %s to be set...", signal.name)
```
klauer: A bit more info would be nice (current_value, set_value). Also, because I'm in that mood, there's a stray EOL at EOF

ophyd/controls/device.py
danielballan: It makes it *more* forgiving but not universally forgiving, depending on what you are doing.

ophyd/controls/device.py
danielballan: Yes, and the short-circuiting (`if self._staged`...) helps protect the residual values from having unintended consequences.

ophyd/controls/areadetector/filestore_mixins.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def new_short_uid():
+    "uuid4, skipping the last stanza because of AD length restrictions."
+    return '-'.join(new_uid().split('-')[:-1])
+
+
+class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
+    "Base class for FileStore mixin classes"
+    def __init__(self, *args, write_file_path=None, read_file_path=None,
```
klauer: `write_file_path` and `read_file_path` in the initializer might make it seem like you can only set them in the class definition, which in the case of 2 detector instances would be annoying. Might be worthwhile to make a note that these can be set at any point.

Actually, with the check for `read_file_path` vs `write_file_path` it may make more sense to bury the logic in a pair of properties.

ophyd/controls/device.py
```diff
@@ -233,7 +240,77 @@ def __new__(cls, name, bases, clsdict):
...
+            if hasattr(signal, 'stage'):
+                signal.stage()
+
+    def unstage(self):
+        """
+        Restore the device to 'standby'.
+
+        Multiple calls (without a new call to 'stage') have no effect.
+        """
+        if not self._staged:
```
klauer: A `logger.debug` message here would probably be worthwhile if we ever have to try and debug this mess of inheritance. One in `stage` would be nice, too.

ophyd/controls/areadetector/trigger_mixins.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, *args, **kwargs):
+        print("TriggerBase before super")
+        super().__init__(*args, **kwargs)
+        print("TriggerBase after super")
+        # settings
+        self.stage_sigs.extend([(self.cam.acquire, 0),  # If acquiring, stop.
+                                (self.cam.image_mode, 1),  # 'Multiple' mode
+                               ])
+        self._acquisition_signal = self.cam.acquire
+        self._acquisition_signal.subscribe(self._acquire_changed)
```
klauer: There's no `TriggerBase` definition of `_acquire_changed`

ophyd/controls/device.py
danielballan: Many of the methods using during a scan (stage, describe, read) involve recursively searching child signals. So I assumed that at this point loading signals is unavoidable. I guess we could only stage the signals that are in read_signals and configure_signals. That seems fraught to me, but I'd defer to you on this.

ophyd/controls/device.py
danielballan: No accounting for taste.

ophyd/controls/ophydobj.py
danielballan: Neat idea. Done.

ophyd/utils/epics_pvs.py
danielballan: While I'm here, I'll refresh the current value on each loop, so we can't see if it's wait for `/tmp/` to be updated to `/tmp` or some other near miss.

ophyd/controls/areadetector/trigger_mixins.py
danielballan: That's intentional. I will document that subclasses must define this.

ophyd/controls/device.py
danielballan: We'll get log messages from `set_and_wait`, below. You're think of a message saying `"I am about to stage the signals for %r", self`?

klauer: Was mainly thinking that if somehow we hit the `if not self._staged` clause, a debug message to know something went really wrong would be a good idea.

ophyd/controls/areadetector/filestore_mixins.py
danielballan: You lost me on that specific point: how does making something an arg in the init make it seem like it can only be set in the class definition?

But I agree that the pattern I chose here -- required, keyword-only arguments -- is a little strange. I chose it because it plays very safely with the MI. It might become more obvious what the *best* way is once we begin to roll it out.

klauer: It's a minor thing, really. It's just that if you expect the logic shown in the initializer to be consistent, it may surprise you:

```python
class Detector(Device):
    p = Cpt(FilePlugin, write_file_path='/tmp')

det1 = Detector()
# ok, good

det2 = Detector()
det2.p.write_file_path = '/something_new'
# but det2.p.read_file_path = '/tmp'
```

ophyd/controls/device.py
danielballan: Done. Also, reconsidering stage/unstage, I think repeated calls to `stage` should raise. Our options are:

1. Just stage again -- thus erasing the "original values", which is bad.
2. Fail loudly.
3. Fail silently.

I don't want to have to wonder whether it is passing silently (debug logging notwithstanding). Repeated calls to unstage are harmless, though, so that one can pass silently with a debug message, as it was.

ophyd/controls/areadetector/filestore_mixins.py
danielballan: Ah, gotcha. Agreed. Will fix...

tacaswell: I agree that the read/write file_path nonsense needs to be wrapped up it a bit more tooling, however from watching beamlines actually run, they seem to always save their data to the same 'root' folder on a giant network mount.  I think for the first cut leaving this a) in the `__init__` b) telling people that if they change it they need to know what they are doing is ok.

ophyd/controls/device.py
tacaswell: I am in favor of option 2

ophyd/controls/areadetector/filestore_mixins.py
danielballan: Hmm. The logic here is annoyingly subtle. Should `write_file_path` *always* blow away `read_file_path`? Should `read_file_path` become independent once it has been set by the user? That involves some potentially-confusing state -- if I found that in someone's code I would be angry about it. ;- )

Maybe I'm missing an obvious way to handle this.

ophyd/controls/device.py
danielballan: Good. That is what I implemented here. But I gather you still agree that repeated calls to `unstage` can be "indempotent"?

ophyd/controls/areadetector/filestore_mixins.py
tacaswell: the obvious way to fix this in to unify the mount points :stuck_out_tongue_winking_eye: 

The problem is although this things must be perfectly in sync, but it is dependent on humans not changing things that we can not see.

ophyd/controls/device.py
tacaswell: Yes re unstage.

danielballan:
Builds are failing because Travis is acting out today. @tacaswell can you run these locally in your Docker?

ophyd/controls/areadetector/filestore_mixins.py
klauer: I think in the case of `read_file_path` being unset, it should default to the current `write_file_path` (my suggestion at the top, though not expressed very well, was to do this in a property).

danielballan:
Superceded by #192, finally passing Travis

PR #188, opened 2015-12-23
WIP: mca support. Do not merge yet.

dchabot:
Attn @cmazzoli

There are still a few hours of work left:
* additional, common MCA Record attributes on EpicsMCA
* create subclass for DXP Saturn
* more tests (live and simulated-PV)

Merry Ho Ho.


ophyd/controls/mca.py
```diff
@@ -0,0 +1,107 @@
...
+from collections import namedtuple
+
+from .signal import (EpicsSignal, EpicsSignalRO)
+from .device import OphydDevice
+from .device import (Component as C, DynamicDeviceComponent as DDC)
+
+logger = logging.getLogger(__name__)
+
+
+_roi_field_map = {'name': (EpicsSignal, '.R{n}NM'),
```
tacaswell: This needs to be an ordered dict or the named tuple will randomly re-order process-to-process.

ophyd/controls/mca.py
```diff
@@ -0,0 +1,107 @@
...
+            kws = {k: v[0](self.prefix + v[1].format(n=roi))
+                   for k, v in _roi_field_map.items()}
+            setattr(self, 'roi{n}'.format(n=roi), _Roi(**kws))
+
+    def stage(self):
+        '''Stage the MCA for data acquisition'''
+        self._old_mode = self.mode.get()
+        self.mode.put(0)
+
+    def configure(self, d=None):
```
tacaswell: Isn't this taken care up by a parent class now?  I think it will 'do the right thing' if you just remove this.

ophyd/controls/mca.py
dchabot: Easy enough to fix.
But, why might I care about order here?

ophyd/controls/mca.py
dchabot: Are you referring to `stage( )`?
Then, no...

The parent, `OphydDevice` takes a `pass` on this.
The parent should `raise NotImplementedError`, or similar.

ophyd/controls/mca.py
tacaswell: so the `_Roi` has a deterministic order or `mca.roi1[0]` will be a different value each time.

ophyd/controls/mca.py
tacaswell: no the `configure` method

ophyd/controls/mca.py
```diff
@@ -0,0 +1,107 @@
...
+           This will add 5 ROIs, vtx.roi0 through vtx.roi4, and their
+           EpicsSignals: roiN.(name,cnt,net_cnt,preset_cnt, is_preset,
+                               bkgnd_chans, hi_chan, lo_chan)
+           '''
+        for roi in rois:
+            assert 0 <= roi < 32
+
+            kws = {k: v[0](self.prefix + v[1].format(n=roi))
+                   for k, v in _roi_field_map.items()}
+            setattr(self, 'roi{n}'.format(n=roi), _Roi(**kws))
```
tacaswell: is it ok that calling `add_roi` multiple times may over-write an existing roi?

ophyd/controls/mca.py
danielballan: I think @klauer ran out of time to implement this but it can and probably should be done in the base class.

danielballan:
The `configuration` method discussion in previous line comments is now in PR #189

ophyd/controls/mca.py
```diff
@@ -0,0 +1,103 @@
...
+           This will add 5 ROIs, vtx.roi0 through vtx.roi4, and their
+           EpicsSignals: roiN.(name,cnt,net_cnt,preset_cnt, is_preset,
+                               bkgnd_chans, hi_chan, lo_chan)
+           '''
+        for roi in rois:
+            assert 0 <= roi < 32
+
+            kws = {k: v[0](self.prefix + v[1].format(roi))
+                   for k, v in _roi_field_map.items()}
+            setattr(self, 'roi{}'.format(roi), _Roi(**kws))
```
klauer: You could make a `class ROI(OphydDevice)`, then instantiate them here with a prefix of `{mca_prefix}.R{roi}`. It would probably be a lot more clear what's going on then. (Or is there something that makes this impossible?)

danielballan:
This is looking good. You can replace stage/unstage with the following line in `__init__`, after the call to `super()`:

```python
self.stage_sigs.extend([(self.call, 0)])
```

The implementation of stage/unstage in `Device` will then handle:
* correctly raising/passing on repeated unstage/stage respectively
* using `set_and_wait` instead of `put`

as well as stashing the original values, as you did manually here.

ophyd/mca.py
```diff
@@ -0,0 +1,153 @@
...
+from .areadetector import EpicsSignalWithRBV as SignalWithRBV
+
+
+logger = logging.getLogger(__name__)
+
+
+class ROI(Device):
+
+    # 'name' is not an allowed attribute
+    label = C(EpicsSignal, 'NM', lazy=True)
```
klauer: Thinking aloud: maybe we can test to see if someone tries overriding `OphydObject` `name`/etc with components.

klauer:
Looks great!

ophyd/mca.py
dchabot: FYI (and giggles), that operation goes BOOM with:
```python

/home/dchabot/Dropbox/workspace/ophyd/ophyd/device.py in create_component(self, instance)
     70         '''Create a component for the instance'''
     71         kwargs = self.kwargs.copy()
---> 72         kwargs['name'] = '{}_{}'.format(instance.name, self.attr)
     73
     74         for kw in self.add_prefix:

RuntimeError: maximum recursion depth exceeded while calling a Python object
```

tacaswell:
Some of that was fixed in #195

danielballan:
@dchabot Actually it looks like you don't have the commits from #195 in here. Rebasing on master should fix the issue with '0' and 0.

danielballan:
I'm going to close and reopen to re-run the tests based on current master, particularly to verify that #204 plays fine with this. (No reason it shoudln't... just being cautious.)

ophyd/mca.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # NOTE: all SignalWithRBV are "lazy=True"
+    # Trigger Filter PVs
+    trigger_peaking_time = C(SignalWithRBV, 'TriggerPeakingTime')
+    trigger_threshold = C(SignalWithRBV, 'TriggerThreshold')
+    trigger_gap_time = C(SignalWithRBV, 'TriggerGapTime')
+
+    max_width = C(SignalWithRBV, 'MaxWidth')
+
+    # Energy Filter PVs
```
danielballan: The neatness of this makes me really happy. :- )

PR #189, opened 2016-01-04
Implement a richer configure method in OphydDevice.

danielballan:
We have been returning `{}, {}`. This fills out a basic implementation that returns actual data. It relies on `configuration_attrs`.

### DIFF
--- a/ophyd/controls/device.py
+++ b/ophyd/controls/device.py
@@ -479,8 +479,9 @@ class OphydDevice(OphydObject, metaclass=ComponentMeta):
     def configure(self, d=None):
         '''Configure the device for something during a run
 
-        The default implementation does nothing; subclasses should override
-        this to make configuration specific to the device.
+        This default implementation allows the user to change any of the
+        `configuration_attrs`. Subclasses might override this to perform
+        additional input validation, cleanup, etc.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


ophyd/controls/device.py
```diff
@@ -489,10 +490,22 @@ def configure(self, d=None):
...
+        for key, val in d.items():
+            if key not in self.configuration_attrs:
+                # a little extra checking for a more specific error msg
+                if key not in self.signal_names:
+                    raise ValueError("there is no signal named %s", key)
+                else:
+                    raise ValueError("%s is not one of the "
+                                     "configuration_fields, so it cannot be "
+                                     "changed using configure", key)
+            getattr(self, key).put(val)
```
tacaswell: Is there a way to wait at the end for all of the puts to be 'done'?

tacaswell:
Did this commit add the failure or was that lurking on master?

ophyd/controls/device.py
```diff
@@ -489,10 +490,22 @@ def configure(self, d=None):
...
+        (old, new) tuple of dictionaries
         Where old and new are pre- and post-configure configuration states.
         '''
-        return {}, {}
+        old = self.read_configuration()
+        for key, val in d.items():
+            if key not in self.configuration_attrs:
+                # a little extra checking for a more specific error msg
+                if key not in self.signal_names:
+                    raise ValueError("there is no signal named %s", key)
```
klauer: `Exception('', key)` is invalid here and in the `else` clause below.
`'there'.capitalize()` would also make me slightly happier.

ophyd/controls/device.py
klauer: This should probably be a `wait=True` for pvs that support it. Finicky device support (and slow IOC-to-device communication) might make some delay here necessary.

It's worth noting that `read_configuration()` returns an ordered dictionary, so the configuration will be done in the order that was given - this can be very important for some devices.

klauer: @tacaswell serial, not parallel, puts should be the default here (with reference to waiting for them at the end). Underlying AD stuff doesn't usually handle rapid reconfiguration very well.

cowanml:
On Mon, 4 Jan 2016, Thomas A Caswell wrote:
> Did this commit add the failure or was that lurking on master?

this one.

ophyd/controls/device.py: 497:             for key, val in d.items():

just needs protection with a try or if for when d=None

ophyd/controls/device.py
danielballan: Oops, thanks. Still shaking off that vacation rust.

klauer:
Looks much better now. Think you could add a quick test or two?

tacaswell:
Why is the input to `configure` optional?  I think that is a hold over from before we pulled `stage` and `configure` apart and should not stay.  This also means that our tests are probably calling `configure` when they should be calling `stage`.

danielballan:
Good point. I think, yes, that was a hold-over. As we discussed offline, if someone wants to call `configure({})` they should be explicit about that `{}`.

Stand by for tests....

PR #193, opened 2016-01-10
MNT: Flatten file structure

danielballan:
as discussed with @dchabot, @klauer, and @tacaswell: while we are in the business of breaking the ophyd API anyway, we are flattening the namespace. Everything in ophyd/controls/ is moved to ophyd/.

The motivation for this change is partly "less typing" and partly reflection on [this talk](https://www.youtube.com/watch?v=o9pEzgHorH0). One of the points in the talk: the standard library modules have namespaces that are relatively flat, and it works well for core Python.

Renamed using:
`find ./ -iname "*.py" | xargs sed -i "s/from \.controls import/from .
import/g" && find ./ -iname "*.py" | xargs sed -i "s/\.controls//g"`

...plus two or three manual edits. A look at `git grep controls` seems to confirm that I scrubbed it, but we'll have to wait for Travis to tell us whether I broke things.


tacaswell:
While we are at it do we want to flatten utils too?

danielballan:
The API already flat, thanks to an import. I thought about copying everything into one file but I want to talk to Daron about it first.

Issue #190, opened 2016-01-04
Naming suggestion: OphydDevice -> Device


danielballan:
It is currently `ophyd.device.OphydDevice`. I think `ophyd.device.Device` is sufficient. Only worth doing this if we do it before we begin to deploy.

tacaswell:
Might even be worth using importing to flatten the name space so we get `from ophyd import Device as OD`

klauer:
I'm OK with this. `sed` away, @danielballan.

PR #194, opened 2016-01-11
API: OphydDevice -> Device

danielballan:
one last change to the API before we deploy -- already cleared with @klauer (on GH in #190) and @dchabot (in person)


danielballan:
Incidentally, the "flattening" and this PR together turn `ophyd.controls.device.OphydDevice` into `ophyd.Device`. Definitely an improvement.

PR #195, opened 2016-01-11
New configure

tacaswell:
replaces #189

### DIFF
--- a/tests/test_scaler.py
+++ b/tests/test_scaler.py
@@ -10,6 +10,8 @@ import epics
 from ophyd import scaler
 from ophyd.utils import enum
 from .test_signal import FakeEpicsPV
+from .test_utils import assert_OD_equal_ignore_ts
+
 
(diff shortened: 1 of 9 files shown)


tests/test_scaler.py
```diff
@@ -86,9 +86,11 @@ def test_scaler_functionality(self):
         str(sca)
 
         sca.stage()
-        sca.configure()
+        old, new = sca.configure({})
```
danielballan: Scaler has a custom `configure` method defined on it which always returns a pair of empty dicts. Gotta scrub out "rogue" ("rouge?") implementations of `configure` and add a non-trivial test.

danielballan: done here, but Travis is stalling on the scaler tests, so there is probably a typo lurking: https://github.com/danielballan/ophyd/tree/new_configure

klauer: It may also be a consequence of using a 'fake' scaler in the test, though I haven't really looked into it yet.

klauer:
Looks good, I think.

PR #199, opened 2016-01-12
FIX: Check that class is not shadowing 'name'

### LINKED ISSUE #197: Don't let Components shadow `name` attribute
Ref: https://github.com/NSLS-II/ophyd/pull/188#discussion_r49373762

danielballan:
Closes #197 

How's this?

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -209,6 +209,8 @@ class ComponentMeta(type):
 
     def __new__(cls, name, bases, clsdict):
         clsobj = super().__new__(cls, name, bases, clsdict)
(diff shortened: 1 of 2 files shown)


tacaswell:
Are there any other reserved attribute names?  Probably 'parent'

danielballan:
Anything that would get overriden in the init should fail loudly.

```
name
parent
signal_names
_signals
read_attrs
configuration_attrs
monitor_attrs
```

klauer:
@danielballan yes, there should be more checks to be safe. Also, there should be tests for these cases if it ever gets broken somehow.

PR #200, opened 2016-01-12
FileStore variations

danielballan:
Needs tests. Work in progress for XPD deployment today.

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -54,14 +54,23 @@ class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
         if write_file_path is None:
             raise ValueError("write_file_path is required")
         self.write_file_path = write_file_path
-        if read_file_path is None:
-            self.read_file_path = write_file_path
-        else:
-            self.read_file_path = read_file_path
+        self._read_file_path = read_file_path
         super().__init__(*args, **kwargs)
         self._point_counter = None
         self._locked_key_list = False
         self._datum_uids = defaultdict(list)
+
+    @property
+    def read_file_path(self):
+        "Returns write_file_path if read_file_path is not set"
+        if self._read_file_path is None:
+            return write_file_path
+        else:
+            return self._read_file_path
+
+    @read_file_path.setter
+    def read_file_path(self, val):
+        self._read_file_path = val
 
     def stage(self):
         self._point_counter = count()
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -15,7 +15,7 @@ To be used like so:
     dest = '/tmp'  # in production, use a directory on your system -- not /tmp
 
     class MyDetector(PerkinElmerDetector, SingleTrigger):  # for example
-        file_plugin = MyPlugin(suffix='HDF1:', write_file_path=dest)
+        file_plugin = MyPlugin(suffix='HDF1:', write_path_template=dest)
 
     det = MyDetector(...)
 """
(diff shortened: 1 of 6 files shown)


ophyd/areadetector/filestore_mixins.py
```diff
@@ -92,15 +100,13 @@ def stage(self):
 
         # AD does this same templating in C, but we can't access it
         # so we do it redundantly here in Python.
-        fn = self.file_template.get() % (full_read_path, self._filename,
-                                         self.file_number.get())
+        self._fn = self.file_template.get() % (full_read_path,
+                                               self._filename,
+                                               self.file_number.get())
 
         if not self.file_path_exists.get():
```
klauer: Am I correct in recalling that directory creation will not be touched on in this file (purposefully)?

ophyd/areadetector/filestore_mixins.py
```diff
@@ -74,7 +83,6 @@ def stage(self):
         path = os.path.join(*(date.isoformat().split('-') + ['']))
         full_write_path = os.path.join(self.write_file_path, path)
         full_read_path = os.path.join(self.read_file_path, path)
```
klauer: Would it be possible to break out this path building into a method that can be overridden?

Because you're going to ask: in the case of tiff files where we save 300x300 = 90,000 files per scan at the HXN, we will get extreme delays when attempting to list these files over NFS. Separate directories per-scan (i.e., per-stage) would help. (And before you say 'HDF5!', it's just not feasible to move to it currently for other reasons)

danielballan: I was thinking the same thing. Would this logic suit your purposes:

```python
In [9]: write_file_path = '/tmp/%Y/%m/%d'

In [10]: datetime.now().strftime(write_file_path)
Out[10]: '/tmp/2016/01/12'

Your write_file_path could be updated in a customized `stage` method to be something like `'.../%Y/%m/%d/' + scan_id`

File: ophyd/areadetector/filestore_mixins.py (hunk 1 above)
Comment (danielballan) [inline ophyd/areadetector/filestore_mixins.py]: Correct.

Comment (dchabot) [inline ophyd/areadetector/filestore_mixins.py]: For clarity - does this scheme (in `stage`) no longer implicitly depend on the structure provided by executing the date-based "directory creation script"?

File: ophyd/areadetector/filestore_mixins.py [hunk 3]
```diff
@@ -47,34 +47,45 @@ def new_short_uid():
...
         self._locked_key_list = False
         self._datum_uids.clear()
         # Make a filename.
         self._filename = new_short_uid()
-        date = datetime.now().date()
-        # AD requires a trailing slash, hence the [''] here
-        path = os.path.join(*(date.isoformat().split('-') + ['']))
-        full_write_path = os.path.join(self.write_file_path, path)
-        full_read_path = os.path.join(self.read_file_path, path)
-        self.file_template.put('%s%s_%6.6d.h5')
```
tacaswell: I think this got lost.

ophyd/areadetector/filestore_mixins.py
```diff
@@ -134,6 +143,25 @@ def unstage(self):
...
+
+
+class FileStoreTIFF(FileStoreBase):
+    # TODO num_captured above, num_images here -- which is correct?
+    def stage(self):
+        self.stage_sigs.extend([(self.file_template, '%s%s_%6.6d.tiff')])
+        super().stage()
+        res_kwargs = {'template': self.file_template.get(),
+                      'filename': self.file_name.get(),
+                      'frame_per_point': self.num_images.get()}
```
tacaswell: This on only safe is squashing in not turned on.

klauer:
Looks good.

tacaswell:
do we need a PR with the squashing as well or do we want to put that into the BL config?

PR #204, opened 2016-01-12
PERF: Stage/unstage without instantiating signals needlessly.

danielballan:
Since `stage` and `unstage` are defined on devices but not on signals, it is actually easy to recursively stage a device and its subdevices without instantiating lazy signals -- or even non-lazy signals.

As you can see, we don't have to make reference to the lazy attribute in this implementation, only the `Component.cls`.


klauer:
Great. Works for me. One minor request - could we have a `_sub_devices` property or something to replace the duplicated code (and also aid in debugging)?

PR #205, opened 2016-01-13
FIX: correctly recursively collect _sig_attrs

### LINKED ISSUE #201: Device inheritance
In its current state, `Device` inheritance breaks `_sig_attrs` (and as @danielballan found, the trigger signal on a subclassed device).
Additionally, the current code puts signals from bases after those in the top-level class (which is reversed).

To do list:
* fix `_sig_attrs`
* ensure `signal_names` and `_sig_attr` order is correct!
* test these more thoroughly

The original changes from my commit here should fix both of the above, I believe:
https://github.com/danielballan/ophyd/commit/bf037c9b8477cb679ba023582394dd8fc2d8e972

tacaswell:
closes #201 

Tests coming.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -263,7 +263,6 @@ class ComponentMeta(type):
         return clsobj
 
 
-
 # These stub 'Interface' classes are the apex of the mro heirarchy for
 # their respective methods. They make multiple interitance more
 # forgiving, and let us define classes that customize these methods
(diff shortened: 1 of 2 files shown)


danielballan:
This looks good to me but @klauer should sign off and merge.

klauer:
Ready to be merged? Looks good to me.

tacaswell:
I have some un-committed more complicated tests, but they can go in later.

PR #208, opened 2016-01-13
ENH: Successful test on HDF5 SimDetector with AD 2.2

danielballan:
This uses Stream mode, which allows us to write one HDF5 file per *run* (strictly speaking, per stage/unstage), not per acquisition.

I executed [this script](https://gist.github.com/danielballan/42e854fdb5e65f2cb809), accessing a SimDetector IOC running AD 2.2 at xf23id (actual 2.2, not the customized branch run in production at CSX).

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -145,8 +145,8 @@ class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
 class FileStoreHDF5(FileStoreBase):
     def stage(self):
         self.stage_sigs.update([(self.file_template, '%s%s_%6.6d.h5'),
-                                (self.file_write_mode, 'Capture'),
-                                (self.capture, 1),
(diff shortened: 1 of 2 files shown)


klauer:
We talked about making different mixins for AD versions that can stream versus others that can't. Have you changed your mind?

danielballan:
@klauer I'm still open to doing both, but developing them in parallel today was too confusing. The IOC-crashing stuff was scattered among different classes.

I think making a backward-compatible version is still a good idea, but I don't yet understand if it's possible to support Stream mode at all in 1.x or if 1.x users have to make do with Capture. Depending on the demand for HDF5, it might make sense to push people to upgrade their AD. Especially with SWMR coming.

Anyhow, a capture-mode class could look something like this:

```python
class FileStoreHDF5CaptureMode(FileStoreHDF5):
        self.stage_sigs[self.file_write_mode] = 'Capture'

    def stage(self):
        set_and_wait(self.num_capture, self.num_images.get())
        super().stage()
```

PR #210, opened 2016-01-14
Fix filestore and a bug in recursive stage/unstage

danielballan:
I took the `ad-hdf5` branch (#208) for more exercise at XPD tonight. I discovered some small but important bugs -- see commits for details.

Automated tests are coming tomorrow.

### DIFF
--- a/doc/area_detector/gen.py
+++ b/doc/area_detector/gen.py
@@ -4,7 +4,7 @@
 # Started off simple, but as is usually the case with this type of
 # project, it gets cryptic pretty quickly...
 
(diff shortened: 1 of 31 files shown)


danielballan:
I should note (and celebrate!) that we were able to retrieve an image array through filestore at XPD. We had no problems with IOC instability while using TIFF. (The HDF5 plugin is another story -- on AD 2.2 at CSX it was robust, but at XPD it crashed the IOC yet again.) But the TIFF plugin integration seems to be ready for prime time.

Issue #213, opened 2016-01-14
`read_attrs` only addresses keys which map cleanly to sub-devices 


tacaswell:
ex, for an AD object

```python
sd.read_attrs = ['tiff1']
sd.tiff1.read_attrs = []
```

Is what is required to get _just_ the `image` key out of `sd.read()` not _everything_ AD knows about.

tacaswell:
I think this would be clearer if we renamed `read_attrs` -> `read_cmpts`, that is the list of a `Devices` components that it will consult as part of the read process.

Tuning how a `Devices` or  `Signal` tunes what keys it includes in the read dict that it it's self produces is a different problem.

klauer:
It's certainly too much information in the case of plugins, but isn't it useful to be able to say "I want `subdevice.**"` to be read? That's why it's implemented as-is.

klauer:
`read_cmpts` then could potentially include components that aren't even part of the `Device` hierarchy. `__getattr__` stuff probably could be removed in that case, though.

tacaswell:
I am super confused

klauer:
Am I misunderstanding your suggestion?
```python
device.read_cmpts = [device.cpt1]
```

tacaswell:
I am suggesting to leave the functionality identical, but to rename it because it confused me.

I was mapping attrs -> keys that will be in the dictionary returned by read, instead a list of the name of sub-components of the current device that will be consulted as part of the read.

klauer:
It was originally `signals` then renamed to `attrs` because of the exact problem you're describing. `attrs` is at least what Python would call them.

danielballan:
To sum up:

* "read_signals" is problematic because the names can refer to Signals or Devices.
* "read_attrs" is problematic because not all objects attributes qualify (only those that support `.get()`).
* "read_cmpts" is heavily abbreviated for my taste (I can hear myself spelling out: "No, C-M-P-T-S....")
* All of these might suggest that their contents are objects, not their string names.

How about:

* `signal_names` -> `component_names`
* `read_signals` -> `read_names` and likewise for `configuration_*`

Better make these changes Monday if we make them at all. Sanjit and Milinda already took notes on the existing terminology.

klauer:
@tacaswell  @danielballan  Are you ok with closing this? I think we agreed that all of the proposed names have their own issues, so we may as well stick with what we have (and it being somewhat late to rename things and all...)

danielballan:
[...]

PR #215, opened 2016-01-14
ENH: Build out MultiTrigger class, our new way to support detector-controlled dark frames

danielballan:
This is about done, but needs testing and probably bug-fix commits before being merged -- hopefully later today.

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -45,6 +45,9 @@ class SingleTrigger(TriggerBase):
     >>> class SimDetector(SingleTrigger):
     ...     pass
     """
+    def __init__(*args, image_name=None, **kwargs):
+        self._image_name = image_name
+        super().__init__(*args, **kwargs)
 
     def trigger(self):
         "Trigger one acquisition."
@@ -54,7 +57,9 @@ class SingleTrigger(TriggerBase):
 
         self._status = DeviceStatus(self)
         self._acquisition_signal.put(1, wait=False)
-        self.dispatch('image', ttime.time())
+        if self._image_name is None:
+            key = '_'.join(self.name, 'image')
+        self.dispatch(key, ttime.time())
         return self._status
 
     def _acquire_changed(self, value=None, old_value=None, **kwargs):
@@ -81,7 +86,8 @@ class MultiTrigger(TriggerBase):
     -------
     >>> class MyDetector(SimDetector, MultiTrigger):
     ...     pass
-    >>> det = MyDetector(acq_cycle={'image_gain': [1, 2, 8]})
+    >>> det = MyDetector(acq_cycle={'name': ['gain1', 'gain2', 'gain8'],
+    ...                             'image_gain': [1, 2, 8]})
     """
     # OphydObj subscriptions
     _SUB_ACQ_DONE = 'acq_done'
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -11,6 +11,7 @@ To be used like so:
 
 import time as ttime
 import logging
+import itertools
 
 from ..ophydobj import DeviceStatus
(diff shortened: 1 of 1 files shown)


ericdill:
closing/opening to try to trigger circle

danielballan:
This is ready for review. Tests are in [this gist](https://gist.github.com/danielballan/be8ff70448744e2f38a4).

Will have to make them work on AD 1.x before incorporating them in the codebase itself.

ophyd/areadetector/trigger_mixins.py
```diff
@@ -71,41 +80,68 @@ class MultiTrigger(TriggerBase):
...
     Example
     -------
     >>> class MyDetector(SimDetector, MultiTrigger):
     ...     pass
-    >>> det = MyDetector(acq_cycle={'image_gain': [1, 2, 8]})
+    # On the first trigger, close the shutter and acquire three images
+    # with different gain settings on the detector. Then open the shutter
+    # and take a light frame.
+    # On the next trigger, just take a light frame.
+    >>> trigger_cycle=[[('gain1', {'shutter': 'close', 'image_gain': 1}),
```
klauer: I realize that this is probably as simple as you can do it in one expression, but even knowing what it's doing I find this really difficult to parse with all the parentheses. Think temporary variables showing the different steps would be any better?

ophyd/device.py
```diff
@@ -454,7 +454,11 @@ def _read_attr_list(self, attr_list):
         '''Get a 'read' dictionary containing attributes in attr_list'''
         values = OrderedDict()
         for name in attr_list:
-            signal = getattr(self, name)
+            try:
+                signal = getattr(self, name)
+            except AttributeError:
+                # This field may be added by a mixin class.
```
klauer: Is there any way we can check? Or should the mixin support  `__getattr__`?

tacaswell: I am with @klauer these changes should be removed

danielballan: I removed the offending commit. So, for now, users must omit `['image']` from the `read_attrs`. It will simply always be read. We can postpone deliberation on how best to address this.

tacaswell:
This is also exactly #213 re the `read_attrs`

ophyd/areadetector/trigger_mixins.py
danielballan: I'm laughing because I was just criticizing @tacaswell for leaving docstrings out of his PR. I should have known better than to include docstrings. :-P

danielballan: Revised. Thanks for your attention to comprehensibility. :- D

ophyd/areadetector/trigger_mixins.py
```diff
@@ -71,41 +80,81 @@ class MultiTrigger(TriggerBase):
...
+    # 1. On the first trigger, close the shutter and acquire three images
+    # with different gain settings on the detector. Then open the shutter
+    # and take a light frame.
+    # 2. On the next trigger, just take a light frame.
+    # Repeat.
+    #
+    # Each element of this list specifies one acquisition. It gives a
+    # a label for each kind of image that will be taken and a dictionary
+    # mapping signals to values that must be set for that acquisition.
+    >>> dark_and_light = [('gain1', {'shutter': 'close', 'image_gain': 1}),
```
tacaswell: picky: ordered dict

tacaswell:
:+1: the logic makes sense to me

ophyd/areadetector/trigger_mixins.py
danielballan: Is the note at the end of the docstring enough?

tacaswell: no one reads the notes, they copy-paste examples

klauer: I somewhat agree, but `OrderedDict` is also really ugly to see in an example...

PR #217, opened 2016-01-14
MNT: reduce the default poll time on set_and_wait

tacaswell:
This causes about 3 extra times through the loop and could provide 10x
speed up.

### DIFF
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -344,7 +344,7 @@ def raise_if_disconnected(fcn):
     return wrapper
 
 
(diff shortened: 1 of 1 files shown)


klauer:
That leaves us no room to decrease the delay and effect major speed-ups in the future, though...

Kidding aside, this decrease may not actually be a good thing. AreaDetector can be a finicky beast regarding timing and toggling things too quickly.

tacaswell:
Can we inject that extra time only on the ad cases?

tacaswell:
@klauer said "Just merge it, if we have problems we will figure it out"

PR #221, opened 2016-01-18
Documentation, again

danielballan:
This supercedes #122. Read the build version [here](http://nsls-ii.github.io/ophyd/).

Still needed:

* Examples of using plugins with AD
* Documentation of the standard low-level (get, put) and high-level (read, describe) interfaces
* Documentation of read_attrs, configuration_attrs


tacaswell:
- can you put a table at the top of the interactive commands?
 - in positioners move the positioner entry above the pseudo pos docs so it is adjacent to the example
 - Signal needs mention that it can be used to inject python-only stuff into Devices
 - many plugins missing from inheritance diagram

danielballan:
I made the diagram non-comprehensive on purpose. (See the prose above the diagram.) Bad call?

doc/source/architecture.rst
```diff
@@ -0,0 +1,15 @@
+Project Architecture
+********************
+
+This is the class inheritance diagram for the key pieces of ophyd. This is
+a selection meant to give a readable and representative picture of the
+package's organization.
+
+.. inheritance-diagram:: ophyd.Device ophyd.Component ophyd.EpicsSignal ophyd.EpicsSignalRO ophyd.Signal ophyd.EpicsMotor ophyd.EpicsScaler ophyd.EpicsMCA ophyd.AreaDetector ophyd.HDF5Plugin ophyd.DynamicDeviceComponent ophyd.PVPositioner ophyd.SingleTrigger
```
tacaswell: You only need to list the leaves, it will build the rest of the graph.  Looks like there are knobs to change both size and direction of diagram, but no obvious way to touch them from rst...

tacaswell:
I think those diagrams are most useful when they are comprehensive.  We probably need to turn it to be up-down due to the depth of our tree.

doc/source/architecture.rst
danielballan: There are some like in `conf.py` which control the diagram's settings.

danielballan:
I experimented with left-to-right and top-down and found this way most readable. The knobs controlling font size did not work for me.

danielballan:
self-merging (docs only + a little API import that passes the tests)

remaining issues are not super quick, punting to a future issue -- could be awhile before we return to this

ericdill:
Your arrows are backwards for an inheritance diagram. Children point to their parents. [Example] (http://www.zentut.com/wp-content/uploads/2013/03/php-inheritance-class-diagram.png). I agree with Tom that it should probably be top-to-bottom instead of left-to-right. It is challenging to read on this smartphone

Docs and class diagrams are painful and tedious... Thanks for doing this!

danielballan:
Interesting. This would be an issue with sphinx then. Feel free to report it there and see what they say. Maybe there are different conventions in different subfields?

ericdill:
That is very strange then, because UML class diagrams are pretty clearly specified. Ohhhhh this is an "inheritance" diagram
Inheritance diagrams and Class diagrams have opposite arrows which is dumb

Issue #219, opened 2016-01-15
PVPositioner trigger method


klauer:
Placeholder for further discussion (@danielballan)

PR #223, opened 2016-01-19
MNT: if no trigger signal, assume no need

tacaswell:
If `self.trigger_signals` is an empty list, assume there is no need to
trigger the device (and it is immediatly done) instead of raising.

My proposal for NSLS-II/bluesky#279

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ before_install:
   - conda update conda --yes
   - conda install conda-build anaconda-client jinja2
   - conda config --add channels lightsource2
+  - conda config --add channels soft-matter
+
 
   # MAKE THE CONDA RECIPE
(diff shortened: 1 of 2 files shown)


danielballan:
Travis can't find tifffile. @ericdill was that removed from lightsource2 today?

danielballan:
Closes #219

danielballan:
@ericdill Looks like it was remove today. Why? I tried to fix myself but `anaconda copy` is still broken.

```
$ anaconda copy --to-owner lightsource2 soft-matter/tifffile/0.3.1
Using binstar api site https://api.anaconda.org
[ServerError] Copying anaconda packages temporarily disabled. This issue is being tracked here: https://github.com/Anaconda-Server/docs.anaconda.org/issues/119
```

PR #227, opened 2016-01-20
FIX: Device.read_configuration() was calling component.read()

klauer:


klauer:
(As mentioned in the above commit message)

Scaler test was failing due to `EpicsSignal.read_configuration` returning an empty dict, whereas it expected a dict including a value. I _think_ that this implies `Signal.read_configuration` should have both values and metadata (limits/EGU/etc)...  Devices, on the other hand, only respect `configuration_attrs`.

PR #228, opened 2016-01-20
FIX: plugin stage

klauer:
* Allow PluginBase to be used outside of Detectors (as can be necessary or desirable) - this comes in the form of just checking to see if parent has a 'cam' attr
* PEP8 cleanups on filestore/trigger_mixins
* Missing import on trigger_mixins shows our tests and coverage are lacking

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -15,6 +15,7 @@ import itertools
 
 from ..ophydobj import DeviceStatus
 from ..device import BlueskyInterface
+from ..utils import set_and_wait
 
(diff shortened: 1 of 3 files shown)


ophyd/areadetector/trigger_mixins.py
```diff
@@ -15,6 +15,7 @@ class MyDetector(PerkinElmerDetector, SingleTrigger):
 
 from ..ophydobj import DeviceStatus
 from ..device import BlueskyInterface
+from ..utils import set_and_wait
```
danielballan: Good catch. This missing import occurs in one of two lines that my test script did not touch.

PR #235, opened 2016-01-22
EpicsSignalRO refactor

klauer:
Should be a backward-compatible refactor

* Both EpicsSignal and EpicsSignalRO inherit from EpicsSignalBase. Some leftovers from Signal still being read-write, but I think it's an improvement. 97% coverage with tests (only trigger is untested)

* Additionally, odd ReadOnlyError bug should be fixed (thanks, @tacaswell). PVs cached and already connected on the pyepics level were calling callback functions prior to the EpicsSignal even getting their instance.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -492,7 +492,7 @@ class EpicsSignal(EpicsSignalBase):
         self._setpoint = value
         self._setpoint_ts = timestamp
 
-        if self._read_pv != self._write_pv:
+        if self._read_pv is not self._write_pv:
             self._run_subs(sub_type=self.SUB_SETPOINT,
                            old_value=old_value, value=value,
                            timestamp=self._setpoint_ts, **kwargs)
@@ -525,7 +525,7 @@ class EpicsSignal(EpicsSignalBase):
         old_value = self._setpoint
         self._setpoint = value
 
-        if self._read_pv == self._write_pv:
+        if self._read_pv is self._write_pv:
             # readback and setpoint PV are one in the same, so update the
             # readback as well
             super().put(value, timestamp=time.time(), force=True)
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -222,7 +222,9 @@ class EpicsSignalBase(Signal):
         self._read_pv = epics.PV(read_pv, form=pv_form,
                                  auto_monitor=auto_monitor,
                                  **pv_kw)
-        self._read_pv.add_callback(self._read_changed, run_now=True)
+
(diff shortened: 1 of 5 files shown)


klauer:
Stand by for fix - callback stuff still isn't exactly right causing the test failure. Will add better solution + test.

ophyd/signal.py
```diff
@@ -392,6 +305,201 @@ def get(self, as_string=None, **kwargs):
...
+    name : str, optional
+        Name of signal.  If not given defaults to read_pv
+    put_complete : bool, optional
+        Use put completion when writing the value
+    '''
+    def __init__(self, read_pv, write_pv=None, *, pv_kw=None,
+                 put_complete=False, string=False, limits=False,
+                 auto_monitor=None, name=None, **kwargs):
+
+        if 'rw' in kwargs:
```
tacaswell: this check could be moved up to the base-class init?  Given that this class is the default behavior, my guess would be that there are far more read-only signals with the `rw=False` kwarg.

ophyd/signal.py
```diff
@@ -183,17 +177,8 @@ def describe_configuration(self):
...
-
-    =======  =========  =====  ==========================================
-    read_pv  write_pv   rw     Result
-    =======  ========   ====   ==========================================
-    str      None       True   read_pv is used as write_pv
-    str      None       False  Read-only signal
-    str      str        True   Read from read_pv, write to write_pv
-    str      str        False  write_pv ignored.
-    =======  ========   ====   ==========================================
+class EpicsSignalBase(Signal):
```
tacaswell: *bikeshed* maybe name this `EpicsSignalRead` and get rid of `EpicsSignalRO` as that way the naming makes sense (`EpicsSignal` is a `EpicsSignalRead` vs `EpicsSignalRW` is a `EpicsSignalReadOnly` (wtf!) ) *\bikeshed*

ophyd/signal.py
```diff
@@ -392,6 +305,201 @@ def get(self, as_string=None, **kwargs):
...
+        '''Timestamp of setpoint PV, according to EPICS'''
+        return self._write_pv.timestamp
+
+    @property
+    def setpoint_pvname(self):
+        '''The setpoint PV name'''
+        return self._write_pv.pvname
+
+    def _repr_info(self):
+        yield from super()._repr_info()
```
tacaswell: we are going with we do not care about re-ordering this?

klauer:
After a lot of discussion with @tacaswell (and @dchabot, @danielballan) I believe this is good to go.

PR #241, opened 2016-01-23
ENH/API: Logical operations on StatusObject

danielballan:
Here's a start on an idea I had this morning.

Considering the `MergedStatusObject` at SRX -- which fires when both of its "children" are finished -- I thought it might be useful to be able to perform logical AND and OR on status objects, returning a `StatusBase`. With these, we can construct arbitrary combinations without creating a new class for every special case.

Currently, you can only register one callback function on a status object. For this to work, status objects must support multiple callbacks. If this isn't obvious, here's why:

```python
s3 = s1 | s2  # registers a callback on s1 that will fire s3
s4 = s1 | s3  # registers another callback on s1 to fire s4
```

Making this change is straightforward, but unfortunately I think it is not possible to do this with a backward-compatible API. Fortunately, the surface area of this API is not large. No non-test ophyd code (outside of `StatusBase` itself) is affected. About six lines of test code have to be changed, plus three lines downstream in bluesky and one place in the SRX beamline config.

This is still WIP -- just posting it for discussion.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -61,15 +61,19 @@ class StatusBase:
         finally:
             self._timeout_thread = None
 
-    @_locked
     def _finished(self, *, success=True, timestamp=None, **kwargs):
         # args/kwargs are not really used, but are passed - because pyepics
         # gives in a bunch of kwargs that we don't care about
-        self.success = success
-        self.done = True
-        if timestamp is None:
-            timestamp = time.time()
-        self.finish_ts = timestamp
+        with self._lock:
+            if self.done:
+                return
+
+            self.success = success
+            self.done = True
+            if timestamp is None:
+                timestamp = time.time()
+            self.finish_ts = timestamp
+
         for cb in list(self._callbacks):
             cb()
             self._callbacks.remove(cb)
@@ -92,7 +96,7 @@ class StatusBase:
 
     def __and__(self, other):
         """
-        Returns a new 'composite' status object, OrStatus,
+        Returns a new 'composite' status object, AndStatus,
         with the same base API.
 
         It will finish when both `self` or `other` finish.
@@ -120,21 +124,23 @@ class AndStatus(StatusBase):
 
         def inner():
             with self._lock:
-                l_success = self.left.success  # alias for readability below
-                r_success = self.right.success
+                with self.left._lock:
+                    with self.right._lock:
+                        l_success = self.left.success
+                        r_success = self.right.success
 
-                # At least one is done.
-                # If it failed, do not wait for the second one.
-                if (not l_success) and (l_success is not None):
-                    self._finished(success=False)
-                elif (not r_success) and (r_success is not None):
-                    self._finished(success=False)
+                        # At least one is done.
+                        # If it failed, do not wait for the second one.
+                        if (not l_success) and (l_success is not None):
+                            self._finished(success=False)
+                        elif (not r_success) and (r_success is not None):
+                            self._finished(success=False)
 
-                elif l_success and r_success:
-                    # both are done, successfully
-                    success = l_success and r_success
-                    self._finished(success=success)
-                # else one is done, successfully, and we wait for #2
+                        elif l_success and r_success:
+                            # both are done, successfully
+                            success = l_success and r_success
+                            self._finished(success=success)
+                        # else one is done, successfully, and we wait for #2
 
         self.left.add_callback(inner)
         self.right.add_callback(inner)
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -28,13 +28,18 @@ class StatusBase:
         The default timeout to use for a blocking wait, and the amount of time
         to wait to mark the operation as failed
     """
-    def __init__(self, *, timeout=None):
+    def __init__(self, *, timeout=None, timestamp=None):
         super().__init__()
         self._lock = RLock()
-        self._cb = None
+        self._callbacks = set()
         self.done = False
-        self.success = False
+        self.success = None
         self.timeout = None
+        self.finish_ts = None
+
+        if timestamp is None:
+            timestamp = time.time()
+        self.start_ts = timestamp
 
         if timeout is not None:
             self.timeout = float(timeout)
@@ -57,35 +62,85 @@ class StatusBase:
             self._timeout_thread = None
 
     @_locked
-    def _finished(self, success=True, **kwargs):
+    def _finished(self, *, success=True, timestamp=None, **kwargs):
         # args/kwargs are not really used, but are passed - because pyepics
         # gives in a bunch of kwargs that we don't care about
-
         self.success = success
         self.done = True
-
-        if self._cb is not None:
-            self._cb()
-            self._cb = None
+        if timestamp is None:
+            timestamp = time.time()
+        self.finish_ts = timestamp
+        for cb in list(self._callbacks):
+            cb()
+            self._callbacks.remove(cb)
 
     @property
-    def finished_cb(self):
+    def callbacks(self):
         """
-        Callback to be run when the status is marked as finished
+        Callbacks to be run when the status is marked as finished
 
         The call back has no arguments
         """
-        return self._cb
+        return self._callbacks
 
-    @finished_cb.setter
     @_locked
-    def finished_cb(self, cb):
-        if self._cb is not None:
-            raise RuntimeError("Can not change the call back")
+    def add_callback(self, cb):
         if self.done:
             cb()
         else:
-            self._cb = cb
+            self._callbacks.add(cb)
+
+    def __and__(self, other):
+        """
+        Returns a new 'composite' status object, OrStatus,
+        with the same base API.
+
+        It will finish when both `self` or `other` finish.
+        """
+        return AndStatus(self, other)
+
+    @property
+    def elapsed(self):
+        if self.finish_ts is None:
+            return time.time() - self.start_ts
+        else:
+            return self.finish_ts - self.start_ts
+
+    def __repr__(self):
+        return '{0}(done={1.done}, elapsed={1.elapsed:.1f}, ' \
+               'success={1.success})'.format(self.__class__.__name__, self)
+
+
+class AndStatus(StatusBase):
+    "a Status that has composes two other Status objects using logical and"
+    def __init__(self, left, right, *, timestamp=None, timeout=None):
+        super().__init__(timestamp=timestamp, timeout=timeout)
+        self.left = left
+        self.right = right
+
+        def inner():
+            with self._lock:
+                l_success = self.left.success  # alias for readability below
+                r_success = self.right.success
+
+                # At least one is done.
+                # If it failed, do not wait for the second one.
+                if (not l_success) and (l_success is not None):
+                    self._finished(success=False)
+                elif (not r_success) and (r_success is not None):
+                    self._finished(success=False)
+
+                elif l_success and r_success:
+                    # both are done, successfully
+                    success = l_success and r_success
+                    self._finished(success=success)
+                # else one is done, successfully, and we wait for #2
+
+        self.left.add_callback(inner)
+        self.right.add_callback(inner)
+
+    def __repr__(self):
+        return "({self.left!r} & {self.right!r})".format(self=self)
 
 
 class MoveStatus(StatusBase):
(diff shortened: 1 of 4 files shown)


klauer:
Cool and useful idea. Any thoughts about exception handling?

Every time we return to StatusObjects, it sure makes me wish we were using coroutines/Futures/asyncio. This would be so much nicer:
```python
def series():
    yield from motor1.move(1, wait=False)
    yield from motor2.move(2, wait=False)
    yield from motor3.move(3, wait=False)
```
or
```python
def parallel():
   tasks = [motor1.move(1), motor1.move(2), motor1.move(3)]
   loop.run_until_complete(asyncio.wait(tasks), timeout=2.0)
```

danielballan:
I guess that that is available in bluesky (kind of):

```python
def series(motors):
    for motor in motors
        yield from [Msg('set', motor), Msg('wait', motor)]

def parallel(motors):
    for motor in motors:
        yield from Msg('set', motor, block_group='A')
    for motor in motors:
        yield from Msg('wait', 'A')
```

But I agree that making this work at a lower level as well would be pretty. At some point, @tacaswell and I concluded that to *really* make pyepics work on an Event Loop -- with all the tools that asyncio provides to work with network sockets and take advantage of network latency -- you'd be rewriting not only pyepics but libca itself.

What error handling do you think would be appropriate here? If one of that Status objects raises, the scan is killed. Should that be different when they are composed?

tacaswell:
Adding hooks to let the run engine ask a Device for a plan to run on it's self might not be a bad idea.

```python

class DeviceWithAPlan(Device):
    ...

    def move_plan(self, position, parallel=True):
         if parallel:
            yield from parallel_motors(self.motors)
        else:
            yield from sereial_motors(self.motors)


ms = Msg('planned_move', dwap, pos, palallel=False)

class RunEngine:
    @asycio.coroutine
    def _planned_move(self, obj, *args, **kwarg):
        new_plan = iter(obj.move_plan(*args, **kwargs)
        self._gen_stack.appendleft(new_plan)
        return None
```

This also means that if you are part of the way through a move and it gets interrupted it _should_ be replayed correctly which is kinda cool.

klauer:
I disagree with the conclusion you came regarding the need to rewrite libca, but this isn't the thread for that sort of discussion. As to writing Devices with the concept of plans, we don't want dependencies from either bluesky to ophyd or vice-versa.

Anyway, back on topic: regarding my exception comment, I guess there's not much you can do to make it easier to figure out what was going on during the callbacks, so never mind.

tacaswell:
On further thought adding a `run_sub_plan` message to the default set of things the RunEngine knows how to do might be cute.  On the other hand, if it does not know to look for said plans, then your user plans would be


```python
def my_plan(obj):
    g = obj.move_plan()
    yield Message('sub_plan', g)
```

where you could also just do

```python
def my_plan(obj):
    yield from obj.move_plan()
```

which is much simpler and does not involve touching the generator stack.

tacaswell:
@danielballan We should start a 'plan_snippet' module for things like the parallel and serial moves you wrote above.

In 3.5 you can yield from inside of a try-except

```python
plan = list('abced')


def broken_plan():
    yield from plan
    raise RuntimeError
    yield from plan
    
def run_contex(plan):
    yield 'start'
    try:
        yield from plan
    finally:
        yield 'stop'
```

Which makes me think we can provide a `run` and `event` decorators so one could write

```python
@event_decorator
def collect_single_point(det_list):
    bg_uid = str(uuid4())
    for d in det_list:
        yield Msg('trigger', d, block_group=bg_uid)
    yield Msg('wait_for', bg_uid)
    rets = []
    for d in det_list:
        ret = yield Msg('read', d)
        rets.append(ret)
    return rets

def read_diode(diode):
    yield Msg('trigger', diode)
    ret = yield Msg('read', diode)
    return ret

@run_decorator   
def do_scan(some_args, motors, dets):
    for p in f(some_args):
         yield from move_motors(motors, p)
         yield from collect_single_point(motors + dets)

def master_plan(sample_id, diode, some_args, motors, dets):
    for s is sample_id:
        try:
            pre = yield from read_diode(diode)
            yield from load_sample(s)
            post = yield from read_diode(diode)
            if pre['data']['diode'] - post['data']['diode'] > thresh:
                yield from do_scan(some_args, motors, dets)
        finally:
            yield from unload_sample()
             
```

tacaswell:
well, I lead the comments in this PR an a wild goose hunt.

This needs a companion PR in bluesky?

tacaswell:
I think this also needs a test of at least 2 generations of the merged status objects.  As we learned with slicerator, sometimes issues do not show up until you go that deep.

danielballan:
Noted -- will add tests later today. In the meantime, this is the companion PR in bluesky: https://github.com/NSLS-II/bluesky/pull/291/files

danielballan:
This is ready for a last round of review.

ophyd/ophydobj.py
```diff
@@ -52,28 +54,55 @@ def _finished(self, *args, **kwargs):
...
-        if self._cb is not None:
-            raise RuntimeError("Can not change the call back")
+    def add_callback(self, cb):
         if self.done:
             cb()
         else:
-            self._cb = cb
+            self._callbacks.add(cb)
+
+    def __and__(self, other):
```
klauer: Docstrings sure would be nice for the logical operations.

ophyd/ophydobj.py
```diff
@@ -52,28 +54,55 @@ def _finished(self, *args, **kwargs):
...
+            with composite._lock:
+                composite._done_count += 1
+                if composite._done_count == 2:
+                    composite._finished()
+
+        self.add_callback(inner)
+        other.add_callback(inner)
+        return composite
+
+    def __or__(self, other):
```
klauer: Creating the new composite object should lock both `self` and `other`, otherwise one may complete while the composite is being generated

ophyd/ophydobj.py
```diff
@@ -52,28 +54,55 @@ def _finished(self, *args, **kwargs):
...
+            self._callbacks.add(cb)
+
+    def __and__(self, other):
+        composite = StatusBase()
+        composite.ancestors |= set([self, other])
+        composite.ancestors |= (self.ancestors | other.ancestors)
+
+        def inner():
+            with composite._lock:
+                composite._done_count += 1
```
klauer: Why do `_done_count` and `ancestors` live in StatusBase? Isn't that something that should be in some `CompositeStatus` class? Currently, the result of a `(MoveStatus() | MoveStatus())` would give you a `StatusBase` as a result...

ophyd/ophydobj.py
```diff
@@ -38,9 +38,11 @@ class StatusBase:
     def __init__(self):
         super().__init__()
         self._lock = RLock()
-        self._cb = None
+        self._callbacks = set()
```
klauer: Would the order of the callback firing ever be something desirable to control?

klauer:
Can you also nuke `DetectorStatus` while you're at it?

ophyd/ophydobj.py
klauer: Additionally, even with the locking, checks should be done to see if `self` or `other` is already done and set the appropriate status on the composite object (along with anything else necessary)

ophyd/ophydobj.py
tacaswell: :imp:  Using a set also prevents having the same callback called twice.  Not clear if this is a feature or a bug.

ophyd/ophydobj.py
tacaswell: :+1: to this so long as it does not cause too much pain.  I _think_ it should go cleanly.

ophyd/ophydobj.py
tacaswell: This is taken care of by the `add_callback` which runs the callback immediately and is mutually locked with setting the status to done.

danielballan:
> @danielballan, can you point out an example of using this sort of thing in the wild?

@klauer helpfully provided an example [here](https://github.com/NSLS-II/ophyd/pull/268#discussion_r53085766)

This is ready for another round of review. All changes since last review are in the latest commit. See message for details.

danielballan:
Rebased (basically rewritten) in the wake of the positioner refactor. I had to make some off-the-cuff decisions that might not be the right ones, and I'm not attached. Let's have a healthy discussion here. Major points:

* `AndStatus` succeeds if both its parents succeed.
* If the first parent to finish fails, the `AndStatus` finishes early with failure. (Is this desired behavior?)
* `OrStatus` is gone. It's unclear how to handle merging `success` in that case, and it's hard to think of a reasonable use case. If one comes up with can revisit this.
* `success` is initialized as `None` (not `False`) until the status object finishes. This is not a philosophical change, but a practical one -- I use it make `AndStatus` work.
* As before, the `timestamp` and `elapsed` logic move into the base class.
* Minor point: `_finished` takes a kwarg called 'timestamp' to set the `finish_ts` property. Now `__init__` works the same way to set the `start_ts` property.
* More arguments are made keyword-only arguments.

ophyd/status.py
```diff
@@ -55,35 +60,85 @@ def _timeout_thread(self):
...
+        return AndStatus(self, other)
+
+    @property
+    def elapsed(self):
+        if self.finish_ts is None:
+            return time.time() - self.start_ts
+        else:
+            return self.finish_ts - self.start_ts
+
+    def __repr__(self):
```
tacaswell: :+1: to repr on base class

ophyd/status.py
```diff
@@ -55,35 +60,85 @@ def _timeout_thread(self):
...
+
+
+class AndStatus(StatusBase):
+    "a Status that has composes two other Status objects using logical and"
+    def __init__(self, left, right, *, timestamp=None, timeout=None):
+        super().__init__(timestamp=timestamp, timeout=timeout)
+        self.left = left
+        self.right = right
+
+        def inner():
```
tacaswell: Probably need to use all three locks.....which means we can deadlock!

ophyd/status.py
```diff
@@ -55,35 +60,85 @@ def _timeout_thread(self):
...
+    def add_callback(self, cb):
         if self.done:
             cb()
         else:
-            self._cb = cb
+            self._callbacks.add(cb)
+
+    def __and__(self, other):
+        """
+        Returns a new 'composite' status object, OrStatus,
```
klauer: Docstring mismatch (for previous `OrStatus` implementation)

ophyd/status.py
klauer: Though the locks are `RLocks`, you're right, deadlock is a possibility here. A deadlock condition would depend on where left/right `finished` callbacks comes from. 

Let's see, if these two things happen around the same time:
1. Finished callback 1 comes from libca -> filtered through ophyd `MonitorDispatcher` to its Python thread -> status1._finished()
2. Finished callback 2 comes from some other Python thread -> status2._finished()

`status1` and `status2` run their callbacks simultaneously which correspond to `inner()` here. Assuming we attempt to lock both, that would be a deadlock.

OK, so it really depends on the callback implementation...

ophyd/status.py
```diff
@@ -55,35 +60,85 @@ def _timeout_thread(self):
             self._timeout_thread = None
 
     @_locked
-    def _finished(self, success=True, **kwargs):
+    def _finished(self, *, success=True, timestamp=None, **kwargs):
```
klauer: I think not locking during the callbacks could help avoid some deadlock conditions:

```python
def _finished(self, *, success=True, timestamp=None, **kwargs):
    # args/kwargs are not really used, but are passed - because pyepics
    # gives in a bunch of kwargs that we don't care about
    with self._lock:
        if self.done:
            return

        self.success = success
        self.done = True
        if timestamp is None:
            timestamp = time.time()
        self.finish_ts = timestamp

    for cb in list(self._callbacks):
        cb()
        self._callbacks.remove(cb)
```

danielballan: That looks right. I'll put this in.

danielballan:
Travis is locking up, reproducibly, on the first move-related test. I seem to have broken this in my refactor. Can someone with a working ophyd docker take a quick look and give me a hint?

klauer:
The issue is that `finished_cb` was nuked and thus the api broke - all calls to that setter are now just setting an unused attribute on the status object instead of adding the callable argument it to the callback list.

klauer:
Sorry again that the internals keep changing requiring re-re-refactoring of this PR. We had to prioritize getting the things with immediate need in first.

PR #242, opened 2016-01-26
ENH: Add duck-typing namespace scraper.

danielballan:
This device does not get collected by `get_all_positioners`:

```python
class XYMotor(Device):
    x = Cpt(EpicsMotor, ...)
    y = Cpt(EpicsMotor, ...)

Should we add a `set` method and use duck-typing to identify positioners? See attached function.

### DIFF
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -116,7 +116,7 @@ def _normalize_positioners(positioners):
         res = []
         for device in positioners:
             if not isinstance(device, (Device, Positioner)):
-                raise TypeError("input is not a device")
+                raise TypeError("Input is not a Device: %r" % device)
             res.extend(_recursive_positioner_search(device))
     return res
 
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -42,6 +42,22 @@ FMT_PREC = 6
 DISCONNECTED = 'disconnected'
 
 
+def scrape_namespace():
+    """
+    Get all public objects from the user namespace, sorted by name.
+
+    If we are not in an IPython session, warn and return an empty list.
+    """
+    ip = IPython.get_ipython()
+    if ip is None:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ophyd/commands.py [hunk 1]
```diff
@@ -61,6 +61,26 @@ def instances_from_namespace(classes):
                 if isinstance(val, classes) and not var.startswith('_')]
 
 
+def ducks_from_namespace(attr):
```
tacaswell: This should take *args of attr names (with implicit `&` logic)

ophyd/commands.py
```diff
@@ -392,17 +443,16 @@ def log_pos(positioners=None, extra_msg=None):
...
-    if positioners is not None:
-        msg += logbook_add_objects(positioners)
+    msg += logbook_add_objects(positioners)
 
-        for p in positioners:
-            try:
-                pdict['values'][p.name] = p.position
-            except DisconnectedError:
-                pdict['values'][p.name] = DISCONNECTED
+    for p in positioners:
```
tacaswell: This needs the `None` guard added back.

ophyd/commands.py
```diff
@@ -51,19 +67,54 @@ def instances_from_namespace(classes):
...
+    "input normalization used by wh_pos, log_pos, log_pos_mov"
+    if positioners is None:
+        # Grab IPython namespace, recursively find Positioners.
+        positioners = get_all_positioners()
+    elif isinstance(positioners, (Device, Positioner)):
+        # Explore children in case this is a composite Device.
+        positioners = _recursive_positioner_search(positioners)
+    else:
+        # Assume this is a list of Devices.
+        positioners = []
```
tacaswell: I think there is a line missing here?

ophyd/commands.py
danielballan: The issue was upstream.

PR #244, opened 2016-01-27
FIX: more careful recursion that works on PseudoSingle

danielballan:
`wh_pos()` died when it tries to call `_signals` on a PseudoSingle. We should only be recursing through Devices, not Positioners.

All of this will go away later -- just trying to make something reasonably clean that works for this cycle.

### DIFF
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -94,12 +94,14 @@ def get_all_positioners():
 
 def _recursive_positioner_search(device):
     "Return a flat list the device and any subdevices that can be 'set'."
+    # TODO Refactor this as a method on Device.
     res = []
-    if hasattr(device, 'set'):
+    if hasattr(device, 'position'):  # duck-typed as a Positioner
         res.append(device)
-    for d in device._signals.values():
-        if isinstance(d, (Device, Positioner)):
-            res.extend(_recursive_positioner_search(d))
+    if isinstance(device, Device):  # only Devices have `_signals`
+        for d in device._signals.values():
+            if isinstance(d, (Device, Positioner)):
+                res.extend(_recursive_positioner_search(d))
     return res
 
 


danielballan:
Sorry I didn't mean that to be cryptic or worse ominous. At the request of @klauer it may be refactored into Device -- ideally we shouldn't be touching _signals like this outside the class. See the TODO comment added in this PR.

On Tue, Jan 26, 2016 at 5:30 PM -0800, "dchabot" <notifications@github.com<mailto:notifications@github.com>> wrote:


ok, @danielballan<https://github.com/danielballan>, I'll bite: why will this "go away later"?

danielballan:
In a disappointing and just utterly boring turn of events, even *this* implementation does not quite work. It does not raise, but for some reason it is missing some devices. Can anyone else take a crack at this?

klauer:
Can you give an example of where this fails?

danielballan:
```python
class Thing(Device):
    m = Cpt(EpicsMotor, '...')
```

`m` does not show up.

klauer:
```python
from ophyd import Device, Component as Cpt, EpicsMotor
from ophyd.commands import get_all_positioners

class Thing(Device):
    m = Cpt(EpicsMotor, 'XF:31IDA-OP{Tbl-Ax:X1}Mtr')

t = Thing('')

```

```python
$ ipython -i dev_test.py
Python 3.4.4 |Anaconda 2.4.1 (64-bit)| (default, Jan 11 2016, 13:54:01)
In [1]: get_all_positioners()
Out[1]: [EpicsMotor(prefix='XF:31IDA-OP{Tbl-Ax:X1}Mtr', name='_m', parent='', egu='', timeout=0.0, read_attrs=['user_readback', 'user_setpoint'], configuration_attrs=['motor_egu'], monitor_attrs=[], settle_time=0.05)] ```

danielballan:
False alarm. I thought ophyd had already been repackaged with this PR, but it had now. I was looking at the old `wh_pos`. :sheep:

danielballan:
confusing typo: it had *not*

klauer:
Good to hear!

PR #247, opened 2016-01-27
Fix: order issues in unstage

tacaswell:
- _orginals_vals needs to be an ordered dict
 - unstage child devices (in reverse order) before unstaging self


klauer:
Looks good to me; waiting on travis

tacaswell:
Looks like it is choking the the ophyd.commands tests (which pass for me locally).  Added a flag to make the tests more verbose to see which of the `mov` commands is breaking.

danielballan:
Neither @tacaswell or I understand that intermittent fail. Leaving this verbose for now is probably a virtue, my 4G data plan notwithstanding.

PR #248, opened 2016-01-27
MNT: default to not auto-monitoring

### LINKED ISSUE #246: need a way to make sure AD does not burn all the CPU monitoring diagnostic cams
With Diag6 spinning at CSX ipython burns 102% of a cpu in the background, with the cam stopped it uses ~5%.   With it spinning the ipython prompt is laggy and super annoying to use.

@klauer @dchabot

tacaswell:
Be a bit more conservative about setting up monitors.

closes #246

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -166,7 +166,7 @@ class EpicsSignalBase(Signal):
     def __init__(self, read_pv, *,
                  pv_kw=None,
                  string=False,
-                 auto_monitor=None,
+                 auto_monitor=False,
                  name=None,
                  **kwargs):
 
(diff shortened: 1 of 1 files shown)


klauer:
@dchabot it definitely is. Things that use subscriptions rely on the monitor being in place. @tacaswell and I talked about making things always `auto_monitor=False` until someone subscribes to its value, but have yet to figure out a nice way to do it (re-create the `epics.PV`?).

klauer:
Agreed - it is more simple, but it'll take expert discrimination as you say and could be more prone to mistakes.

klauer:
That may help. Still, when you have a beamline config with 10k PVs and a single monitor callback thread - any amount of unnecessary monitoring is going to be costly.

klauer:
Well, not quite. I'm suggesting only 'auto monitor' if the signal has been `ophydobj.subscribe()`'d to.

The way `EpicsSignal.get()` currently works is to ignore the monitored value. The reason for that is because the value returned can change depending on the kwargs you pass in -- with `as_string`, `count`, and `as_numpy`. So, (I think) the only place a monitored value becomes important is for subscriptions (monitor callbacks).

PR #249, opened 2016-01-28
FIX: re-initialize automonitor when an EpicsSignal subscription is made

### LINKED ISSUE #246: need a way to make sure AD does not burn all the CPU monitoring diagnostic cams

klauer:
Replaces #248 
Should fix #246 

Tests passing locally, but we should make extra sure the logic is correct for re-initialization of `epics.PV`s

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -106,11 +106,17 @@ class EpicsMotor(Device, Positioner):
         self._started_moving = False
 
         try:
-            self.user_setpoint.put(position, wait=wait)
-            return super().move(position, wait=wait, **kwargs)
+            if not wait:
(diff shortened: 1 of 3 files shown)


tacaswell:
do we want to backport this to 0.2.x?  The 'fix' at CSX was to just not include the stats plugins on those cameras.

klauer:
It should be tested more extensively at a beamline prior to backporting. If you could give it a go at CSX including the stats plugins, that would be good, too.

It worked in tests and could offer very significant performance improvements to beamlines with a large number of PVs.

PR #251, opened 2016-01-29
Hotfix AD again

danielballan:


danielballan:
For anyone reading along, the key fix here is that file name/path/number have to be set *before* capture mode is switched on.

Also, different beamlines need us to look in different places to know how many image planes go into one datum ("frames per point"). Examples are `num_captured`, `num_images`, `number_of_sets`. We now access this through a simple method that can be overriden by a subclass.

PR #253, opened 2016-01-30
More stage debug

danielballan:
- debug messages giving old and new values of stage_sigs as they are set
- one more subtlety with stage/unstage logic: a device should not be marked `_staged = False` until unstaging has successfully finished so that you can't stage until everything is unstaged

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -28,7 +28,7 @@ from datetime import datetime
 from collections import defaultdict
 from itertools import count
 
-from ..device import GenerateDatumInterface, BlueskyInterface
+from ..device import GenerateDatumInterface, BlueskyInterface, Staged
 from ..utils import set_and_wait
 
 logger = logging.getLogger(__name__)
@@ -124,7 +124,7 @@ class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
     def describe(self):
         # One object has been 'described' once, no new keys can be added
         # during this stage/unstage cycle.
-        self._locked_key_list = self._staged
+        self._locked_key_list = (self._staged == Staged.yes)
         res = super().describe()
         for k in self._datum_uids:
             res[k] = self.parent.make_data_key()  # this is on DetectorBase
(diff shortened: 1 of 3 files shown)


danielballan:
Tri-state flag with staged, unstaged, and partially staged?

danielballan:
This is ready for another round of review. Once a device is "partially staged", it must be successfully unstaged before it can be staged. Triggers expect a device to be fully staged. Three states are indeed necessary to capture this.

ophyd/areadetector/trigger_mixins.py
```diff
@@ -142,7 +142,7 @@ def trigger_cycle(self, val):
 
     def trigger(self):
         "Trigger one or more acquisitions."
-        if not self._staged:
+        if not (self._staged == Staged.yes):
```
klauer: `!=`?

ophyd/device.py
```diff
@@ -341,10 +349,16 @@ def describe(self):
 
     def stage(self):
         "Prepare the device to be triggered."
-        if self._staged:
-            raise RuntimeError("Device is already stage. Unstage it first.")
-        self._staged = True
+        if self._staged == Staged.no:
+            pass  # to short-circuit checking individual cases
```
klauer: Why is short-circuiting here necessary?

danielballan: First check if there is problem and only then invest the cycles in diagnosing exactly *which* problem we have.

ophyd/areadetector/trigger_mixins.py
danielballan: Ha. Fair.

ophyd/device.py
klauer: So a placeholder?

ophyd/areadetector/filestore_mixins.py
```diff
@@ -124,7 +124,7 @@ def generate_datum(self, key, timestamp):
     def describe(self):
         # One object has been 'described' once, no new keys can be added
         # during this stage/unstage cycle.
-        self._locked_key_list = self._staged
+        self._locked_key_list = (self._staged == Staged.yes)
```
tacaswell: not to do in this PR, but did we change our mind that this was too clever or not?

tacaswell:
:+1:

ophyd/device.py
danielballan: My response got lost in the ether of intermittent cell service:

I wouldn't call it placeholder. It optimizes the case where everything is normal (making one comparison instead of two) at the cost of slowing down the case where we have to raise an error (making two or three comparisons instead of one or two).

klauer: Allow me to leave this here: https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize :wink:

PR #258, opened 2016-02-05
WIP: Positioner describe

dchabot:
The `Positioner` base-class was missing `read/describe` methods.

I added these so that `PseudoSingle` (inheriting from `Positioner`), as a `Component` of a `PseudoPositioner`, would contribute to event-document content when called upon.

### DIFF
--- a/ophyd/positioner.py
+++ b/ophyd/positioner.py
@@ -12,6 +12,7 @@ import logging
 import time
 
 from .utils import TimeoutError
+from .utils.epics_pvs import (data_type, data_shape)
 from .ophydobj import (MoveStatus, OphydObject)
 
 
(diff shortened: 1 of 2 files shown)


tacaswell:
This may cause issues with the MI with `Device` as now not all of the implementations of these methods are cooperative.

klauer:
@tacaswell it may cause MI issues as you say - but does it in the current hierarchy?

klauer:
When the refactor was initially being done, everyone was against `Positioner` being a `Device` (as it has no components of its own and should be setup through MI). For whatever reason opinions have shifted. 

We should be working toward the next refactor where `Positioner` is a `Device` and can play nicely in the class hierarchy.

tacaswell:
I am :+1: on `PositionerBase` as @dchabot says.

danielballan:
Agreed.

klauer:
I'm stealing this implementation for #271, sorry @dchabot...

PR #259, opened 2016-02-09
FIX: wait for setpoint record to process

tacaswell:
Before trying to actuate the positioner, wait for the
setpoint to be set.

### DIFF
--- a/ophyd/pv_positioner.py
+++ b/ophyd/pv_positioner.py
@@ -161,7 +161,7 @@ class PVPositioner(Device, Positioner):
     def _move_async(self, position, **kwargs):
         '''Move and do not wait until motion is complete (asynchronous)'''
         if self.actuate is not None:
-            self.setpoint.put(position, wait=False)
+            self.setpoint.put(position, wait=True)
             self.actuate.put(self.actuate_value, wait=False)
         else:
             self.setpoint.put(position, wait=False)


ophyd/pv_positioner.py
```diff
@@ -161,7 +161,7 @@ def _move_wait(self, position, **kwargs):
     def _move_async(self, position, **kwargs):
         '''Move and do not wait until motion is complete (asynchronous)'''
         if self.actuate is not None:
-            self.setpoint.put(position, wait=False)
+            self.setpoint.put(position, wait=True)
```
tacaswell: maybe this should be set_and_wait?

klauer:
That will cause what's supposed to be an async move to block if setpoint has any sort of put completion support.

ophyd/pv_positioner.py
klauer: `set_and_wait` is probably a better solution.

tacaswell:
Short of sub-classing the EpicsSignal used for the setpoint, I do not see where else to address this.

danielballan:
@tacaswell @klauer Is now a good time to mention that `set_and_wait` is evidence of a leaky abstraction, and that anything that has to be `set_and_wait` should probably be `EpicsSignalThatCantBeTrustedToKnowWhenItIsDone`? I think the logic of `set_and_wait` should live inside `put`.

danielballan:
Sorry @dchabot tried to tag you there and failed. Seems like we are on the same wavelength.

tacaswell:
Dealt with this by sub-classing PV positioner in the SRX config.

PR #261, opened 2016-02-10
MNT: hook for filestore mixin make_filename

klauer:
Need to be able to:

1. Modify the filenames being saved through FileStoreBase (NFS+PIMS choke with hundreds of thousands of files in one directory)
2. Create the directory and set permissions

With this small change, I can do both in my subclass. For (1) I'm making a subdirectory with the start of the filename uid since I don't have access to any scan information during `stage()`.

P.S. @danielballan here's an opportunity to disagree!

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -126,12 +126,31 @@ class FileStorePluginBase(FileStoreBase):
                                 (self.num_capture, 0),
                                 ])
 
-    def stage(self):
-        # Make a filename.
+    def make_filename(self):
+        '''Make a filename.
+
+        This is a hook so that the read and write paths can either be modified
+        or created on disk prior to configuring the areaDetector plugin.
+
+        Returns
+        -------
+        filename : str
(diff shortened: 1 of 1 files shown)


tacaswell:
Is there an obvious way to inject information (like the runstart uid) into stage?  One way would be a `configure` message to the AD before the `stage`?  Another would be to start allowing stage to take arguments.

tacaswell:
Other wise :+1:

klauer:
I do other per-run configuration-related things in detector `set` methods. That's the only place in the current design it'll go as far as I can tell.

tacaswell:
I think you should be doing that through `configure`, if you `set` a detector in the middle of a run you can make the descriptor lie about it's settings.

klauer:
Well, (`un`)`stage` used to be (`de`)`configure`, so `set` was the only option... I suppose you're correct post-refactor - my old `set` should become `configure`.

tacaswell:
Ah, sorry I lost track of which version of ophyd you ar eworking against.

PR #262, opened 2016-02-12
FIX: minor mca tweaks

klauer:
* Split mca record class into just the EPICS mca record and the other one with additional Start and EraseStart records (@dchabot thoughts about naming this better, or if those separate records belong on the subclassed `Device`s instead?)
* Can't have namedtuple keys with leading underscores - as such, no components with leading underscores ( :( )

Use case is the (mca-mode) Struck SIS3820 scaler, used on the vast majority of beamlines: (config ref [struck_scaler.py](https://github.com/klauer/hxntools/blob/84c8f2831b23475e229d206ba16b49eebc3b31d6/hxntools/struck_scaler.py))

### DIFF
--- a/ophyd/mca.py
+++ b/ophyd/mca.py
@@ -68,7 +68,7 @@ class EpicsMCARecord(Device):
     background = C(EpicsSignalRO, '.BG')
     mode = C(EpicsSignal, '.MODE', string=True)
 
-    rois = DDC(add_rois(range(0, 31)))
+    rois = DDC(add_rois(range(0, 32)))
 
     def __init__(self, prefix, *, read_attrs=None,
                  configuration_attrs=None, monitor_attrs=None, name=None,
--- a/ophyd/mca.py
+++ b/ophyd/mca.py
@@ -62,8 +62,11 @@ def add_rois(range_, **kwargs):
 class EpicsMCARecord(Device):
     '''SynApps MCA Record interface'''
     stop_signal = C(EpicsSignal, '.STOP')
-    preset_real_time = C(EpicsSignal, '.ERTM', write_pv='.PRTM')
-    preset_live_time = C(EpicsSignal, '.ELTM', write_pv='.PLTM')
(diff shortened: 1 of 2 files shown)


klauer:
Let's wait on @dchabot getting back before we merge - I think we should chat about naming first.

ophyd/mca.py
```diff
@@ -96,7 +96,12 @@ def __init__(self, prefix, *, read_attrs=None,
         self.stage_sigs.update([(self.mode, 'PHA')])
 
     def stop(self):
-        self._stop.put(1)
+        self.stop_signal.put(1)
+
+
+class EpicsMCA(EpicsMCARecord):
```
klauer: Bumping this because I keep forgetting to bring it up in person.

@dchabot any thoughts about the naming or logical break-up of these things?

ophyd/mca.py
```diff
@@ -59,26 +59,26 @@ def add_rois(range_, **kwargs):
...
-
-    _stop = C(EpicsSignal, '.STOP')
-    preset_real_time = C(EpicsSignal, '.ERTM', write_pv='.PRTM')
-    preset_live_time = C(EpicsSignal, '.ELTM', write_pv='.PLTM')
     spectrum = C(EpicsSignalRO, '.VAL')
     background = C(EpicsSignalRO, '.BG')
     mode = C(EpicsSignal, '.MODE', string=True)
 
-    rois = DDC(add_rois(range(0, 31)))
+    rois = DDC(add_rois(range(0, 32)))
```
dchabot: Thanks for catching that off-by-one.

klauer:
My only real concern was the `EpicsMCARecord` and `EpicsMCA` stuff, but if you're OK with that then I'd say it's ready.

klauer:
Note the branch it was pulled into: https://github.com/NSLS-II/metadatastore/pull/195

tacaswell:
It also looks like bluesky has a layer of numpy normalization.  That needs some attention 'later'.

PR #263, opened 2016-02-12
ENH: Derived signal

klauer:
The concept of a 'derived signal' with a value and description based on another.

Submitting to see what people think - not saying this has to go in or necessarily even should. 

As an example, I've used this in an ROI's low and high eV `Component`s, which calculate user-friendly eV units from bin numbers.:

```python
class EvSignal(DerivedSignal):
    '''A signal that converts a bin number into electron volts'''
    def __init__(self, parent_attr, *, parent=None, **kwargs):
        bin_signal = getattr(parent, parent_attr)
        super().__init__(derived_from=bin_signal, parent=parent, **kwargs)

    def get(self, **kwargs):
        bin_ = super().get(**kwargs)
        return bin_to_ev(bin_)

    def put(self, ev_value, **kwargs):
        bin_value = ev_to_bin(ev_value)
        return super().put(bin_value, **kwargs)

    def describe(self):
        desc = super().describe()
        desc[self.name]['units'] = 'eV'
        return desc
```
Note that this inherits from DerivedSignal and gets a component from its parent, so in a way this example is `DerivedFromSibling`, I suppose.

Pluses
* Stands out from other signals - `isinstance()` indicates it's a soft signal based on another one
* May save a couple lines of code here or there
* `describe` could do something special to indicate from where it was derived

Minuses
* How's it any different from just passing in the `Signal` and working with it in any other class?
* More classes = more confusion
* Discussion, blah

### DIFF
--- a/ophyd/__init__.py
+++ b/ophyd/__init__.py
@@ -7,7 +7,7 @@ from . import *
 
 
 # Signals
-from .signal import (Signal, EpicsSignal, EpicsSignalRO)
+from .signal import (Signal, EpicsSignal, EpicsSignalRO, DerivedSignal)
 
 # Positioners
 from .positioner import (PositionerBase, SoftPositioner)
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -198,6 +198,11 @@ class DerivedSignal(Signal):
         '''Mirrors the connection state of the original signal'''
         return self._derived_from.connected
 
+    @property
+    def limits(self):
+        '''Limits from the original signal'''
+        return self._derived_from.limits
+
     def _repr_info(self):
         yield from super()._repr_info()
         yield ('derived_from', self._derived_from)
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -146,6 +146,26 @@ class Signal(OphydObject):
         return self.limits[1]
 
 
+class DerivedSignal(Signal):
+    '''A signal which is derived from another one'''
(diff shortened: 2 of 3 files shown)


danielballan:
I like this. I went back for it when you shared it with my over GChat but forget to say anything -- sorry. I think it is a useful concept to add.

klauer:
Thanks @danielballan. Anyone else disagree (or feel strongly either way)?

klauer:
@dchabot no valid reason, except for the Xspress3 IOC being a nightmare to work with. :) I definitely agree that it should live there. 

Not quite sure where to draw the line saying "you must do this in an IOC" or "this is okay on the Python level".

This example is definitely trivial. However, I think the benefit of having scipy/numpy/devices/etc at one's disposal in Signal value calculations would be of benefit to someone.

tacaswell:
This also provides a way to do read-only pseudo axes.

This could also be super useful as a prototyping tool.  If someone has a 'great idea' for a new transformed PV you can hack it up in python in 10min and if it is actually useful push it down to an IOC and if not; you only spent 10min on it :smiling_imp:.

danielballan:
Right. Refusing to implement something in Python on principle is like adopting an engineering control to solve a problem that calls for an administrative control.

danielballan:
@dchabot Fair.

klauer:
Fleshed this out a bit more, rebased, and added tests. Leaving it up to others to decide if it belongs or not.

PR #267, opened 2016-02-16
ENH: Adding home and jog pvs to EpicsMotor

hhslepicka:


hhslepicka:
Sorry for my misunderstanding, @dchabot . I thought that Bluesky was the responsible for Data Collection and Ophyd was a representation of hardware in Python.

Here is a question that I have: How in the current model do we check a motor limit and home switch status?

hhslepicka:
@dchabot Sure! Just let me know when and where! :+1: 

In which moment the movement is checked against the soft limit values? Is it done automatically prior to all movement requests?

But IMHO, as Ophyd is described as an representation of the hardware in python, I think that we should provide as many as possible controls of this hardware for the user, maybe use some kind of on-demand load of PVs to avoid unnecessary resources.

klauer:
These are useful to have - and others will probably ask for them at some point. 

Homing and jogging should be returning `MoveStatus` objects so that their status can be checked asynchronously. Before we get more into those details, though, let's try to come to a compromise as to how to best represent such functionality of the motor record (+ tpmac axes, which are on every beamline) on the Python level (and still have it work well in the whole `Positioner` hierarchy).

hhslepicka:
Based on the yesterday meeting, I'm closing this PR and I will open an issue with the comments/definitions on what we think that should be implemented.
Thank you all for the input.

PR #268, opened 2016-02-16
FIX: pseudo positioners, again

klauer:
* Complete round of tests post-ophyd0 refactor (at least coverage is up to 90% or so)
* Add describe(), read() on pseudosingle (from @tacaswell 's srx config)
* Fix for some namedtuple-related bugs (I consistently get  `namedtuple(a, b, c)` works the same as `tuple((a, b, c))`)
* Fix target position for PseudoSingle
* Fix check_value for PseudoPositioner
* Refactor move, breaking into sequential and concurrent cases

### DIFF
--- a/ophyd/pseudopos.py
+++ b/ophyd/pseudopos.py
@@ -10,10 +10,10 @@
 
 import logging
 import time
+import threading
+import collections
 
 from collections import (OrderedDict, namedtuple)
-
-import numpy as np
 
 from .utils import (TimeoutError, DisconnectedError)
 from .positioner import Positioner
@@ -77,7 +77,7 @@ class PseudoSingle(Positioner):
         self._target = None
 
     def check_value(self, pos):
-        self._parent.check_single(self._idx, pos)
+        self._parent.check_single(self, pos)
 
     @property
     def moving(self):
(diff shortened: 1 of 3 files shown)


ophyd/pseudopos.py
```diff
@@ -327,45 +361,54 @@ def target(self):
...
+        for real, value in zip(self._real, real_pos):
+            logger.debug('[sequential] Moving %s to %s (timeout=%s)',
+                         real.name, value, timeout)
+            if timeout <= 0:
+                raise TimeoutError('Failed to move all positioners within '
+                                   '%s s'.format(timeout))
+
+            t0 = time.time()
+            try:
+                real.move(value, wait=True, timeout=timeout, **kwargs)
```
tacaswell: This means that we can not use sequential mode from bluesky (as we it needs non-blocking).  I suspect we need to build a callback chain here (each one finishing triggers the next one to start and the last one flips the status object.

klauer: Yes, you're right. It's never been able to support non-blocking moves in sequential mode.

@danielballan fixing up #241 would make this much easier.

klauer:
Closing in favor of #271.

PR #170, opened 2015-12-10
REF: Ophyd0 - the refactor to end all refactors

klauer:
For outsiders looking in, a lot of in-person discussion has been going on between @danielballan, @dchabot and myself to get to this point. Before it diverges any further, a merge to master would _probably_ be a good idea.

As @danielballan says, the goal here after such a major refactor should probably just be to "merge the monster" and then open up issues/PRs.

#### What does this _mean_?
* A path to capturing _everything_ during a bluesky scan (or whatever the user desires)
* Nifty new `OphydDevice`s that will ease hardware integration (and hopefully make the DAMA group's lives easier)
* Some of the original ugliness from ophyd has been cleaned up
* Passing continuous-integration tests (!)
* ... massive beamline config refactors incoming

And in more detail...

#### New features
* All hardware devices are now `OphydDevice`s - all signals are are outlined in the class definition
* Add old test suite and fix it up for the refactor
* Coverage via coveralls/codecove, testing via travis/circleci and a Debian docker with EPICS IOCs *

#### Unused features stripped away
* Channel access server (moved to new package [pypvserver](https://github.com/klauer/pypvserver))
* Run engine
* Scans (this is `bluesky` territory now)
* Session manager
* `SignalGroup` is gone, everything is an `OphydDevice`
* Oft-used but undesirable `ophyd.controls.area_detector`

#### Still things that are works-in-progress (likely worthy of creating an issue for each)
* This incomplete PR message
* ~~Still want to re-(re-)do `__repr__` and `__copy__` on all `OphydObject`s such that `copy()` can work again (had this in my testing branch a year ago which never got merged)~~
* `PseudoPositioner` refactor
* Some old 'examples' remain (that are more unit-testy than example-like and should be cleaned now that we finally have CI working)
* More coverage
* And more!

#### On-going discussions (likely worthy of creating an issue for each)
* ~~`device.trigger`~~
* ~~`configure`/`deconfigure` functionality requires some discussion (@dchabot, @danielballan, @tacaswell)~~
* ~~Signal naming, verify it works as desired~~ I think this is finally good
* My hkl branch needs a rebase and refactor to work with the new pseudopositioners. (@dchabot  and @cmazzoli are working on testing this for me, for which I'm extremely grateful).

#### Unimportant notes
* Probably should have squashed/deleted more commits in the rebase to remove intermediate `DevSignal` solution
* `EpicsSignalRO` is only a quick hack on top of `EpicsSignal`
[...]

PR #270, opened 2016-02-17
API: Change default name of user_readback.

danielballan:
Reasons:

* Backward compatibility with routines developed when motors only returned one value
* The vast majority of the time, people want the user_readback, and if they need to check the user_setpoint of course they can still access it.

We talked about doing this in #170 but it seemed like a small detail at the time. Don't worry: this is is not the first wave in a deluge of naming-related PRs into ophyd Device attribute names. EpicsMotor is the only place where I would advocate for overriding the verbose (if nicely explicit) default. Just as users like the special `position` and `value` attributes, they want their `eta` data to show up as "eta," not "eta_user_readback." We might as well make this a facility-wide standard; we are already writing this into most of the beamline configs by hand

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -62,6 +62,10 @@ class EpicsMotor(Device, Positioner):
                          monitor_attrs=monitor_attrs,
                          name=name, parent=parent, **kwargs)
 
+        # Make the default alias for the user_readback the name of the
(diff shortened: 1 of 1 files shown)


klauer:
I'm fine with this change.

klauer:
Merging as tests are passing now.

Issue #230, opened 2016-01-21
Conclusions from preliminary discussion about positioners


danielballan:
- [ ] PVPositioner -> something with a word like "Template" in it
- [ ] PseudoPositioner -> something with a word like "Template" in it

- [x] Refactor `move` in Positioner to express a "synchronous" move in terms of a "async" move, merely watching the status object.

- [x] Add docstrings to PseudoPositioner and more docstring to Positioner.

- [x] Remove "soft positioner" args, attributes, and assumptions from Positioner base class.

- [ ] Add an `actuate` method to Positioner and make it the ultimate source of both `set` and `actuate`. Obviously need to add this Msg to bluesky plans.

- [ ] Ensure the following:
* get and put -- low level, comprehensive
* Positioners set(*args, **kwargs), actuate() -- pick apart setting the set point and moving to the set point
* All devices trigger() and read() -- analogue to set and actuate. Trigger has nothing to do with moving, setting.
* Maybe `move` is refactored as set + actuate (+ wait for status to complete) to maintain current actual behavior

klauer:
Revisit naming of `Template`s - no one was happy about `PVPositionerTemplate` `PVPositionerPCTemplate`. Will leave for next refactor phase.

tacaswell:
Closing this as no longer useful.

PR #271, opened 2016-02-18
WIP: Positioner refactor

klauer:
Supercedes #268 

TODO: discuss in the near future, address comments below

Positioner changes
--------------------------
* Positioner -> PositionerBase + SoftPositioner
* Optional timeout added to status objects, better reporting of failures
* Refactored all positioners in the hierarchy, removing ancient and confusing synchronous code. Blocking moves now use the same asynchronous code by waiting on the returned `MoveStatus`
* PVPositioner (not put-complete version) waits on the setpoint signal (after @tacaswell's PVPositioner PR and a chat with @dchabot, we all agreed that bad blocking behavior here would be more likely due to misconfiguration)
* `Positioner` is still not a `Device`, but nothing stopping this (as far as I know)

PseudoPositioner changes
------------------------------------
* Complete round of tests post-ophyd0 refactor (at least coverage is up to 90% or so)
* Add describe(), read() on pseudosingle (from @tacaswell 's srx config)
* Fix for some namedtuple-related bugs (I consistently get  `namedtuple(a, b, c)` works the same as `tuple((a, b, c))`)
* Fix target position for PseudoSingle
* Fix check_value for PseudoPositioner
* Refactor move, breaking into sequential and concurrent cases
* Added @tacaswell's `MagicPseudoSetPositioner` ([ref](https://github.com/NSLS-II-SRX/ipython_ophyd/blob/master/profile_xf05id1/startup/00-startup.py#L55)) `set(*pos)` implementation, which does not require a PseudoPosition tuple

Other things
-----------------
* Should look over #230 and see what else might be a good idea at this time.

* Implications on the hkl branch?


ophyd/ophydobj.py
```diff
@@ -34,22 +35,45 @@ class StatusBase:
...
         self._lock = RLock()
         self._cb = None
         self.done = False
         self.success = False
+        self.timeout = None
+
+        if timeout is not None:
+            self.timeout = float(timeout)
+
+        if self.timeout is not None:
```
tacaswell: I am not stoked about the threading logic living here.

ophyd/ophydobj.py
```diff
@@ -34,22 +35,45 @@ class StatusBase:
...
     @_locked
-    def _finished(self, *args, **kwargs):
-        # args/kwargs are not really used, but are passed.
-        # uncomment these if you want to go hunting
-        # if args:
-        #     print("this should be empty: {}".format(args))
-        # if kwargs:
-        #     print("this should be empty: {}".format(kwargs))
+    def _finished(self, success=True, **kwargs):
+        # args/kwargs are not really used, but are passed - because pyepics
```
tacaswell: I thought this callback was only handed to ophyd-level callbacks, not pyepics callbacks.

klauer: `pv.put(callback=st._finished)`

ophyd/ophydobj.py
klauer: Then suggest something better. Or perhaps wait (as we discussed in the chat) to talk about the overall implications together rather than the little things that bug you.

ophyd/ophydobj.py
```diff
@@ -151,6 +211,28 @@ def elapsed(self):
...
+        Raises
+        ------
+        TimeoutError
+            If time waited exceeds specified timeout
+        RuntimeError
+            If the status failed to complete successfully
+        '''
+        if timeout is None:
+            timeout = self.timeout
+        super().wait(timeout=timeout, poll_rate=poll_rate)
```
tacaswell: should this stop it's positioner on fail?

tacaswell: and by fail I mean timeout

ophyd/positioner.py
```diff
@@ -85,67 +81,30 @@ def move(self, position, wait=True, moved_cb=None, timeout=30.0):
...
-
-                if check_timeout():
-                    raise TimeoutError('Failed to move %s to %s '
-                                       'in %s s (no motion)' %
-                                       (self.name, position, timeout))
-
-            while self.moving:
-                time.sleep(0.05)
+        if moved_cb is not None:
+            status.finished_cb = partial(moved_cb, obj=self)
```
tacaswell: can you document that `moved_cb` must take and `obj` kwarg?

ophyd/positioner.py
```diff
@@ -85,67 +81,30 @@ def move(self, position, wait=True, moved_cb=None, timeout=30.0):
...
+        self._run_subs(sub_type=self._SUB_REQ_DONE, success=success,
+                       timestamp=timestamp)
         self._reset_sub(self._SUB_REQ_DONE)
 
     def stop(self):
         '''Stops motion'''
-
-        self._run_subs(sub_type=self._SUB_REQ_DONE, success=False)
-        self._reset_sub(self._SUB_REQ_DONE)
+        self._done_moving(success=False)
```
tacaswell: Ah, great, this solves an issue I was having with the undulator motors last night.

ophyd/pseudopos.py
```diff
@@ -105,10 +104,31 @@ def _started_moving(self, value):
...
+
+    def read_configuration(self):
+        return OrderedDict()
 
-class PseudoPositioner(Device, Positioner):
+    def describe_configuration(self):
+        return OrderedDict()
+
+
+class PseudoPositioner(Device, PositionerBase):
```
tacaswell: I am becoming convinced that this is the right place to inject `Device`.

ophyd/pseudopos.py
```diff
@@ -227,22 +246,34 @@ def connected(self):
...
 
     def stop(self):
-        for pos in self._real:
-            pos.stop()
-
-        Positioner.stop(self)
+        try:
+            del self._move_queue[:]
+            for pos in self._real:
+                pos.stop()
```
tacaswell: picky, try needs to be inside or will only make it as far as the first failing positioner

ophyd/pseudopos.py
```diff
@@ -227,22 +246,34 @@ def connected(self):
...
-        pseudo_pos = self.PseudoPosition(pseudo_pos)
+        try:
+            pseudo_pos = self.PseudoPosition(*pseudo_pos)
+        except TypeError as ex:
+            raise ValueError('Not all required values for a PseudoPosition: {}'
+                             '({})'.format(self.PseudoPosition._fields, ex))
+
         for pseudo, pos in zip(self._pseudo, pseudo_pos):
-            pseudo.check_value(pos)
+            low, high = pseudo.limits
```
tacaswell: fine for now, but should think of a sketch about how to support coupled limits.

ophyd/pseudopos.py
```diff
@@ -327,47 +364,94 @@ def target(self):
...
-                    raise TimeoutError('Failed to move all positioners within '
-                                       '%s s'.format(timeout))
+    def _sequential_move(self, real_pos, timeout=None, **kwargs):
+        '''Move all real positioners to a certain position, in series'''
+        # without proper timeout support for status callbacks, we can't safely
+        # chain motions on other callbacks for sequential motion.  so,
+        # 'sequential_move' above always blocks
+        # this is a big TODO
+
+        self._move_queue[:] = zip(self._real, real_pos)
```
tacaswell: huh, learned something about unpacking generators.

ophyd/pseudopos.py
```diff
@@ -327,47 +364,94 @@ def target(self):
...
+        def move_next(obj=None):
+            # last motion complete message came from 'obj'
+            logger.debug('[%s:sequential] move_next called', self.name)
+            with self._finished_lock:
+                if pending_status:
+                    last_status = pending_status[-1]
+                    if not last_status.success:
+                        logger.error('Failing due to last motion')
+                        self._done_moving(success=False)
+                        return
```
tacaswell: I would like this to raise (which will kill the RE)

tacaswell: well, maybe not, have to think about where that exception will go and I suspect it is not where I want it to be.

danielballan:
I read this, focusing on the big picture. I really like this simplification of moving. I will save the big-picture discussion for tomorrow. I'm leaving these piddly comments as a reminder for us during tomorrow's discussion:

- Do we still have good reasons to make PositionerBase not a Device? It leads to lots of duplication -- reimplementing the BlueskyInterface -- and I did not understand these implications when I initially recommended against it.
- Based about about five minutes of checking, I think `_started_moving` is never used, though it is still set at various places. Is it still needed? It's a minor point, but consolidating this logic as much as possible really helps me.
- The `egu`/`_egu` that has nothing to do with hardware can probably be removed entirely, as previously discussed. Likewise, I think `self._timeout` is not referred to and should be remove for clarity.

klauer:
TODO
---------
* PositionerBase should inherit from BlueskyInterface
* PseudoPositioner: smarter move/set/forward/inverse/etc to accept either a sequence or make one
* Timeout stuff

ophyd/pseudopos.py
klauer: Added a note about how we can't really support coupled limits generically - that sort of logic should be handled in subclasses when the default implementation doesn't apply.

ophyd/pseudopos.py
klauer: Not raising here - the returned `status` object should be the one indicating the failure.

danielballan: Does "should be the one indicating failure" mean "should be the one raising"? *Somebody* should raise. Are you suggesting that the RunEngine should check `success` and then raise itself?

klauer: As the client you request a motion with `move` and get a status object back. The actual motion process happens asynchronously in the background. If you do a `status.wait()` the wait logic will raise on motion failure. If you don't use that logic, it's up to you to check whether or not the motion succeeded.

Even if an exception were raised here, it would not funnel back where (I think) you're expecting. Well, to be more specific, it would only come back to the caller if the first move were to fail, not any of the subsequent moves.

ophyd/ophydobj.py
```diff
@@ -75,6 +99,40 @@ def finished_cb(self, cb):
         else:
             self._cb = cb
 
+    def wait(self, timeout=30.0, *, poll_rate=0.05):
```
danielballan: Maybe the default should be `None` here also, lacking a universally appropriate default. If this is mainly for interactive use, the user still has ^C.

dchabot: Just want to point out that, for the case of `EpicsMotor`, once we have added "velocity" and "acceleration" `Components`, we may be able to calculate a reasonable timeout value per move.

danielballan: Cute.

ophyd/pseudopos.py
tacaswell: See https://github.com/NSLS-II/bluesky/pull/350

tacaswell:
This needs a rebase apparently :disappointed:

ophyd/ophydobj.py
tacaswell: I have become convinced that we do want this in all cases.  Should revisit this in future if we get ophyd/epics better integrated with asyncio.

ophyd/ophydobj.py
tacaswell: @klauer This is still an issue.

ophyd/positioner.py
```diff
@@ -62,7 +54,7 @@ def low_limit(self):
     def high_limit(self):
         return self.limits[1]
 
-    def move(self, position, wait=True, moved_cb=None, timeout=30.0):
+    def move(self, position, moved_cb=None, timeout=30.0):
```
tacaswell: This lost the wait input, was that intentional?

tacaswell: reading down I see that it was.  I think I am ok with this as all of the wait behavior is implemented in the sub-classes, but is one more detail about sub-classing these that needs to be documented and gotten right.

ophyd/positioner.py
```diff
@@ -190,7 +148,82 @@ def set(self, new_position, *, wait=False,
...
+            except RuntimeError:
+                raise RuntimeError('Motion did not complete successfully')
+
+        return status
+
     def _repr_info(self):
         yield from super()._repr_info()
         yield ('egu', self._egu)
-        yield ('timeout', self._timeout)
+        yield ('limits', self._limits)
```
tacaswell: why not also include the timeout?

tacaswell: Because I can not read.

ophyd/pseudopos.py
```diff
@@ -10,33 +10,53 @@
...
+    This should not be instantiated on its own, but rather used as a Component
+    in a PseudoPositioner subclass.
+
+    Parameters
+    ----------
+    limits : (low_limit, high_limit)
+        User-defined limits for this pseudo axis.
+    egu : str, optional
+        The engineering units (EGU) for the position
+    prefix : str, optional
```
tacaswell: picky: order to match signature.

ophyd/positioner.py
klauer: Waiting can only happen post-return from `PositionerBase`, so all subclass `move` methods should accept it but `PositionerBase` never sees it

ophyd/ophydobj.py
klauer: Is stopping what's desirable here? (cc @dchabot @danielballan)

ophyd/pseudopos.py
```diff
@@ -10,33 +10,53 @@
...
+from .positioner import PositionerBase
 from .device import Device
 
 
 logger = logging.getLogger(__name__)
 
 
-class PseudoSingle(Positioner):
-    '''A single axis of a PseudoPositioner'''
+class PseudoSingle(PositionerBase):
```
tacaswell: should this sub-class from `SoftPostioner`?

ophyd/pseudopos.py
```diff
@@ -105,20 +149,184 @@ def _started_moving(self, value):
...
+        d = OrderedDict()
+        d[self.name] = {'value': self.position,
+                        'timestamp': time.time()}
+        return d
+
+    def describe(self):
+        desc = OrderedDict()
+        desc[self.name] = {'dtype': 'number',
+                           'shape': [],
+                           'source': 'computed',
```
tacaswell: bikeshed: might want to make this controllable via init.  ex, shove the formula in here?

ophyd/pseudopos.py
klauer: Good catch - I _think_ so. Checking it out now.

ophyd/pseudopos.py
```diff
@@ -105,20 +149,184 @@ def _started_moving(self, value):
...
+        Keyword arguments not related to the position tuple
+
+    Raises
+    ------
+    TypeError
+        On an empty or invalid namedtuple
+    ValueError
+        On a mismatch of parameters
+    '''
+    usage_info = '''Positions can be passed in a number of ways:
```
tacaswell: Is the indentation of all subsequent lines intentional?

ophyd/pseudopos.py
```diff
@@ -105,20 +149,184 @@ def _started_moving(self, value):
...
+
+    # separate position tuple kwargs from other kwargs
+    position_kw = {field: kwargs[field] for field in fields}
+    other_kw = {key: value for key, value in kwargs.items()
+                if key not in fields}
+    position = cls(**position_kw)
+    return position, other_kw
+
+
+class PseudoPositioner(Device, PositionerBase):
```
tacaswell: same here on softpositioner

ophyd/pseudopos.py
klauer: It was in the docstring originally. I'll left-align the rest or use `textwrap.dedent` - any preference?

ophyd/pseudopos.py
```diff
@@ -126,11 +334,14 @@ class PseudoPositioner(Device, Positioner):
     '''
     def __init__(self, prefix, *, concurrent=True, read_attrs=None,
                  configuration_attrs=None, monitor_attrs=None, name=None,
-                 **kwargs):
+                 egu='', **kwargs):
```
tacaswell: how do we want to deal with mixed unit positions, ex '(rad, mm, mm)'  (theta, r, z) in cylindrical cord.  Is that a 'later' problem?

ophyd/pseudopos.py
klauer: I'll do that and also factor in some of #258 (sorry, @dchabot, I can't effectively merge your code at this point...)  at the SoftPositioner level

ophyd/pseudopos.py
danielballan: Maybe egu only makes sense on PseudoSingle.

tacaswell: It makes sense on a `SoftPositioner` as well, everything needs to have a unit at some level.

klauer: `PseudoPositioner` egu, if unspecified, could default to being a composite of its `PseudoSingle` egus?

ophyd/pseudopos.py
```diff
@@ -327,55 +642,169 @@ def target(self):
...
+        real_position : RealPosition
+            The real position output
+        '''
+        # return self.RealPosition()
+        raise NotImplementedError()
 
-    def __call__(self, pseudo_pos):
-        return self.forward(pseudo_pos)
+    @pseudo_position_argument
+    def __call__(self, position):
```
tacaswell: This surprised me a bit to find this.  The forward calculation is not _clearly_ what I would expect `p_pos(foo)` to do.

ophyd/pv_positioner.py
```diff
@@ -77,20 +75,19 @@ class PVPositioner(Device, Positioner):
...
                          monitor_attrs=monitor_attrs,
-                         name=name, parent=parent, timeout=timeout, **kwargs)
+                         name=name, parent=parent, **kwargs)
 
         if self.__class__ is PVPositioner:
             raise TypeError('PVPositioner must be subclassed with the correct '
                             'signals set in the class definition.')
 
-        self.settle_time = float(settle_time)
+        self._egu = egu
```
tacaswell: shouldn't this read from the `readback`?

ophyd/pseudopos.py
klauer: Then I'll remove it.

ophyd/pseudopos.py
danielballan: Yeah, that's what I had in mind. And, is there an example where PseudoSingle and its PseudoPositioner would disagree about the units?

ophyd/pseudopos.py
tacaswell: mild preference to `textwrap.dedent` because I use the left whitespace to see structure in python files.  The other option is to move this block of text to the module level.

ophyd/pseudopos.py
klauer: Maybe where 'mm, mm, mm, mm' might be annoying to see (it isn't a fast-food commercial after all).
I think you'd probably want to be able to specify that the pseudopositioner units are just 'mm' all around

ophyd/pv_positioner.py
klauer: Multiple signals that could specify the units (or not) - setpoint/readback. 

What's the preferred logic here?

ophyd/pseudopos.py
klauer: Can't use the `SoftPositioner` base for `PseudoPositioner` as `SoftPositioner` `move()` adds in things to indicate motion start/stop. `PseudoSingle` works as it never calls `super().move`... Thoughts?

ophyd/pseudopos.py
danielballan: I guess that in that case we can combine them a la `np.squeeze`, but we've gone far enough down this rabbit hole that I'm convinced there's no real problem here. Revisit later.

klauer:
Rebased (because of a one-line PR with epicsmotor default readback name.)

ophyd/pseudopos.py
klauer: Added the logic in to automatically combine PseudoSingle EGUs if PseudoPositioner egu is unspecified.

klauer:
I think I've addressed everything. Requesting one last review, guys.

danielballan:
I took another slow read through this. I'm sure we'll find stuff in production, but I think this is ready to try. :+1:

ophyd/pv_positioner.py
```diff
@@ -105,9 +102,9 @@ def __init__(self, prefix='', *, settle_time=0.05, limits=None, name=None,
             raise ValueError('A setpoint or a readback must be specified')
 
         if self.done is None and not self.put_complete:
-            msg = ('PVPositioner %s is mis-configured. A "done" Signal must be '
-                   'provided or use PVPositionerPC (which uses put completion '
-                   'to determine when motion has completed).'
+            msg = ('PVPositioner %s is mis-configured. A "done" Signal must be'
```
dchabot: `%s` ---> `{}` ?

klauer: Ugh, this is why I'm worried about the `f''` strings in 3.6... I already can't keep them straight!

klauer:
Thanks for the help, @dchabot, @danielballan, @tacaswell! 

This is ready for that button-click (finally).

PR #272, opened 2016-02-18
ENH/API: stage and unstage report objects they touched

danielballan:
I think this will give us the information we need to address https://github.com/NSLS-II/bluesky/issues/343. Without a reproducible bug report it's hard to be sure, but either way this is useful. Now a plan can know what children have been staged to avoid staging a parent and then redundantly trying to stage its child.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -348,7 +348,14 @@ class BlueskyInterface:
         return OrderedDict()
 
     def stage(self):
-        "Prepare the device to be triggered."
+        """
+        Prepare the device to be triggered.
+
+        Returns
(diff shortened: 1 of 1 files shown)


klauer:
Looks good to me.

The only annoying thing will be doing a `stage()` on the command line during debugging and getting a 5 page long `repr` of all the sub-devices, though. Even if you switched this to returning attribute names, you wouldn't be able to represent `self`. Oh well, merging.

danielballan:
Yeah, I had the same thought about the annoying avalanche of device reprs. Oh well.

Issue #274, opened 2016-02-19
Timestamps on EpicsSignal subscriptions report connected time


klauer:
@tacaswell reports it to be even down to the `epics.PV` instance...

klauer:
The issue appears to be a bit different than how it was described. Using a changing analog signal from the pyepics test IOC:

`timestamp_test.py`:
```python
import epics
import time


print('---no auto monitor---')
pv = epics.PV('Py:ai1', auto_monitor=False)
for i in range(5):
    print(i, pv.get(), pv.timestamp)
    time.sleep(0.5)

print('---auto monitor---')
pv = epics.PV('Py:ai1', auto_monitor=True)
for i in range(5):
    print(i, pv.get(), pv.timestamp)
    time.sleep(0.5)
```

The result is:
```
---no auto monitor---
0 26014765.98530756 1456159986.097599
1 26014808.778748155 1456159986.097599
2 26014849.83631311 1456159986.097599
3 26014882.24231184 1456159986.097599
4 26014925.603762623 1456159986.097599
---auto monitor---
0 26014965.722436387 1456159988.707162
1 26015000.455023106 1456159989.160914
2 26015040.582511295 1456159989.685176
3 26015075.78901469 1456159990.145105
4 26015115.85461942 1456159990.668578
```

When auto_monitor is disabled, `pv.timestamp` never updates. Re-creating the PV with auto_monitor on gives back the correct timestamps.

PR #275, opened 2016-02-23
FIX: non-automonitored EpicsSignal timestamps

klauer:
Ref #274

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -273,6 +273,9 @@ class EpicsSignalBase(Signal):
     @raise_if_disconnected
     def timestamp(self):
         '''Timestamp of readback PV, according to EPICS'''
+        if not self._read_pv.auto_monitor:
+            # force updating the timestamp when not using auto monitoring
+            self._read_pv.get_timevars()
         return self._read_pv.timestamp
 
     @property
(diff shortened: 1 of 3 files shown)


tacaswell:
Does this mean we can get into the state where we get the value from one epics 'update' (record processing? sorry, do not know the right terms of art) and then the timestamp from the next in a `read` call?

klauer:
It would be the same if `automonitor` were enabled and a value update callback came inbetween a call to `get()` and `.timestamp` - it's not an atomic operation in either case.

tacaswell:
fair enough.  I take it `pyepics` does not have an atomic operation for this?

klauer:
If automonitored, `epics.PV` value and timestamp come back in the same C `struct` (assuming `form='time'`). However, as it requires two calls on the pyepics level (`get()` and `.timestamp`), therein lies the possibility for a mismatch.

If not automonitored, two `libca` calls are required (at least in the current implementation) to get the value then the timestamp (`get_timevars` at the `epics.PV` level). There's certainly more of a possibility for a mismatch in this case.

We should probably return to the case of automonitoring when timestamps matter - updating the internal state of `EpicsSignal` objects based on the callback data (as the timestamp/value will come bundled together). This would mean that `EpicsSignal.get()` would have to start replicating the logic of the various permutations of `epics.PV.get`'s keyword arguments.

tacaswell:
:+1: 

This needs to get merged before tagging @danielballan

PR #276, opened 2016-02-23
MNT: Example cleanup

klauer:
Referring to the awful results of https://landscape.io/github/NSLS-II/ophyd/5/messages/smell, it became apparent that the examples needed a trimming. The remaining examples are not exactly good examples or terribly useful - however, with this PR they should all at least run.

There were a few configuration-related mistakes I found in the main repo, though, so this may be worth retagging for. (Sorry, @danielballan)


klauer:
Rebased and further cleaned up - landscape reports 96% now.

ophyd/areadetector/plugins.py
```diff
@@ -419,7 +411,14 @@ class OverlayPlugin(PluginBase):
                              ('y', 'MaxSizeY_RBV'))),
                    doc='The maximum size in XY')
 
-    overlay_1 = C(Overlay, 'Overlay1:')
```
danielballan: Does this call for a DynamicDeviceCompoent? I don't super care, though. Everything else looks good. Self-merge at will.

examples/config.py
```diff
@@ -75,4 +75,5 @@ def setup_loggers(logger_names, fmt=LOG_FORMAT):
                   ]
 
 # this is a real instrument - it is not currently used, but be cautious!
-scalers = ['XF:23ID2-ES{Sclr:1}']
+# scalers = ['XF:23ID2-ES{Sclr:1}']
+scalers = ['XF:03IDC-ES{Sclr:2}']
```
tacaswell: is this still a real insturment?

ophyd/areadetector/plugins.py
klauer: Probably would have been a good idea.

examples/config.py
klauer: Yes, we don't have a software-only scaler.

PR #280, opened 2016-02-27
FIX: If a sub-device is staged, the entire device is staged.

danielballan:
Consider a device `d` with a subdevice `d.s`.

Before #272, `RE(Count([d, d.s]))` errored because the RunEngine tried to redundantly stage `d.s`. Now that `d.stage()` returns `[d, d.s]`, the RunEngine knows to skip staging `d.s`. But `RE(Count([d.s, d]))` (note the reverse order) still errors. The RunEngine has no way of telling `d` not to stage its child `d.s`. (And that is probably a good thing -- it's not the RunEngine's business.)

Possible solutions:
1. Stop being strict that a device can't be staged redundantly. Downside: the risk of failing to clean up and, say, writing into an old file by mistake.
2. Accept arguments to `stage()` that would affect which child devices get staged.
3. Insist that staging a child causes its parent to staged. Thus, `d.s.stage()` would stage `d` as well and return `[d, d.s]`. Thus the RunEngine knows to skip staging `d`.

This PR is Solution 3. That is to say, you can't stage *part* of a device. That makes sense to me: the device was grouped for a reason. And one could imagine people getting into trouble trying to stage part of an area detector (the file-writer but not the trigger, for example).

Implementation note: The ~~`stage_rachet`~~ [renamed] `_defer_to_parent` flag is way of avoiding this infinite loop: child defers to parent, which tries to stage child, which defers to parent, which tries to stage child.... Credit to @tacaswell for thinking of this pattern.

cc @ambarb @cmazzoli Once this gets merged, you can put detectors into `gs.DETS` in any order.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -372,7 +372,6 @@ class BlueskyInterface:
         if self.parent is not None and self._defer_stage_to_parent:
             # Stage parent, which will then stage self.
             return self.parent.stage()
-        self._defer_stage_to_parent = True  # reset for next time
         logger.debug("Staging %s", self.name)
         self._staged = Staged.partially
 
@@ -401,9 +400,8 @@ class BlueskyInterface:
                     try:
                         device._defer_stage_to_parent = False
                         device.stage()
-                    except Exception:
+                    finally:
                         device._defer_stage_to_parent = True
-                        raise
                     devices_staged.append(device)
         except Exception:
             logger.debug("An exception was raised while staging %s or "
@@ -430,7 +428,6 @@ class BlueskyInterface:
         if self.parent is not None and self._defer_unstage_to_parent:
             # Stage parent, which will then stage self.
             return self.parent.unstage()
-        self._defer_unstage_to_parent = True  # reset for next time
         logger.debug("Unstaging %s", self.name)
         self._staged = Staged.partially
         devices_unstaged = []
@@ -442,9 +439,8 @@ class BlueskyInterface:
                 try:
                     device._defer_unstage_to_parent = False
                     device.unstage()
-                except Exception:
+                finally:
                     device._defer_unstage_to_parent = True
-                    raise
                 devices_unstaged.append(device)
 
         # Restore original values.
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -362,11 +362,13 @@ class BlueskyInterface:
         if self._staged == Staged.no:
             pass  # to short-circuit checking individual cases
         elif self._staged == Staged.yes:
-            raise RedundantStaging("Device is already staged. Unstage it first.")
+            raise RedundantStaging("Device {!r} is already staged. "
+                                   "Unstage it first.".format(self))
         elif self._staged == Staged.partially:
-            raise RedundantStaging("Device has been partially staged. Maybe the "
-                               "most recent unstaging encountered an error "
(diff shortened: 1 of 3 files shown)


ophyd/device.py
```diff
@@ -360,11 +360,19 @@ def stage(self):
...
+                                "to cleanup from last use before staging "
+                                "again.")
         elif self._staged == Staged.partially:
             raise RuntimeError("Device has been partially staged. Maybe the "
                                "most recent unstaging encountered an error "
                                "before finishing. Try unstaging again.")
+            if self.parent is not None:
+                # Defer to self.parent, which then stages self and siblings.
+                try:
+                    return self.parent.stage()
```
tacaswell: Won't this cause an infinite loop still?

danielballan:
@tacaswell This gets tricky because child calls parent calls sibling calls parent (again) which is at this point in a "partially staged" stage and raises. Latest force-pushed commit might finally have this right.

tacaswell:
I think the parent needs to set `defer_to_parent` to `False` in each of it's children before it calls them.

danielballan:
When you're right, you're right.

On Sat, Feb 27, 2016 at 5:24 PM -0800, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


I think the parent needs to set defer_to_parent to False in each of it's children before it calls them.

klauer:
The PR message makes me think I understand this, but then I look at the code and end up confused. @danielballan can we have a quick chat about this tomorrow?

danielballan:
@klauer Do the new tests help? I also corrected the implementation. I'm happy to discuss in person tomorrow as well.

ophyd/device.py
```diff
@@ -422,7 +437,12 @@ def unstage(self):
         for attr in self._sub_devices[::-1]:
             device = getattr(self, attr)
             if hasattr(device, 'unstage'):
-                device.unstage()
+                try:
+                    device._defer_unstage_to_parent = False
+                    device.unstage()
+                except Exception:
```
tacaswell: I think this should be a `finally`

danielballan:
More tests! In an ideal world, these tests would have less repeated code. I know how to do this in nose, but ten minutes of reading didn't make it clear how to do this in py.test. I'm hoping we can punt and just be grateful for copious tests. :- D

tacaswell:
@ericdill any advice on how to do generative tests?

ericdill:
@tacaswell I assume that by "generative tests" you are talking about the `yield func, arg1, arg2, ...` syntax in nose. Here is a basic example that uses the `@pytest.mark.parameterize` fixture.  @danielballan If I can see your implementation in nose I can show you how I would translate it to pytest

```python
################# Yield style with nose #################

def test_func(num, char):
    # do something interesting with num and char
    pass

def yield_test():
    list1 = [1, 2, 3]
    list2 = list('abc')

    for num, char in zip(list1, list2):
        yield test_func, num, char


################# Fixture style with pytest #################

@pytest.mark.parametrize('num char', zip([1, 2, 3], ['a', 'b', 'c']))
def test_func(num, char):
    # do something interesting with num and char
    pass
```

klauer:
(In case anyone else is thinking to click 'merge' here, please wait - we discussed an alternative offline)

PR #281, opened 2016-03-01
FIX:Add the RuntimeError handling.

hhslepicka:


hhslepicka:
@klauer

klauer:
This was a bit of an oversight which @hhslepicka ran into (myself separately as well).
The timeout status check all happens in a background thread, so the end result is that this just dumps an annoying exception to the console. 

The reason why this PR is not also calling `_finished` when `RuntimeError` is raised is because at this point, the status object has already been marked as completed.

So:

1. done=True, success=True -> return silently
2. done=True, success=False -> RuntimeError
3. operation didn't complete in the given time -> TimeoutError

In short :+1:

tacaswell:
For my curiosity, where does the `RunTimeerror` come from?

klauer:
@tacaswell https://github.com/hhslepicka/ophyd/blob/7ed004a716abb9196a958b4488de651b55fded97/ophyd/status.py#L212

tacaswell:
Fair enough.

<bikeshe> maybe we want a custom exception here?  I am a bit worried about shadowing a `RuntimeError` from something else </bikeshed>

tacaswell:
sigh, html needs to add a bikeshed tag so what I typed into the previous message would have worked.

klauer:
```python
exception RuntimeError:
Raised when an error is detected that doesn’t fall in any of the other categories. 
```

`OperationFailedError`, `StatusCompletedButNotSuccessfullyError`? I don't really care either way... If you do, PR and bikeshed away

ericdill:
@tacaswell, HTML might have a bikeshed tag, but you spelled the opening tag "bikeshe" :smile:

PR #282, opened 2016-03-02
Expose new 'root' attribute, pointing to ultimate ancestor

danielballan:
This, plus a forthcoming bluesky PR, solves the problem that #280 meant to solve, better. Thanks, @klauer.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -5,7 +5,7 @@ from collections import (OrderedDict, namedtuple)
 
 from .ophydobj import OphydObject
 from .status import DeviceStatus
-from .utils import (ExceptionBundle, set_and_wait)
+from .utils import (ExceptionBundle, set_and_wait, RedundantStaging)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


ophyd/device.py
```diff
@@ -348,6 +348,15 @@ def read(self):
     def describe(self):
         return OrderedDict()
 
+    @property
+    def root(self):
```
klauer: `Device` is an `OphydObject` - do we need this on both?

ophyd/device.py
```diff
@@ -805,3 +816,7 @@ def _repr_info(self):
         yield ('read_attrs', self.read_attrs)
         yield ('configuration_attrs', self.configuration_attrs)
         yield ('monitor_attrs', self.monitor_attrs)
+
+
+class RedundantStaging(Exception):
```
klauer: Can you please move this to `ophyd.utils.errors` and subclass from `OpException`?

danielballan:
See https://github.com/NSLS-II/bluesky/pull/369

ophyd/device.py
danielballan: Good call. Done.

klauer:
:+1:

Issue #279, opened 2016-02-25
settle_time for motors class for motor.subdevice1 seems ineffective


ambarb:
Tried to increase settle_time for a motor due to noise in scan from 0.5 to 1.0 but scan time did not roughly double (31.6 s for 21 pts to 32.0 s for 21 pts).  Unable to confirm if change was effective because ````Nanopositioner.bz.describe```` is not an option as it is for ````theta.describe````.

````
In [104]: dscan(nanop.bz,-.1,.1,20)
+------------+--------------+-------------------------+--------------------------+--------------------------+
|   seq_num  |        time  |  dif_beam_stats2_total  |  nanop_bz_user_readback  |  nanop_bz_user_setpoint  |
+------------+--------------+-------------------------+--------------------------+--------------------------+
Transient Scan ID: 59052
Persistent Unique Scan ID: 'c7d647b8-52d1-4db1-8683-6695cc9fb47b'
|         1  |  23:49:10.1  |               1.77e+07  |                   -2.28  |                   -2.28  |
|         2  |  23:49:11.7  |               1.77e+07  |                   -2.27  |                   -2.27  |
|         3  |  23:49:13.3  |               1.77e+07  |                   -2.26  |                   -2.26  |
|         4  |  23:49:14.9  |               1.77e+07  |                   -2.25  |                   -2.25  |
|         5  |  23:49:16.5  |               1.77e+07  |                   -2.24  |                   -2.24  |
|         6  |  23:49:18.0  |               1.78e+07  |                   -2.23  |                   -2.23  |
|         7  |  23:49:19.6  |               1.78e+07  |                   -2.22  |                   -2.22  |
|         8  |  23:49:21.3  |               1.78e+07  |                   -2.21  |                   -2.21  |
|         9  |  23:49:22.9  |               1.78e+07  |                   -2.20  |                   -2.20  |
|        10  |  23:49:24.4  |               1.78e+07  |                   -2.19  |                   -2.19  |
|        11  |  23:49:26.0  |               1.78e+07  |                   -2.18  |                   -2.18  |
|        12  |  23:49:27.6  |               1.78e+07  |                   -2.17  |                   -2.17  |
|        13  |  23:49:29.3  |               1.78e+07  |                   -2.16  |                   -2.16  |
|        14  |  23:49:30.9  |               1.78e+07  |                   -2.15  |                   -2.15  |
|        15  |  23:49:32.5  |               1.78e+07  |                   -2.14  |                   -2.14  |
[...]

PR #283, opened 2016-03-02
ENH: Settle time for all positioners, and more

klauer:
Proposed fix for #279

`settle_time` is added at the lowest level possible, making it possible to use throughout the hierarchy.
I think this is the cleanest it can possibly be done. Commence the bikeshedding.

If we can agree this isn't a terrible idea, I'll add it to `Device` as a kwarg as well.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -27,6 +27,9 @@ class StatusBase:
     timeout : float, optional
         The default timeout to use for a blocking wait, and the amount of time
         to wait to mark the operation as failed
+    settle_time : float, optional
+        The amount of time to wait between the caller specifying that the
+        status has completed to running callbacks
     """
     def __init__(self, *, timeout=None, settle_time=None):
         super().__init__()
@@ -61,6 +64,10 @@ class StatusBase:
         finally:
             self._timeout_thread = None
 
+    def _settled(self):
+        '''Hook for when status has completed and settled'''
+        pass
+
     def _settle_then_run_callbacks(self, success=True):
         # wait until the settling time is done to mark completion
         if self.settle_time > 0.0:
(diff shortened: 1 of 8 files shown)


klauer:
cc @ambarb; we're working on it.

danielballan:
What about putting the settle time in MoveStatus but not StatusBase? Now I recuse myself from the discussion because there's a nasty rebase in the balance, and I am an interested party. ;- )

klauer:
@danielballan that's definitely an option. A reason for shoving it in the base is to allow `Device`s to have a settling time as well. `Device.set()` would return a `DeviceStatus` or some subclass of `StatusBase`/`DeviceStatus`, but likely not a `MoveStatus`.

ophyd/status.py
```diff
@@ -56,17 +64,36 @@ def _timeout_thread(self):
...
         # gives in a bunch of kwargs that we don't care about
-
-        self.success = success
-        self.done = True
-
-        if self._cb is not None:
-            self._cb()
-            self._cb = None
+        if success and self.settle_time > 0:
+            # delay gratification until the settle time is up
```
dchabot: @klauer you are a riot

tacaswell:
I think this is reasonable.  It makes sense that devices (not scans) know how long they need to settle.  Long term we need to work out a way to communicate to the run engine that it needs to add a sleep in after this move (which may be by plans asking motors if they have a settle time?) but :+1: for now.

klauer:
I believe this is ready, broken tests were fixed. Merge at will.

tacaswell:
@dchabot The idea would be to move the sleep out of a thread on to the asycio event loop so it would be an either-or setting.  Same with my desire to move the timeout logic on to a `call_later`  task on the event loop.

PR #284, opened 2016-03-03
MNT: do not include precision if not an int

tacaswell:


klauer:
`precision` is not required in the schema. If you're ok with the change, merge it @dchabot

tacaswell:
I am biased towards _not_ adding it if it does not make sense vs including bogus values.  Checking 'do I have this' is simpler than checking 'is this valid'.  

This also makes it easier to find IOCs that should be returning precision but are not.

PR #285, opened 2016-03-03
FIX: correct inheritance of squashing tiff

tacaswell:


klauer:
No tests cover this, but @tacaswell vouches for its correctness

PR #286, opened 2016-03-03
ENH: Positioner/EpicsMotor Additions Partially Fix Issue: #273

### LINKED ISSUE #100: Scan continues when hard limit switch is hit
ascan does not stop or give an error when scanned motor hits a _hard_ limit switch.

### LINKED ISSUE #273: Positioner/EpicsMotor Additions
@danielballan, @dchabot, @klauer 
Per discussed on the meeting on Feb, 17th regarding the EpicsMotor we decided that Positioner/EpicsMotor should implement the following:

- [ ] homing
- [ ] calibration (exists in commands.py?)
- [ ] limit switch stat (+ exception raising + status object stuff)
- [ ] velocity
- [ ] acceleration
- anything else?

hhslepicka:
- [x] Homing
- [x] Calibration - set_current_position
- [x] Limit Switch Status
- [x] Velocity
- [x] Acceleration
- [x] Offset and Offset Freeze
- [x] Check the limit switches on the _move_changed prior to invoke _done_moving.
- [x] Tests

Closes #273 
Closes #100

### DIFF
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -236,6 +236,16 @@ def mov(positioner, position):
                 flag += 1
 
     print(tc.Normal + '\n')
+    for err in [s for s in stat if s.success == False]:
+        device = err.pos
+        reason = "Unknown"
+        if isinstance(device, EpicsMotor):           
+            if device.at_high_limit_switch:
+                reason = "Motor reached the high limit switch."
+            elif device.at_low_limit_switch:
+                reason = "Motor reached the low limit switch."
+
+        print('Warning: {} failed to reach the target position. Reason: {}'.format(device.name, reason))
 
 
 @ensure(PositionerBase, None)
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -236,7 +236,7 @@ def mov(positioner, position):
                 flag += 1
 
     print(tc.Normal + '\n')
-    for err in [s for s in stat if s.success == False]:
+    for err in [s for s in stat if not s.success]:
         device = err.pos
         reason = "Unknown"
         if isinstance(device, EpicsMotor):           
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -176,9 +176,9 @@ class EpicsMotor(Device, PositionerBase):
         status = super().move(position, **kwargs)
 
         if direction == HomeEnum.forward:
-            self.motor_home_forward.put(1, wait=False)
+            self.home_forward.put(1, wait=False)
         else:
-            self.motor_home_reverse.put(1, wait=False)
+            self.home_reverse.put(1, wait=False)
 
         try:
             if wait:
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -156,9 +156,9 @@ class EpicsMotor(Device, PositionerBase):
            Position to set.
 
         '''
-        self.motor_set_use_switch.put(1, wait=True)
+        self.set_use_switch.put(1, wait=True)
         self.user_setpoint.put(pos, wait=True)
-        self.motor_set_use_switch.put(0, wait=True)
+        self.set_use_switch.put(0, wait=True)
 
     @raise_if_disconnected
     def home(self, direction, wait=True, **kwargs):
@@ -217,7 +217,7 @@ class EpicsMotor(Device, PositionerBase):
         if was_moving and not self._moving:
             success = True
             # Check if we are moving towards the low limit switch
-            if self.motor_direction_of_travel.get() == 0:
+            if self.direction_of_travel.get() == 0:
                 if self.low_limit_switch.get() == 1:
                     success = False
             # No, we are going to the high limit switch
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -55,15 +55,15 @@ class EpicsMotor(Device, PositionerBase):
     motor_is_moving = Cpt(EpicsSignalRO, '.MOVN')
     motor_done_move = Cpt(EpicsSignalRO, '.DMOV')
     motor_stop = Cpt(EpicsSignal, '.STOP')
-    motor_offset_freeze_switch = Cpt(EpicsSignal, '.FOFF')
-    motor_velocity = Cpt(EpicsSignal, '.VELO')
-    motor_acceleration = Cpt(EpicsSignal, '.ACCL')
-    motor_set_use_switch = Cpt(EpicsSignal, '.SET')
-    motor_at_high_limit_switch = Cpt(EpicsSignal, '.HLS')
-    motor_at_low_limit_switch = Cpt(EpicsSignal, '.LLS')
-    motor_home_forward = Cpt(EpicsSignal, '.HOMF')
-    motor_home_reverse = Cpt(EpicsSignal, '.HOMR')
-    motor_direction_of_travel = Cpt(EpicsSignal, '.TDIR')
+    offset_freeze_switch = Cpt(EpicsSignal, '.FOFF')
+    velocity = Cpt(EpicsSignal, '.VELO')
+    acceleration = Cpt(EpicsSignal, '.ACCL')
+    set_use_switch = Cpt(EpicsSignal, '.SET')
+    high_limit_switch = Cpt(EpicsSignal, '.HLS')
+    low_limit_switch = Cpt(EpicsSignal, '.LLS')
+    home_forward = Cpt(EpicsSignal, '.HOMF')
+    home_reverse = Cpt(EpicsSignal, '.HOMR')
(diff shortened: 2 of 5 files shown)


klauer:
Great work, @hhslepicka! 

You'll notice that this PR also updates circle to use py.test, with overhauled epicsmotor tests (which are much more granular).

ophyd/epics_motor.py
```diff
@@ -42,12 +48,23 @@ class EpicsMotor(Device, PositionerBase):
...
         The amount of time to wait after moves to report status completion
     '''
+    user_offset = Cpt(EpicsSignal, '.OFF')
     user_readback = Cpt(EpicsSignalRO, '.RBV')
     user_setpoint = Cpt(EpicsSignal, '.VAL', limits=True)
     motor_egu = Cpt(EpicsSignal, '.EGU')
     motor_is_moving = Cpt(EpicsSignalRO, '.MOVN')
     motor_done_move = Cpt(EpicsSignalRO, '.DMOV')
     motor_stop = Cpt(EpicsSignal, '.STOP')
+    motor_offset_freeze_switch = Cpt(EpicsSignal, '.FOFF')
```
dchabot: Suggest dropping the redundantly redundant 'motor_' naming.

ophyd/epics_motor.py
```diff
@@ -42,12 +48,23 @@ class EpicsMotor(Device, PositionerBase):
...
+    motor_velocity = Cpt(EpicsSignal, '.VELO')
+    motor_acceleration = Cpt(EpicsSignal, '.ACCL')
+    motor_set_use_switch = Cpt(EpicsSignal, '.SET')
+    motor_at_high_limit_switch = Cpt(EpicsSignal, '.HLS')
```
dchabot: How about just 'high_limit_switch', 'low_limit_switch'?

ophyd/epics_motor.py
```diff
@@ -155,7 +259,17 @@ def _move_changed(self, timestamp=None, value=None, sub_type=None,
...
-            self._done_moving(timestamp=timestamp, value=value)
+            success = True
+            # Check if we are moving towards the low limit switch
+            if self.motor_direction_of_travel.get() == 0:
+                if self.at_low_limit_switch:
+                    success = False
+            # No, we are going to the high limit switch
+            else:
+                if self.at_high_limit_switch:
+                    success = False
```
dchabot: Is checking the direction of travel necessary?
```python
if self.at_low_limit_switch or self.at_high_limit_switch:
    success = False

self._done_moving(...)
```

ophyd/epics_motor.py
```diff
@@ -129,6 +146,93 @@ def position(self):
...
+            self.motor_home_reverse.put(1, wait=False)
+
+        try:
+            if wait:
+                status_wait(status)
+        except KeyboardInterrupt:
+            self.stop()
+            raise
+
+        return status
```
dchabot: I have a few questions about `def home`.
Not urgent...

ophyd/commands.py
```diff
@@ -236,6 +236,16 @@ def mov(positioner, position):
                 flag += 1
 
     print(tc.Normal + '\n')
+    for err in [s for s in stat if s.success == False]:
```
danielballan: minor point: `if not s.success` is preferred so that anything "false-y" would work correctly here.

ophyd/epics_motor.py
danielballan: Good idea. I suggest we make any changes to existing names in separate PR. I think we have an "alias component" concept that could provide back-compat if we decide we need it. If not, we should add it.

ophyd/epics_motor.py
danielballan: Is this a boolean?

ophyd/epics_motor.py
```diff
@@ -129,6 +146,93 @@ def position(self):
...
+
+        Returns
+        -------
+        frozen : bool
+        '''
+        return bool(self.motor_offset_freeze_switch.get())
+
+    @property
+    @raise_if_disconnected
+    def at_high_limit_switch(self):
```
danielballan: Properties like this are convenient because they save the user from typing `.get()`, but they crowd the namespace. And they might confuse users who try to do `some_motor.offset = 5` and have to be told to do `some_motor.user_offset.put(5)` instead. I think I am against adding more of them, but I could be persuaded that I am wrong. Thoughts?

danielballan:
Yes, I'm very happy to see this contribution! Thank you for the good docstrings and clear communication through the development process. I left a few minor comments, which you should interpret as opinions that you might politely consider but freely ignore. :- )

ophyd/epics_motor.py
hhslepicka: Yes, 'high_limit_switch' and 'low_limit_switch' are boolean.

ophyd/epics_motor.py
hhslepicka: Yes, it is. Imagine the following situation: 
If the user try to send a movement in the opposite direction of the limit switch which is active but the distance wasn't enough to deactivate it we need to consider this a valid movement. That's why the direction of travel is being considered! ;)

ophyd/epics_motor.py
hhslepicka: :+1: to remove the redundant 'motor_' . Just worried about the back-compat as @danielballan pointed.

hhslepicka:
Monday seems perfect to me!

On Thu, Mar 3, 2016 at 5:38 PM, dchabot <notifications@github.com> wrote:

> Looks good, @hhslepicka <https://github.com/hhslepicka> . Thanks!
> Maybe we can chat about this Monday? I'm out tomorrow...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NSLS-II/ophyd/pull/286#issuecomment-191999539>.
>



-- 
Hugo Slepicka
*hhslepicka@gmail.com <hhslepicka@gmail.com>*
*+1 631 855 5717*

hhslepicka:
Thanks, guys! It was a great opportunity to learn a bunch of new stuff!
Hats off to @klauer for all the help through the process and for all the patience to teach me! :+1:

ophyd/epics_motor.py
klauer: `motor_egu` is a necessity, otherwise it clashes with the `egu` property (`'stop'` is similar, and `'is_moving'`/`'done_move'` sounds too much like existing methods). `namedtuple` makes it such that we can't name `motor_egu` `_egu` instead (no underscores in its keys). 

I'm very much in favor of removing the leading `motor_` where it won't cause confusion.

ophyd/epics_motor.py
klauer: I'm not crazy about the properties, either. It'll likely lead to confusion if/when people add `'at_high_limit_switch'` to `read_attrs` and wonder why it doesn't work.

klauer:
I think things look pretty good, otherwise. Let's wait for the chat on Monday with @dchabot before merging.

ophyd/epics_motor.py
tacaswell: I am also against using properties where we can use a soft-signal instead for exactly the reason @klauer said.  Properties can not be exposed to `read`.

hhslepicka:
@dchabot @klauer @danielballan .... what time would be best for you to discuss this today?

danielballan:
Late afternoon is best for me.

On Mon, Mar 7, 2016 at 6:57 AM -0800, "Hugo Slepicka" <notifications@github.com<mailto:notifications@github.com>> wrote:


@dchabot<https://github.com/dchabot> @klauer<https://github.com/klauer> @danielballan<https://github.com/danielballan> .... what time would be best for you to discuss this today?

ophyd/epics_motor.py
```diff
@@ -18,11 +18,17 @@
 from .positioner import PositionerBase
 from .device import (Device, Component as Cpt)
 from .status import wait as status_wait
+from enum import Enum
 
 
 logger = logging.getLogger(__name__)
 
 
+class HomeEnum(str, Enum):
```
tacaswell: why mix `Enum` with `str`?

klauer: Like `enum.IntEnum`, it allows you to compare the enum value with the builtin-typed value. For example, `HomeEnum.forward == 'forward'` == `True`. Without this, the comparison will fail.

hhslepicka: To support who wants to use `HomeEnum.forward` and just  `"forward"` (the same for reverse).
I'm accepting suggestions here! =)

tacaswell: nifty

hhslepicka:
@klauer, @dchabot , @danielballan .... here are the fixes.
Is there anything else that needs to be changed? :pray:

klauer:
Looks good to me, if the tests pass

danielballan:
Thanks again, @hhslepicka!

PR #287, opened 2016-03-04
ENH/TST: Define OphydObj.__contains__

danielballan:
This is just to generate useful discussion around https://github.com/NSLS-II/wishlist/issues/108. I wanted something to concrete to look at.

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -54,6 +54,19 @@ class OphydObject:
                 return root
             root = root.parent
 
+    def __contains__(self, other):
+        "Check whether other is a sub-device or sub-signal."
+        # Walk other's ancestry, looking for self.
+        ancestor = other.parent
+        if ancestor is None:
+            return False
+        while True:
+            if ancestor is self:
+                return True
+            if ancestor is other.root:
+                return False
+            ancestor = ancestor.parent
+
     def _run_sub(self, cb, *args, **kwargs):
         '''Run a single subscription callback
 
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -67,6 +67,32 @@ class OphydObject:
                 return False
             ancestor = ancestor.parent
 
+    def common_ancestor(self, other):
+        """Find first common ancestor Device.
+
+        Parameters
+        ----------
+        other : OphydObj
+        
+        Returns
+        -------
+        ancestor : OphydObj
+        """
+        other_ancestry = []
+        ancestor = other
+        while True:
+            other_ancestry.append(ancestor)
+            ancestor = ancestor.parent
+            if ancestor is None:
+                break
+        ancestor = self
+        while True:
+            if ancestor is None:
+                return None
+            if ancestor in other_ancestry:
+                return ancestor
+            ancestor = ancestor.parent
+
     def _run_sub(self, cb, *args, **kwargs):
         '''Run a single subscription callback
 
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -37,6 +37,12 @@ class Signal(OphydObject):
 
         self._timestamp = timestamp
 
+    def __contains__(self, other):
+        """
+        Signals are the leaves of the Device tree. This is always False.
+        """
+        return False
+
     def trigger(self):
         '''Call that is used by bluesky prior to read()'''
         # NOTE: this is a no-op that exists here for bluesky purposes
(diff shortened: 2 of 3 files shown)


danielballan:
Then bluesky plans would do something like:

```python
# objs_read is a list of objs we have read so far in this Event bundle
# obj is an object we are about to read

for obj_read in objs_read:
    if obj in obj_read or obj_read in obj:
         # redundnat -- cut out
# carry on
```

tacaswell:
Signals also need an always false `__contains__`

danielballan:
Haha, did we skip the discussion and jump right to performance optimizations? Are you :+1: on this API?

tacaswell:
It isn't a performance optimization, signal can go in the detector list so bluesky will be checking `obj in signal` which will fail.

I do not see how this as 'policy', this is the hardware abstraction knowing enough about it's self to allow the code that implements 'policy' to do so.  Knowing if two python objects are part of the same `Device` seems pretty fundamental to the abstraction.

`common_ancestor` still seems independently useful but not immediately needed.

danielballan:
> It isn't a performance optimization, signal can go in the detector list so bluesky will be checking obj in signal which will fail.

Why would it fail? I believe it's even covered by a test in the original commit. Note that this is on OphydObj (superclass of both Device and Signal) not Device itself.

danielballan:
> common_ancestor still seems independently useful but not immediately needed.

Is this different that `obj.root == other.root`? Were you thinking that `common_ancestor` returned the ancestor itself, not a boolean?

tacaswell:
@danielballan Ah, I missed where it went (thought this was on `Device`) nvm :sheep: 

Yes,  `common_ancestor` should return the first common ancestor

```python

a.b.c.common_ancestor(a.b.d) is a.b

```

danielballan:
I agree that we will very likely want a method like that as well. I have added it, with tests, in a separate commit. OK with this, @dchabot, @klauer?

klauer:
Sorry for the late input...

I'm not a fan of adding on `__contains__` to `Device`. It seems to indicate that we should also have `__getitem__`/etc defined. It also seems to me that it should be more of `'attribute_name' in device` versus `obj in device`-style checks...

For whatever reason, it seems less ugly to me to define `Device.__iter__` such that you can walk through its immediate children -- sub-devices and signals. From there, a simple (external) utility function could generate a whole tree without poking into any `Device` internals.

Come to think of it, all of these functions (`root` included) could/should have been implemented outside of the device itself...

danielballan:
> I'm not a fan of adding on __contains__ to Device. It seems to indicate that we should also have __getitem__/etc defined. 

I disagree that `__contains__` and `__getitem__` (etc.) go hand in hand. I think supporting `in` gives users a terse and expressive way of testing the hierarchy. I do agree your feeling that we should avoid an exploding litany of convenience methods and generally leave graph exploration to external code.

>  It also seems to me that it should be more of 'attribute_name' in device versus obj in device-style checks...

We *sort of* already have this: `other.name in self.signal_names` would accomplish this with strings. I guess the point of adding `in` is to avoid exposing `signal_names` as another specially-named piece of API. Python provides `__contains__`, and `signal_names` is sort of reinventing that.

klauer:
> I do agree your feeling that we should avoid an exploding litany of convenience methods and generally leave graph exploration to external code.

If that's the case, then we still have to define what's the minimal amount of things to leave on `Device` so that tree traversal/relationship-determination can be done with external utility functions. From a high-level standpoint, `parent` and `children` are the only things that are strictly necessary. I suppose it's too late to remove `root`, so we can leave it in. The only question is how to expose `children`, then.

> We sort of already have this: other.name in self.signal_names would accomplish this with strings. I guess the point of adding in is to avoid exposing signal_names as another specially-named piece of API. Python provides __contains__, and signal_names is sort of reinventing that.

In hindsight, leaving `signal_names` public that was probably a mistake (beside for its name). We probably should've used `__contains__`, `__iter__`, etc. Oh well.

danielballan:
Actually, `signal_names`is not public in the sense of being used by downstream (bluesky). It could be deprecated/made private/left undocumented. Do we agree that `__iter__` would iterate over the subdevices/signals in order? That  would give us `__contains__` for free. The rest of the traversal could be handled by external functions if you prefer.

klauer:
I think so (on all accounts), but perhaps we should wait until Monday to act on this.

danielballan:
Agreed. I'm underwater with beamline support today. To be continued.

danielballan:
It still seems like `__iter__` and/or `__contains__` would be useful to add, but there is no immediate need in light of https://github.com/NSLS-II/bluesky/pull/382

klauer:
We definitely could use some more options to introspect `Device`s. One of my main desires to allow such hierarchy traversal at the class-level (i.e., `@classmethod`s) without poking into private attributes (or requiring signal instantiation). Those methods could then be used at the instance level as well. More on that some other time.

PR #291, opened 2016-03-08
FIX: make base Signal.get permissive

tacaswell:
@klauer was right

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -64,7 +64,7 @@ class Signal(OphydObject):
         yield ('value', self.value)
         yield ('timestamp', self.timestamp)
 
-    def get(self):
+    def get(self, **kwargs):
         '''The readback value'''
         return self._readback
 


ophyd/signal.py
```diff
@@ -64,7 +64,7 @@ def _repr_info(self):
         yield ('value', self.value)
         yield ('timestamp', self.timestamp)
 
-    def get(self):
+    def get(self, **kwargs):
```
ericdill: why does this need kwargs?

klauer:
Well, you sure know how to get me to click the merge button, don't you @tacaswell?

ophyd/signal.py
tacaswell: because sub-classes do and we in various places in PVPositioner we pass key words in.  When mocking things out with soft signals you need to be able to accept the extra (unused) inputs.

ericdill: Ah, thanks for the clarification.  I might suggest that future you would probably want a note in here why **kwargs is in this function

PR #293, opened 2016-03-10
ENH: make report property cooperative

### LINKED ISSUE #292: Remove 'report' property, or add to PseudoSingle, SoftPositioner, etc
See [#89](https://github.com/NSLS-II/Bug-Reports/issues/89).

Using `SoftPositioner` as a null-axis in a pseudopositioner, a call to `log_pos` will `raise`, as there is no `report` property. See also CSX1 startup file, "99-tardis.py". The _only_ place `report` is used is in `commands.logbook_add_objects`.

tacaswell:
closes #292

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -763,11 +763,6 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
         '''
         return cls._device_tuple
 
-    @property
-    def report(self):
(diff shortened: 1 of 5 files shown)


klauer:
I'd have preferred to remove `report` altogether, but this works too.

tacaswell:
This was less work :imp:

PR #296, opened 2016-03-10
DOC: documentation updates

klauer:
(@ericdill or @danielballan) any thoughts on getting the restructured text in the dynamic device component dynamic default docstring displaying as intended? It comes out as squashed, unformatted text.

I think overall this is an improvement...


klauer:
Fixed default component docstrings to render well with sphinx, with help from @ericdill

ericdill:
Lgtm. I'll add auto building ophyd docs to my to-do list

PR #297, opened 2016-03-10
ENH: Default timeout

klauer:
Re-added per-Positioner default timeouts, at the request of @tacaswell


tacaswell:
attn @cmazzoli

cmazzoli:
Thank you for the attn... I am not sure understanding.
Before there was a common 30.0 sec waiting time, now we have it split in several values depending on the kind of motor/motion?
A stupid question: in general, why is timeout needed??

PR #302, opened 2016-03-14
Mnt timeout stop

tacaswell:


danielballan:
:+1: for finishing inside a finally block

PR #303, opened 2016-03-14
ENH/WIP: partially working Pixirad configuration

tacaswell:
attn @gjwillms 

This needs to be re-tested at SRX as we lost the working version due to upgrade thrashing.

PR #304, opened 2016-03-16
ENH: flesh out remaining areadetector.cams

klauer:
Supersedes #303; Adds: linux firewire, lightfield, pixirad, point grey cameras
Note: I do not have the detectors to check this.

Generated with `ophyd.areadetector.utils.create_detector_stub()` and tweaked by hand.

@gjwillms some of your components were in `CamBase`, and I tried to tweak some names to keep backward compatibility with yours. There may be some others I didn't catch. If you have name suggestions (complaints) now's probably the time to fix this up...

### DIFF
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -241,9 +241,13 @@ class BrukerDetectorCam(CamBase):
 
 
 class FirewireLinDetectorCam(CamBase):
-    _html_docs = ['FirewireWinDoc.html']
-    # TODO
-    pass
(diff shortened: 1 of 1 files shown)


danielballan:
@gjwillms Any comments on this, particularly naming inconsistencies noted by @klauer above?

gjwillms:
The changes described above will not negatively impact our operations and I agree that we should maintain consistency as much as we possibly can.  Thanks for cleaning up.

danielballan:
Thanks, @gjwillms.

PR #305, opened 2016-03-18
WIP/ENH: add FlyerDevice for discussion

klauer:
This is a test implementation of what I had in mind for `monitor_attrs`. Basically, for every `monitor_attr` you configure, you get a separate data stream with its own timestamps.

... that said, I'm not suggesting that this is a good solution - it's more for discussion than anything.

**Problems**
* Can't monitor sub-devices
* `describe` gets over-ridden by a new one that gives a list of descriptors, one for each attribute (required by bluesky for flyers)
* `collect` output may not be 100% in the form bluesky is looking for (as of the first commit at least)
* `stop()` (or `collect()`) needs to be called asynchronously to tell it to stop monitoring. 

A way around the last point may be to have `kickoff()` return a finished status object. This would require some other detector to be blocking/waiting to complete during a run.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1,4 +1,5 @@
 import time as ttime
+import functools
 import logging
 import textwrap
 from enum import Enum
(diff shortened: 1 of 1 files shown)


klauer:
We're going to nuke `monitor_attrs`; closing

danielballan:
Let's not lose track of this code, though. Something like this should be in ophyd or bluesky to make it easy to build flyers out of normal devices.

PR #306, opened 2016-03-18
REF: remove Device.monitor_attrs

klauer:
As discussed. Is it safe to say that no beamline configs are using it?


tacaswell:
:+1:

PR #307, opened 2016-03-23
ENH: Add LastUidPublish, which publishes only when run is complete.

danielballan:


klauer:
Note to self: the EPICS waveform PV should have been `FTVL`=`char` with `NELM` set high enough, though the current implementation does work.

ophyd/callbacks.py
```diff
@@ -74,3 +74,43 @@ def __call__(self, name, doc):
...
+        Fail if the UID signal is disconnected
+    put_kw : kwargs, optional
+        Keyword arguments to send to uid_signal.put()
+    '''
+    def __call__(self, name, doc):
+        '''Bluesky callback with document info'''
+        if name == 'start':
+            self.last_start = doc
+
+        if self.last_start is not None and name == 'stop':
```
tacaswell: maybe just publish the uid of the start attached to the stop?

danielballan: Oh. Duh. Yes, that would be better.

klauer: Going to do that or should we just merge as-is?

tacaswell: Not worth it, it would simplify this code, but then make the inheritance funny (by leaving this class with a vestigial attribute.

PR #309, opened 2016-04-13
WIP: add support for quadEM EPICS devices

dchabot:
Do not merge yet.

This provides support for quad electrometers (AH401/501, NSLS EM, Tegra, etc).
More tests and docs to follow.

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -30,11 +30,15 @@ class TriggerBase(BlueskyInterface):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # settings
-        self.stage_sigs.update([(self.cam.acquire, 0),  # If acquiring, stop.
-                                (self.cam.image_mode, 1),  # 'Multiple' mode
-                                ])
+        # careful here: quadEM devices have areadetector components but,
+        # they have no 'cam' plugin. See QuadEM initializer.
+        if hasattr(self, 'cam'):
+            self.stage_sigs.update([(self.cam.acquire, 0),  # If acquiring, stop
+                                    (self.cam.image_mode, 1),  # 'Multiple' mode
+                                    ])
+            self._acquisition_signal = self.cam.acquire
+
         self._status = None
-        self._acquisition_signal = self.cam.acquire
 
 
 class SingleTrigger(TriggerBase):
--- a/ophyd/__init__.py
+++ b/ophyd/__init__.py
@@ -21,6 +21,7 @@ from .device import (Device, Component, FormattedComponent,
                      DynamicDeviceComponent)
 from .status import StatusBase
 from .mca import EpicsMCA, EpicsDXP
+from .quadem import QuadEM
 
 # Areadetector-related
 from .areadetector import *
--- a/ophyd/quadem.py
+++ b/ophyd/quadem.py
@@ -1,0 +1,91 @@
+from ophyd import (EpicsSignalRO, DetectorBase, SingleTrigger,
+                   Component as Cpt, EpicsSignal,
+                   ADComponent as ADCpt
+                  )
+from ophyd.areadetector import (EpicsSignalWithRBV, ImagePlugin, StatsPlugin,
+                                SingleTrigger
+                               )
+
+
+class QuadEM(SingleTrigger, DetectorBase):
+    model = Cpt(EpicsSignalRO, 'Model')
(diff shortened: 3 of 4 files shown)


arkilic:
@hhslepicka Is this QuadEM class any use to LIX?

hhslepicka:
@arkilic It is exactly what I need. @dchabot told me about it! ;)

arkilic:
👍

ophyd/areadetector/trigger_mixins.py
```diff
@@ -30,11 +30,15 @@ class TriggerBase(BlueskyInterface):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # settings
-        self.stage_sigs.update([(self.cam.acquire, 0),  # If acquiring, stop.
-                                (self.cam.image_mode, 1),  # 'Multiple' mode
-                                ])
+        # careful here: quadEM devices have areadetector components but,
+        # they have no 'cam' plugin. See QuadEM initializer.
```
tacaswell: Does it have something that _acts_ like `cam`?

ophyd/areadetector/trigger_mixins.py
```diff
@@ -30,11 +30,15 @@ class TriggerBase(BlueskyInterface):
...
+        if hasattr(self, 'cam'):
+            self.stage_sigs.update([(self.cam.acquire, 0),  # If acquiring, stop
+                                    (self.cam.image_mode, 1),  # 'Multiple' mode
+                                    ])
+            self._acquisition_signal = self.cam.acquire
```
tacaswell: is a version of this provided in the base class?  Bit worried about this producing `AttributeErrors`

tacaswell:
@gjwillms Does SRX have one of these too?

gjwillms:
Yes.


On Apr 14, 2016, at 13:38, Thomas A Caswell <notifications@github.com<mailto:notifications@github.com>> wrote:


@gjwillms<https://github.com/gjwillms> Does SRX have one of these too?

ophyd/areadetector/trigger_mixins.py
dchabot: Yes. The QuadEM class itself has `acquire` and `acquire_mode` attributes, exactly analogous to `cam.acquire` and `cam.image_mode` in `CamBase`.

Naturally, the EPICS records are named differently between quadEM and areaDetector...

ophyd/areadetector/trigger_mixins.py
dchabot: There are certainly cleaner ways to do this.

As it was, `TriggerBase` was essentially hard-coded for areadetector _cameras_, whereas it should be generalized to permit flexibility of _start_acquiring_ and _acquisition_mode_ `Signals`. 
Maybe as kwargs in the initializer?

tacaswell: In some of the other classes the init takes in a string of the name of the component that is the 'cam like'

klauer: The only functionality the base class gives above `BlueskyInterface` is the two `stage_sigs`. These options offer little benefit compared to just updating `stage_sigs` manually:

```python
# attribute names by kwarg
TriggerBase(..., start_acquiring='cam.acquire', start_acquiring_value=0, 
            acquisition_mode='cam.image_mode', acquisition_mode_value=1)

# case where it has no acquisition mode signal for example
TriggerBase(..., start_acquiring=None, acquisition_mode=None)

# specifying just cam requires that 'image_mode' and 'acquire' are the same
TriggerBase(..., cam_attr='cam')
TriggerBase(..., cam_attr=None)
```

Do we need another layer on top of TriggerBase that expects a `cam_attr` (i.e., `ADTriggerBase(prefix, *, cam_attr='cam'`) such that `TriggerBase` can be reused for `cam`-less quadEM and others? Or should `TriggerBase` be exclusively areaDetector-with-cam?

cc @danielballan

tacaswell: On consideration I like adding kwargs over my suggestion.

danielballan: I think adding a `cam_attr` kwarg to `TriggerBase` itself is fine (no need for an extra layer IMO). I would prefer that it have the default value `'cam'` for back-compat. Since `TriggerBase` is scoped in the areadetector module I think it's OK that it makes some assumptions about how area detector attributes are named by default. But making it configurable is a good idea.

klauer:
Good enough for now. Merging to let @hhslepicka and others get access to a dev build.

PR #310, opened 2016-04-19
MNT: scraped positioners sorted by name

klauer:
Positioners as sub-devices get ordered randomly without this in `wh_pos`

### DIFF
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -7,6 +7,7 @@ import functools
 import sys
 import warnings
 from contextlib import contextmanager, closing
+from operator import attrgetter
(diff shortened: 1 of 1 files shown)


ophyd/commands.py
```diff
@@ -7,6 +7,7 @@
 import sys
 import warnings
 from contextlib import contextmanager, closing
+from operator import attrgetter
```
danielballan: Neat. I did not know that this was a thing.

klauer: Yeah, it's a bit cleaner than a lambda. See also `operator.itemgetter`

PR #312, opened 2016-05-03
API: Signal.set = Signal.put for bluesky compatibility

klauer:
Hopefully saving @danielballan a bit of typing

Allows `Signal`s to be used in plans (specifically `count_time` in the bs refactor)


klauer:
Closing in favor of actually tested #313

PR #313, opened 2016-05-03
ENH/TST: Implement Signal.set()

danielballan:
per conversation with @klauer, `Signal.set` is a synonym for `Signal.put`, so that things like `det.count_time.set(2)` work as expected.

### DIFF
--- a/tests/test_signal.py
+++ b/tests/test_signal.py
@@ -512,6 +512,12 @@ class EpicsSignalTests(unittest.TestCase):
         self.assertEquals(desc['dtype'], 'array')
         self.assertEquals(desc['shape'], [1,])
 
+    def test_set_method(self):
+        sig = Signal()
+
+        sig.set(28)
+        self.assertEquals(sig.get(), 28)
+
 
 class DerivedSignalTests(unittest.TestCase):
     def test_soft_derived(self):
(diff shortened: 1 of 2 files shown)


tests/test_signal.py
```diff
@@ -512,6 +512,15 @@ def test_describe(self):
         self.assertEquals(desc['dtype'], 'array')
         self.assertEquals(desc['shape'], [1,])
 
+    def test_set(self):
+        epics.PV = FakeEpicsPV
+        sig = EpicsSignal('my_pv')
+        sig._write_pv.enum_strs = ('enum1', 'enum2')
```
tacaswell: what is this line doing?

klauer: It's unnecessary here

danielballan:
Hmm. The test I added is broken. Do I misunderstand how EpicsSignal works?

klauer:
The `FakeEpicsPV` (a mock `epics.PV`) thing was implemented really poorly (by yours truly). 
I thought the problem you ran into was that write_pv and read_pv were actually 2 separate fake PVs, but that doesn't appear to be the case. Hmm...

tacaswell:
Just use a `Signal`?

danielballan:
Force-pushed a test using `Signal`... we'll see.

klauer:
Yes, an obvious replacement that I should've suggested without going down that rabbit-hole....

danielballan:
(Travis failure is unrelated to PR. New test passes.)

PR #315, opened 2016-05-05
ENH: Make filestore spec configurable through attribute.

danielballan:
Makes it easy to support other specs like 'AD_HDF5_SWMR'.

Now, to handle SWMR specifically it might be better to set up ophyd's HDF5FilePlugin to ask AreaDetector's HDF5 plugin whether it supports SWMR, but that's a bigger project.

This simple knobs seems like a useful addition in any case.

attn @stuwilkins


stuwilkins:
Thanks @danielballan this is now needed as we have SWMR working, are we ready to go? Can someone merge. Thanks!

PR #316, opened 2016-05-05
ENH: fully fleshed out Signal.set and EpicsSignal.set

klauer:
This will give us the option to use `set` similar to `Positioner`s for all signals, with asynchronous status completion objects.

* `Signal.set` uses `set_and_wait` (runs in a separate CAThread) - and so all signals derived from this will do the same
* `EpicsSignal.set` uses put completion if specified by the user (requires no background thread), and falls back to `set_and_wait`
* Switched Signal from using `StatusBase` -> `Status`

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -8,7 +8,9 @@ from .utils import (ReadOnlyError, LimitError)
 from .utils.epics_pvs import (pv_form, waveform_to_string,
                               raise_if_disconnected, data_type, data_shape)
 from .ophydobj import OphydObject
-from .status import DeviceStatus
+from .status import Status
+from .utils import set_and_wait
+
 
 logger = logging.getLogger(__name__)
 
@@ -36,12 +38,13 @@ class Signal(OphydObject):
             timestamp = time.time()
 
         self._timestamp = timestamp
+        self._set_thread = None
 
     def trigger(self):
         '''Call that is used by bluesky prior to read()'''
         # NOTE: this is a no-op that exists here for bluesky purposes
         #       it may need to be moved in the future
-        d = DeviceStatus(self)
+        d = Status(self)
         d._finished()
         return d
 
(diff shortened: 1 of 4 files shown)


ophyd/signal.py
```diff
@@ -97,7 +100,48 @@ def put(self, value, *, timestamp=None, force=False, **kwargs):
...
+        st : Status
+            This status object will be finished upon return in the
+            case of basic soft Signals
+        '''
+        def set_thread():
+            nonlocal timeout
+
+            if timeout is None:
+                timeout = 10
+                # TODO set_and_wait does not support a timeout of None
```
danielballan: Since `set_and_wait` polls and holds control in the meantime, I'm comfortable with this.

danielballan:
Looks good to me. Thanks for tackling this -- a dragon at the bottom of a rabbit hole as Tom says. I'll wait for @tacaswell to sign off and merge.

ophyd/signal.py
```diff
@@ -681,6 +728,45 @@ def put(self, value, force=False, **kwargs):
...
+        '''
+        if not self._put_complete:
+            return super().set(value, timeout=timeout, settle_time=settle_time)
+
+        # using put completion:
+        # timeout and settle time is handled by the status object.
+        st = Status(self, timeout=timeout, settle_time=settle_time)
+
+        def put_callback(**kwargs):
+            st._finished(success=True)
```
klauer: `settle_time` is ignored in this case - would need a background thread for it.

tacaswell:
I would rather push the thread based `set` down to `EpicsSignalBase` and keep a base implementation in `Signal` that just return that it is done, but not going to push that (this is just my standard 'I want a sub-set of ophyd with out epics' line).

klauer:
I think that having the `set_and_wait` implementation at the base of `Signal` makes sense, personally. 

You could make any sort of signal derived from that class, and `set_and_wait` would be called on it for `set()`. If you leave it as a no-operation (a status object that returns immediately), a delay between put() and get() returning it would not be captured.

PR #317, opened 2016-05-06
API/TST: flyers have methods: complete, describe_collect, pause, resume

klauer:
* `describe_collect()` is separate from the normal `Device.describe()` in that it has a description of what `collect()` will return. The return value is also a list of descriptions, whereas `describe` and `describe_configuration` return only a single dictionary each.
* `complete()` returns a status object indicating when data is ready for collection (all implemented flyers return completed status objects currently)
* `pause` pauses flyer data acquisition
* `resume` resumes after a pause

### DIFF
--- a/ophyd/flyers.py
+++ b/ophyd/flyers.py
@@ -89,10 +89,12 @@ class AreaDetectorTimeseriesCollector(Device):
     def pause(self):
         # Stop without clearing buffers
         self.control.put('Stop', wait=True)
+        super().pause()
 
     def resume(self):
         # Resume without erasing
         self.control.put('Start', wait=True)
+        super().resume()
 
     def complete(self):
         if self.control.get(as_string=True) == 'Stop':
--- a/ophyd/flyers.py
+++ b/ophyd/flyers.py
@@ -163,6 +163,8 @@ class WaveformCollector(Device):
                       'timestamps': {self.name: v},
                       'time': v}
                 yield ev
+        else:
+            yield from []
 
     def _repr_info(self):
         yield from super()._repr_info()
--- a/ophyd/flyers.py
+++ b/ophyd/flyers.py
@@ -6,56 +6,10 @@ from collections import OrderedDict
 from .signal import (Signal, EpicsSignal, EpicsSignalRO)
 from .status import DeviceStatus
 from .device import (Device, Component as C, BlueskyInterface)
+from .utils import OrderedDefaultDict
 
 
 logger = logging.getLogger(__name__)
-
-
-class OrderedDefaultDict(OrderedDict):
-    """
-    a combination of defaultdict and OrderedDict
-
-    source: http://stackoverflow.com/a/6190500/1221924
-    """
-    def __init__(self, default_factory=None, *a, **kw):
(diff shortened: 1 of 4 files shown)


klauer:
Resurrected FlyerDevice here in the form of a mixin plus a basic test - I think now's probably the right time to get it sorted.

ophyd/flyers.py
```diff
@@ -106,9 +137,99 @@ def collect(self):
...
                 yield ev
 
-    def stop(self):
-        self.select.put(0, wait=True)  # Stop Collection
-
     def _repr_info(self):
         yield from super()._repr_info()
         yield ('data_is_time', self.data_is_time.get())
+
+    def describe_collect(self):
```
danielballan: I find `describe_flyer` less confusing, but I don't really care.

klauer:
Moved `stop` -> `complete` (+ status object) and added some more docs.
Ready for review, I think.

ophyd/flyers.py
```diff
@@ -1,6 +1,15 @@
+import time as ttime
+import functools
+import logging
+from collections import OrderedDict
+
 from .signal import (Signal, EpicsSignal, EpicsSignalRO)
 from .status import DeviceStatus
-from .device import (Device, Component as C)
+from .device import (Device, Component as C, BlueskyInterface)
```
ericdill: I wonder why github syntax highlighted `C`?

klauer: From the GitHub docs:

> Name your variables in a stupid manner and our new syntax highlighting engine will ensure all of your friends notice and make fun of you!

ophyd/flyers.py
tacaswell: I like `describe_collect` better.

ophyd/flyers.py
```diff
@@ -105,10 +166,151 @@ def collect(self):
...
+        collected['values'].append(value)
+        collected['timestamps'].append(timestamp)
+
+    def describe_collect(self):
+        '''Description of monitored attributes retrieved by collect'''
+        # single stream?
+        desc = OrderedDict()
+        for attr in self.monitor_attrs:
+            desc.update(self._describe_attr_list([attr]))
+        return {self.stream_name: desc}
```
tacaswell: shouldn't each monitor attr get it's own stream?  They will each need their own descriptors.

ophyd/flyers.py
```diff
@@ -105,10 +166,151 @@ def collect(self):
...
+
+    def _clear_monitors(self):
+        '''Clear all subscriptions'''
+        for attr in self._collected_data.keys():
+            obj = getattr(self, attr)
+            try:
+                obj.clear_sub(self._monitor_callback)
```
tacaswell: I do not think this will actually remove the callback.  Above you add a partial of the callback, this tries to remove the method (unless partial is smarter than I think).

klauer: Good call, fixed

klauer:
Rebased and refactored to add multiple stream names for the monitor flyer mixin. Look better @tacaswell?

PR #318, opened 2016-05-07
Additional work on flyers

danielballan:
Stacking commits on top of @klauer's #317 in case we want to merge and tag an RC before we see him on here again.

Several small commits here -- see msgs for details. (I found a way to use OrderedDefaultDict, and I'm excited about it.)

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -492,6 +492,12 @@ class BlueskyInterface:
         self._staged = Staged.no
         return devices_unstaged
 
+    def pause(self):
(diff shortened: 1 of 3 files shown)


tacaswell:
syntax error: https://travis-ci.org/NSLS-II/ophyd/builds/128508084#L934

danielballan:
fix force-pushed

klauer:
Updated my PR with these changes and a bit more to get it passing. Baton passed back.

PR #323, opened 2016-05-16
ENH/TST: add makedirs with mode fix, make_dir_tree

### LINKED ISSUE #84: Put Stuart's directory-generating script into ophyd.utils

klauer:
Closes #84

### DIFF
--- a/ophyd/utils/paths.py
+++ b/ophyd/utils/paths.py
@@ -21,13 +21,14 @@ def makedirs(path, *, mode=0o777, mode_base=None):
         return []
 
     head, tail = os.path.split(path)
-    ret = makedirs(head, mode=mode, mode_base=mode_base)
-    try:
-        os.mkdir(path)
-    except OSError as ex:
-        if 'File exists' not in str(ex):
-            raise
 
+    # Recurse on all directories above:
+    ret = makedirs(head, mode=mode, mode_base=mode_base)
+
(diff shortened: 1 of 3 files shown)


ophyd/utils/paths.py
```diff
@@ -0,0 +1,71 @@
...
+    '''
+    # Permissions not working with os.makedirs -
+    # See: http://stackoverflow.com/questions/5231901
+    if not path or os.path.exists(path):
+        return []
+
+    head, tail = os.path.split(path)
+    ret = makedirs(head, mode=mode, mode_base=mode_base)
+    try:
+        os.mkdir(path)
```
tacaswell: you can use the `exists_ok` kwarg here to avoid this try/except.

klauer: `exists_ok` is in `os.makedirs`, which IIRC didn't set permissions correctly, hence this function. Though not sure if this was only necessary in py2k...

tacaswell: that is what the `os.chmod` a few lines down fixes.  I think you can let python do the recursion checking for you and then juts fix up the permissions?

klauer: Fair enough, it would remove this ugly exception handling at least. Will fix.

klauer:
Rebased to get those test fixes from other PRs and modified to use `os.makedirs`.

PR #324, opened 2016-05-19
ENH: add tolerance, relative tolerance to signals (and set_and_wait)

klauer:
WIP - I (think) we only need this on signals

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -157,16 +157,18 @@ class Signal(OphydObject):
                 if settle_time is not None:
                     time.sleep(settle_time)
             finally:
-                self._set_thread = None
                 st._finished(success=success)
+                self._set_thread = None
 
         if self._set_thread is not None:
             raise RuntimeError('Another set() call is still in progress')
 
         st = Status(self)
+        self._status = st
         self._set_thread = epics.ca.CAThread(target=set_thread)
+        self._set_thread.daemon = True
         self._set_thread.start()
-        return st
+        return self._status
 
     @property
     def value(self):
@@ -644,6 +646,8 @@ class EpicsSignal(EpicsSignalBase):
     @raise_if_disconnected
     def tolerance(self):
         '''The tolerance of the write PV, as reported by EPICS
+
+        Can be overidden by the user at the EpicsSignal level.
 
         Returns
         -------
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -19,6 +19,7 @@ logger = logging.getLogger(__name__)
 class Signal(OphydObject):
     '''A signal, which can have a read-write or read-only value.
 
+
     Parameters
     ----------
     value : any, optional
@@ -29,12 +30,13 @@ class Signal(OphydObject):
     tolerance : any, optional
         The absolute tolerance associated with the value
     rtolerance : any, optional
-        The relative tolerance associated with the value
+        The relative tolerance associated with the value, used in
+        set_and_wait as follows:
+            absolute(setpoint - readback) <= (tolerance + rtolerance *
+                                              absolute(readback))
 
     Attributes
     ----------
-    tolerance : any, optional
-        The absolute tolerance associated with the value
     rtolerance : any, optional
         The relative tolerance associated with the value
     '''
@@ -52,7 +54,8 @@ class Signal(OphydObject):
 
         self._timestamp = timestamp
         self._set_thread = None
-        self.tolerance = tolerance
+        self._tolerance = tolerance
+        # self.tolerance is a property
         self.rtolerance = rtolerance
 
     def trigger(self):
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -75,6 +75,15 @@ class Signal(OphydObject):
         '''Timestamp of the readback value'''
         return self._timestamp
 
+    @property
+    def tolerance(self):
+        '''The absolute tolerance associated with the value.'''
+        return self._tolerance
+
+    @tolerance.setter
+    def tolerance(self, tolerance):
+        self._tolerance = tolerance
+
     def _repr_info(self):
         yield from super()._repr_info()
         value = self.value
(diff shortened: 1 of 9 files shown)


klauer:
Commits f639bdf and 6545461 are just test cleanup and further pytest conversion that needed to be done.

hhslepicka:
:+1:

tacaswell:
@hhslepicka get off the internet!  Even I took a break around my wedding.

ericdill:
> Even I took a break around my wedding.

Only because your commit rights were forcibly taken away from you :laughing:

ophyd/signal.py
```diff
@@ -26,11 +26,23 @@ class Signal(OphydObject):
     timestamp : float, optional
         The timestamp associated with the initial value. Defaults to the
         current local time.
+    tolerance : any, optional
```
tacaswell: when would these not be scalars?

ophyd/signal.py
```diff
@@ -26,11 +26,23 @@ class Signal(OphydObject):
...
+        The absolute tolerance associated with the value
+    rtolerance : any, optional
+        The relative tolerance associated with the value
```
tacaswell: is this a fraction?

ophyd/signal.py
```diff
@@ -612,6 +644,33 @@ def wait_for_connection(self, timeout=1.0):
...
+        Can be overidden by the user at the EpicsSignal level.
+
+        Returns
+        -------
+        tolerance : float or None
+        Using the write PV's precision:
+            If precision == 0, tolerance will be None
+            If precision > 0, calculated to be 10**(-precision)
+        '''
+        if self._tolerance is not None:
```
tacaswell: you through a `self._tolerance = None` in the `__init__`?

ophyd/signal.py
klauer: ```For finite values, isclose uses the following equation to test whether two floating point values are equivalent:```
```python
absolute(a - b) <= (atol + rtol * absolute(b))
```

ophyd/signal.py
klauer: There's nothing stopping you from using `set_and_wait` on an array PV, and with an appropriate tolerance `np.isclose` would handle it just fine, I think.

ophyd/signal.py
klauer: `tolerance.setter` merely sets `self._tolerance`, but it is probably better practice/less confusing to set the internal value

tacaswell: I like to be able to see all of the internal state an object has in the `__init__`, I find it greatly helps later debugging.  I have also been burned by a couple mpl bugs where we set internal state in setters and then if you have _other_ setters that rely on the previous getter you can get very confusing 'attribute not defined' bugs.

ophyd/status.py
```diff
@@ -93,6 +93,9 @@ def _settle_then_run_callbacks(self, success=True):
                 self._cb = None
 
     def _finished(self, success=True, **kwargs):
+        if self.done:
+            return
```
tacaswell: I think this may produce a code path where the callback does not get run.  As `MoveStatus` can take a `done` kwarg (which tracing this looks like it gets reset back to false in the `super.__init__`, which makes be think there is another bug here.

ophyd/signal.py
tacaswell: fair enough.

ophyd/utils/epics_pvs.py
```diff
@@ -344,26 +344,36 @@ def wrapper(self, *args, **kwargs):
...
+        allowed absolute tolerance between the readback and setpoint values
+    atol : float, optional
+        allowed relative tolerance between the readback and setpoint values
 
     Raises
     ------
     TimeoutError if timeout is exceeded
     """
+    if atol is None and hasattr(signal, 'tolerance'):
+        atol = signal.tolerance
```
tacaswell: This can be done as

```python
if atol is None:
    atol = getattr(signal, 'tolerance', None)
```

ophyd/status.py
klauer: I think the `done` init kwarg was only to allow object duplication via `repr`. It doesn't appear that it's used otherwise. I think that `_finished()` definitely should only be called once.

The reset in `super.init` is a bug due to re-arranging `__init__` a bit - will fix.

tacaswell: None of the other `Status` family of objects take this?

ophyd/utils/epics_pvs.py
```diff
@@ -384,15 +406,26 @@ def set_and_wait(signal, val, poll_time=0.01, timeout=10):
...
-    if not isinstance(a, str):
-        a = enums[a]
-    if not isinstance(b, str):
-        b = enums[b]
+
+    # if either relative/absolute tolerance is used, use numpy
+    # to compare:
+    if atol is not None or rtol is not None:
+        return np.allclose(a, b,
+                           rtol=rtol if rtol is not None else 1e-7,
```
tacaswell: Is changing the defaults from numpy intentional?

Might be worth making a dict with keys conditional on `None`-ness

klauer: Guess I got confused. `numpy` mixes them up too on occasion:
```python
def allclose(a, b, rtol=1.e-5, atol=1.e-8):
def assert_allclose(actual, desired, rtol=1e-7, atol=0, err_msg='', verbose=True):
```

Worth it for 2 lines? Or you want these defaults passed in?

ophyd/status.py
klauer: All should pass `done`/`success` through now, added some more docstrings as well

klauer:
Bump @tacaswell

ophyd/signal.py
```diff
@@ -129,20 +163,24 @@ def set_thread():
...
             else:
+                logger.debug('set_and_wait(%r, %s) succeeded => %s', self.name,
+                             value, self.value)
                 success = True
                 if settle_time is not None:
                     time.sleep(settle_time)
             finally:
-                self._set_thread = None
                 st._finished(success=success)
+                self._set_thread = None
```
tacaswell: why this?

klauer: Looks like I swapped lines unintentionally - is it worth bothering to swap them back?

tacaswell: No, just checking that there was not a subtle reason I was not seeing.

PR #326, opened 2016-05-22
ENH: Make automatic plugin-enabling on stage() optional.

danielballan:
Recall https://github.com/NSLS-II/ophyd/issues/245

I believe this is a non-breaking change --- it just adds a knob -- but I am not 100% confident that the change to `detectors.py` is correct because I am not sure what `self.enable.get()` returns.

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -69,11 +69,11 @@ class PluginBase(ADBase):
                                           self._plugin_type,
                                           self.plugin_type.get(), self.prefix))
 
+        self.enable_on_stage()
+        self.ensure_blocking()
         if self.parent is not None and hasattr(self.parent, 'cam'):
             self.stage_sigs.update([('parent.cam.array_callbacks', 1),
                                     ])
-        self.enable_on_stage()
-        self.ensure_blocking()
 
     _default_configuration_attrs = (ADBase._default_configuration_attrs +
                                     ('port_name', 'nd_array_port', 'enable',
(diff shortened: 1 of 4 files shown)


tacaswell:
:+1:

klauer:
From earlier discussion: we should probably have a common Python soft-signal that allows us to enable/disable plugins and stop their `stage_sigs` from even being set. This may lead us into having more shared signals - which I'm hoping we can decide on at some point (like in https://github.com/klauer/hxntools/blob/ophyd0/hxntools/detectors/trigger_mixins.py#L28 )

ophyd/areadetector/plugins.py
```diff
@@ -86,8 +89,9 @@ def __init__(self, *args, **kwargs):
     port_name = C(EpicsSignalRO, 'PortName_RBV', string=True)
 
     def stage(self):
-        # Ensure the plugin is enabled. We do not disable it on unstage.
-        set_and_wait(self.enable, 'Enable')
+        if self.auto_enable:
+            # Ensure the plugin is enabled. We do not disable it on unstage.
+            set_and_wait(self.enable, 'Enable')
```
danielballan: If this was disabled to begin with, it should be disabled on unstage, a la 'stage_sigs'.

hhslepicka:
I am working exactly on this piece of code here at SLAC...
What I did was just a check ```if self.enable not in self.stage_sigs:``` to
keep compatibility and if it is not then I set_and_wait self.enable.
It worked! ;)
One other thing that I noticed is that the data on the buffer is processed
even if you disable the plugin... which means... it finishes the work that
it was doing.

danielballan:
Big thanks to @hhslepicka for a useful offline conversation in this.

PR #329, opened 2016-05-25
ENH: allow monitorflyermixin to be pivoted on insert

klauer:
Also fixes dtype='array' for non-pivoted option

### DIFF
--- a/ophyd/flyers.py
+++ b/ophyd/flyers.py
@@ -196,8 +196,13 @@ class MonitorFlyerMixin(BlueskyInterface):
         A mapping of attribute -> stream name
         If an attribute is not in this dictionary, the stream name will default
         to the object's name.
+    pivot : bool, optional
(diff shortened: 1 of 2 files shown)


ophyd/flyers.py
```diff
@@ -259,11 +265,26 @@ def _get_stream_name(self, attr):
         obj = getattr(self, attr)
         return self.stream_names.get(attr, obj.name)
 
+    def _describe_with_dtype(self, attr, *, dtype='array'):
```
tacaswell: fun thing, you can make mandatory keyword arguments by not providing a default.

klauer: Interesting - I didn't realize that was an option. I'm guessing by your comment you're suggesting it here?

tacaswell: not enough to not merge it, but it is a nice way to enforce readable function calls in cases where there is not a good default.

PR #330, opened 2016-05-27
Remove lingering references to prettytable

ericdill:


tacaswell:
We still use prettytable in the code.

ericdill:
So I guess we do need to continue packaging a prettytable conda recipe then

Issue #331, opened 2016-05-31
Make PVPositioner work without a done signal


klauer:
Rely on tolerance setting, comparing setpoint to readback value to indicate `done` status.

klauer:
#332 fixed this well enough, closing for now.

PR #332, opened 2016-06-01
ENH: add signal/positioner mixin

klauer:
Alternative to PVPositioner modification idea (#331); ~~only for discussion~~

Could be used as follows:
```python
   sig = EpicsSignalPositioner('Prefix:Motor.RBV', write_pv='Prefix:Motor.VAL', tolerance=0.005)
   sig.move(5.0)
```
This would use `set` -> `set_and_wait` for moves on a signal, which boils down to `abs(readback - setpoint) <= tolerance` (ignoring the unset relative tolerance), so once at 4.995 <= position <= 5.005 would run the done move callbacks, etc.

### DIFF
--- a/ophyd/mixins.py
+++ b/ophyd/mixins.py
@@ -1,3 +1,4 @@
+import time
 import logging
 
 from .signal import EpicsSignal
@@ -19,12 +20,15 @@ class SignalPositionerMixin(PositionerBase):
     set_func : callable
         The set() functionality for the class. Must be specified as the mixin
         takes over set() functionality.
+    readback_event : str, optional
+        Readback value subscription event_type
     egu : str, optional
         Engineering units of positioner
     hold_on_stop : bool, optional
         When stop is called on the positioner
     '''
-    def __init__(self, *args, set_func, egu='', hold_on_stop=False, **kwargs):
+    def __init__(self, *args, set_func, readback_event='value', egu='',
+                 hold_on_stop=False, **kwargs):
         super().__init__(*args, **kwargs)
         self._egu = egu
         self._hold_on_stop = hold_on_stop
@@ -34,7 +38,7 @@ class SignalPositionerMixin(PositionerBase):
         # bind method to this class instance:
         self._mixed_set = set_func.__get__(self, self.__class__)
 
-        self.subscribe(self._position_updated)
+        self.subscribe(self._position_updated, event_type=readback_event)
 
     def _position_updated(self, value=None, **kwargs):
         # for readback subscriptions
@@ -95,10 +99,11 @@ class SignalPositionerMixin(PositionerBase):
         self._reset_sub(self._SUB_REQ_DONE)
 
         self._moving = True
-        self._run_subs(sub_type=self.SUB_START)
+        self._run_subs(sub_type=self.SUB_START, timestamp=time.time())
 
         def finished():
             success = self._internal_status.success
+            self._moving = False
             self._done_moving(success=success)
 
             if moved_cb is not None:
@@ -127,7 +132,7 @@ class SignalPositionerMixin(PositionerBase):
 
         if wait:
             try:
-                status_wait(self._internal_status)
+                status_wait(self._external_status)
             except RuntimeError:
                 raise RuntimeError('Motion did not complete successfully')
 
--- a/ophyd/mixins.py
+++ b/ophyd/mixins.py
@@ -1,0 +1,125 @@
+import types
+from .signal import EpicsSignal
+from .positioner import PositionerBase
+
+
+class SignalPositionerMixin(PositionerBase):
+    '''Mixin to make a Signal a Positioner
+
+    Should be mixed in first, with the Signal second, such that:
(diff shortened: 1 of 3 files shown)


danielballan:
Seems useful.

ophyd/mixins.py
```diff
@@ -0,0 +1,125 @@
...
+    ----------
+    set_func : callable
+        The set() functionality for the class. Must be specified as the mixin
+        takes over set() functionality.
+    egu : str, optional
+        Engineering units of positioner
+    hold_on_stop : bool, optional
+        When stop is called on the positioner
+    '''
+    def __init__(self, *args, set_func=None, egu='', hold_on_stop=False,
```
tacaswell: If you drop the `None` it will be a required kwarg

```python

In [35]: def foo(*args, k, b=5): pass

In [36]: foo('a')
TypeError                                 Traceback (most recent call last)
<ipython-input-36-7407a2d90da0> in <module>()
----> 1 foo('a')

TypeError: foo() missing 1 required keyword-only argument: 'k'
```

ophyd/mixins.py
```diff
@@ -0,0 +1,125 @@
...
+
+            if moved_cb is not None:
+                moved_cb(obj=self)
+
+            self._status = None
+
+        # set() functionality depends on the signal
+        self._status = self._mixed_set(self, position, timeout=timeout,
+                                       settle_time=self.settle_time)
+        self._status.finished_cb = finished
```
tacaswell: I don't think this will work, bluesky needs this callback to act as the bridge to trip the asyncio events which it is waiting on.

ophyd/mixins.py
```diff
@@ -0,0 +1,125 @@
...
+        position : any
+        '''
+        return self.get()
+
+    @property
+    def egu(self):
+        '''The engineering units (EGU) for positions'''
+        return self._egu
+
+    def move(self, position, moved_cb=None, timeout=None):
```
tacaswell: This is missing a wait kwarg?

ophyd/mixins.py
```diff
@@ -0,0 +1,125 @@
...
+        def finished():
+            self._done_moving(success=self._status.success)
+
+            if moved_cb is not None:
+                moved_cb(obj=self)
+
+            self._status = None
+
+        # set() functionality depends on the signal
+        self._status = self._mixed_set(self, position, timeout=timeout,
```
tacaswell: probably better to use `types.MethodType` here?

tacaswell:
I like this idea

ophyd/mixins.py
klauer: Reference?

ophyd/mixins.py
klauer: It is actually bound in the initializer, this line is broken here (sorry)

ophyd/mixins.py
tacaswell: https://github.com/NSLS-II/bluesky/blob/master/bluesky/run_engine.py#L1517

The solution is just to have 2 status objects here, one to hand to BS, one to use internally (and the internal one trips the external one).

ophyd/mixins.py
```diff
@@ -0,0 +1,151 @@
...
+                status_wait(self._internal_status)
+            except RuntimeError:
+                raise RuntimeError('Motion did not complete successfully')
+
+        return self._external_status
+
+    def stop(self):
+        '''Stops motion'''
+        if self._hold_on_stop:
+            self.move(self.get(), wait=False)
```
tacaswell: is this an accepted way to stop things?

ophyd/mixins.py
```diff
@@ -0,0 +1,151 @@
...
+            self._internal_status = None
+            self._external_status = None
+
+        # this external status object is fully dependent on the internal status
+        # object and does not have its own timeout/settle_time settings:
+        self._external_status = MoveStatus(self, target=position,
+                                           settle_time=0.0, timeout=None)
+
+        # set() functionality depends on the signal
+        self._internal_status = self._mixed_set(position, timeout=timeout,
```
tacaswell: Why can't this be done with `super`?

tacaswell:
:+1:

tacaswell:
modulo playing with it

ophyd/mixins.py
klauer: If there's nothing but a setpoint and a readback, what else are you supposed to do?

ophyd/mixins.py
klauer: Both `PositionerBase` and `EpicsSignal` have `set()` which do two different things. `set()` for this new Positioner type should call `move()`, but internally `move()` calls the `Signal.set()` (which calls `set_and_wait`) which is the whole point of the mixin. That is,

`Mixin.set()` -> `Mixin.move()` -> `Signal.set()`

Can this be done with `super()`?

klauer:
@tacaswell if that's the case, then I'll add a test first

klauer:
Test added; ready for review

PR #333, opened 2016-06-02
 FIX: pyepics threading tweaks for EpicsSignal

klauer:
Seeing some threading issues at SRX with the global cache from pyepics.
Completely untested, but a potential fix.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -537,10 +537,9 @@ class EpicsSignalBase(Signal):
         with self._lock:
             desc['units'] = self._read_pv.units
 
-        if hasattr(self, '_write_pv'):
-            with self._lock:
-                desc['lower_ctrl_limit'] = self._write_pv.lower_ctrl_limit
-                desc['upper_ctrl_limit'] = self._write_pv.upper_ctrl_limit
+        low_limit, high_limit = self.limits
+        desc['lower_ctrl_limit'] = low_limit
+        desc['upper_ctrl_limit'] = high_limit
 
         if self.enum_strs:
             desc['enum_strs'] = list(self.enum_strs)
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1,6 +1,7 @@
 # vi: ts=4 sw=4
 import logging
 import time
+import threading
 
 import numpy as np
 import epics
@@ -331,6 +332,7 @@ class EpicsSignalBase(Signal):
         if pv_kw is None:
             pv_kw = dict()
 
+        self._lock = threading.RLock()
         self._read_pv = None
         self._string = bool(string)
         self._pv_kw = pv_kw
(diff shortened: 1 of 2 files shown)


klauer:
Needs a more thorough review, including unchanged lines that should be locked

ophyd/signal.py
```diff
@@ -475,12 +483,13 @@ def get(self, *, as_string=None, **kwargs):
         if as_string is None:
             as_string = self._string
 
-        if not self._read_pv.connected:
-            if not self._read_pv.wait_for_connection():
-                raise TimeoutError('Failed to connect to %s' %
-                                   self._read_pv.pvname)
+        with self._lock:
+            if not self._read_pv.connected:
+                if not self._read_pv.wait_for_connection():
```
danielballan: Should the `timeout` be propagated here as well. Perhaps it would be better to make it configurable via a separate kwargs (`configuration_timeout`)?

danielballan: ^ edited confusing typo

ophyd/signal.py
```diff
@@ -780,15 +793,16 @@ def put(self, value, force=False, **kwargs):
...
 
-        if not self._write_pv.connected:
-            if not self._write_pv.wait_for_connection():
-                raise TimeoutError('Failed to connect to %s' %
-                                   self._write_pv.pvname)
-
-        use_complete = kwargs.pop('use_complete', self._put_complete)
+        with self._lock:
+            if not self._write_pv.connected:
+                if not self._write_pv.wait_for_connection():
```
danielballan: ditto

ophyd/signal.py
```diff
@@ -780,15 +793,16 @@ def put(self, value, force=False, **kwargs):
...
+                    raise TimeoutError('Failed to connect to %s' %
+                                       self._write_pv.pvname)
 
-        self._write_pv.put(value, use_complete=use_complete,
-                           **kwargs)
+            use_complete = kwargs.pop('use_complete', self._put_complete)
```
danielballan: Off-topic for this PR, but as long as you're touching this, would you mind making it a proper kwarg for readability's sake?

klauer:
Tweaked according to requests. 
Any comments on the actual threading modifications? cc @tacaswell

ophyd/signal.py
```diff
@@ -447,11 +454,12 @@ def limits(self):
         '''The read PV limits'''
 
         # This overrides the base limits
-        pv = self._read_pv
-        pv.get_ctrlvars()
+        with self._lock:
+            pv = self._read_pv
+            pv.get_ctrlvars()
```
tacaswell: Does the return need to be in the lock block?

klauer: Yes, good catch

ophyd/signal.py
```diff
@@ -776,19 +798,25 @@ def put(self, value, force=False, **kwargs):
...
+                    raise TimeoutError('Failed to connect to %s' %
+                                       self._write_pv.pvname)
 
-        use_complete = kwargs.pop('use_complete', self._put_complete)
+            if use_complete is None:
+                use_complete = self._put_complete
 
-        self._write_pv.put(value, use_complete=use_complete,
-                           **kwargs)
+            self._write_pv.put(value, use_complete=use_complete, **kwargs)
```
tacaswell: does `put` not participate in the caching in pyepics?

tacaswell: ignore my last question.

klauer:
I think (hope) this is ready. Should probably be tested at a single beamline before rolling out to all others, though.

tacaswell:
Does the ` self._read_pv.add_callback(self._read_changed,
                                   run_now=self._read_pv.connected)` on L350 need to be in a lock block?

klauer:
@tacaswell I don't think so, but I locked it anyway since you seem to think so

tacaswell:
I'll merge as soon as travis is happy

PR #334, opened 2016-06-04
MNT: make log_pos friendlier

tacaswell:
closes a bug report

### DIFF
--- a/ophyd/commands.py
+++ b/ophyd/commands.py
@@ -449,8 +449,12 @@ def log_pos(positioners=None, extra_msg=None):
     int
         The ID of the logbook entry returned by the logbook.log method.
     """
-    positioners = _normalize_positioners(positioners)
     logbook = get_logbook()
+    if logbook is None:
+        raise RuntimeError("you do not have a logbook setup")
+
+    positioners = _normalize_positioners(positioners)
+
     if extra_msg:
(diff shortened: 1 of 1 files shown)


klauer:
:+1: though it would probably be good to add `RuntimeError` to the docstring "Raises" section.
CI probably doesn't even smoke test this...

tacaswell:
the original issue was https://github.com/NSLS-II/Bug-Reports/issues/110 where they were confused why this was silently passing.

tacaswell:
Turns out it does test it. Adding a real test is on the to do list.

klauer:
Smoke test is failing due to the new exception being raised when a logbook isn't set

danielballan:
bump @tacaswell

danielballan:
The mechanism this was trying to fix has been removed.

ambarb:
Does this mean that `from ophyd import log_pos` will not work now?

ambarb:
Does this mean we can no longer `import log_pos from ophyd`?

https://github.com/NSLS-II/ophyd/search?utf8=%E2%9C%93&q=log_pos&type=

ambarb:
nevermind.  I see now that we were using
`from pyOlog.ophyd_tools import *`  which is local.

PR #337, opened 2016-06-08
ENH: add methods to walk AD plugins

tacaswell:
- find all plugins that ophyd knows about
 - find a plugin by port name
 - validate that the full processing path is know to ophyd
 - turn blocking_callbacks on by default in all Plugin stage methods

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -8,6 +8,7 @@ import networkx as nx
 from ..signal import EpicsSignal
 from . import docs
 from ..device import (Device, Component)
+from ..signal import (ArrayAttributeSignal)
 
 
 class EpicsSignalWithRBV(EpicsSignal):
@@ -254,3 +255,10 @@ class ADBase(Device):
                 ret.append(node)
 
         return ret
+    configuration_names = Component(ArrayAttributeSignal,
+                                    attr='_configuration_names')
+
+    @property
+    def _configuration_names(self):
+        return [getattr(self, c).name
+                for c in self.configuration_attrs]
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -3,6 +3,7 @@ import inspect
 import re
 import sys
 from collections import OrderedDict
+import networkx as nx
 
 from ..signal import EpicsSignal
 from . import docs
@@ -148,3 +149,108 @@ class ADBase(Device):
 
             if match:
                 match_fcn(attr=attr, signal=getattr(self, attr), doc=doc)
+
+    def stage(self, *args, **kwargs):
+        if not self.validate_asyn_ports():
+            missing_plugins = self.missing_plugins()
+            raise RuntimeError('The asyn ports {!r} are used by plugins '
+                               'that ophyd is aware of but the source plugin '
+                               'is not.  Please reconfigure your device to '
+                               'include the source plugin or reconfigure '
+                               'to not use these ports.'
+                               ''.format(missing_plugins))
+        return super().stage(*args, **kwargs)
+
+    def get_plugin_by_asyn_port(self, port_name):
+        '''Get the plugin which has the given asyn port name
+
+        Parameters
+        ----------
+        port_name : str
+            The port name to search for
+
+        Returns
+        -------
+        ret : ADBase or None
+            Either the requested plugin or None if not found
+
+        '''
+        try:
+            name = self.port_name.get()
+        except AttributeError:
+            pass
+        else:
+            if name == port_name:
+                return self
+
+        for nm in self._sub_devices:
+            cpt = getattr(self, nm)
+            if hasattr(cpt, 'get_plugin_by_asyn_port'):
+                sig = cpt.get_plugin_by_asyn_port(port_name)
+                if sig is not None:
+                    return sig
+        return None
+
+    def get_asyn_port_dictionary(self):
+        '''Return port name : component map
+
+        Returns
+        -------
+        port_map : dict
+            Mapping between port_name and ADBase objects
+        '''
+        # uniqueness of port names enforced at IOC layer
+        ret = {}
+        try:
+            ret.update({self.port_name.get(): self})
+        except AttributeError:
+            pass
+
+        for nm in self._sub_devices:
+            sig = getattr(self, nm)
+            if hasattr(sig, 'get_asyn_port_dictionary'):
+                ret.update(sig.get_asyn_port_dictionary())
+
+        return ret
+
+    def get_asyn_digraph(self):
+        '''Get the directed graph of the ASYN ports
+
+        Returns
+        -------
+        G : networkx.DiGraph
+            Directed graph of pipelines
+
+        port_map : dict
+            Mapping between port_name and ADBase objects
+        '''
+        port_map = self.get_asyn_port_dictionary()
+        G = nx.DiGraph()
+        for out_port, cpt in port_map.items():
+            try:
+                in_port = cpt.nd_array_port.get()
+            except AttributeError:
+                # attribute error because we hit a component which is not
+                # a plugin, but is the 'base' data source
+                G.add_node(out_port)
+            else:
+                G.add_edge(in_port, out_port)
+
+        return G, port_map
+
+    def validate_asyn_ports(self):
+        g, port_map = self.get_asyn_digraph()
+        g = nx.Graph(g)
+        if nx.number_connected_components(g) != 1:
+            return False
+        return True
+
+    def missing_plugins(self):
+        g, port_map = self.get_asyn_digraph()
+        ret = []
+
+        for node in g:
+            if node not in port_map:
+                ret.append(node)
+
+        return ret
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -83,7 +83,8 @@ class PluginBase(ADBase):
 
     _default_configuration_attrs = ('port_name', 'nd_array_port', 'enable',
                                     'blocking_callbacks', 'plugin_type',
-                                    'asyn_pipeline_config', 'configuration_names')
+                                    'asyn_pipeline_config',
+                                    'configuration_names')
 
     _html_docs = ['pluginDoc.html']
     _plugin_type = None
@@ -210,7 +211,6 @@ class ImagePlugin(PluginBase):
     _plugin_type = 'NDPluginStdArrays'
 
     array_data = C(EpicsSignal, 'ArrayData')
-    _default_configuration_attrs = PluginBase._default_configuration_attrs
 
     @property
     def image(self):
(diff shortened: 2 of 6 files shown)


klauer:
Need to add networkx to dependency list

ophyd/areadetector/base.py
```diff
@@ -148,3 +149,110 @@ def default_match(attr, signal, doc):
...
+        signames = self._sub_devices
+
+        for nm in signames:
+            sig = getattr(self, nm)
+            if hasattr(sig, 'get_port_dictionary'):
+                ret.update(sig.get_port_dictionary())
+
+        return ret
+
+    def get_asyn_digraph(self):
```
klauer: (`port` or `asyn`) -> `asyn_port` everywhere, or these method names won't make much sense

ophyd/areadetector/plugins.py
```diff
@@ -106,6 +112,15 @@ def array_pixels(self):
 
         return pixels
 
+    def read_configuration(self):
+        ret = super().read_configuration()
+
+        source_port = self.nd_array_port.get()
+        source_plugin = self.root.get_plugin_by_port(source_port)
+        ret[source_port] = source_plugin.read_configuration()
```
klauer: I like this idea, but won't this turn into:
{'port': dict(attr=value), ...}
where we actually want just a single merged dictionary?

tacaswell: The idea was to to directly encode the ordering in the nesting.  Get the configuration of the parent via `dd[dd['nd_array_port']]`.

I do not think we want to merge these dictionaries as I think eventually will have `understand_what_stats_plugin_was_doing()` functions on the analysis side and it would be nice to not have to allow those to accept (but ignore) extra keys.

ophyd/areadetector/base.py
tacaswell: fair.  I am bad at naming things.

ophyd/areadetector/plugins.py
klauer: ... but this breaks the `read_configuration` API? Call it something else?

tacaswell: how does this break the API?

tacaswell: This breaks api because we do not have a way of doing nested descriptors.

ophyd/quadem.py
```diff
@@ -16,7 +16,9 @@ def _current_fields(attr_base, field_base, range_, **kwargs):
 
     return defn
 
+
 class QuadEM(SingleTrigger, DetectorBase):
+    port_name = Cpt(Signal, value='NSLS2_EM')
```
klauer: Mind leaving a comment here as to why this is necessary?

It's also a bit concerning that the QuadEM IOC is re-used for other devices than the NSLS2-EM. I don't know what their port names are, nor if they actually have an EPICS PV in their databases with the port name exposed...

klauer:
Automatically making mbbi/mbbo and bi/bo records (the ones you would consider to be enums) use `string=True` is something I wish I had done with the class generation from template. Thanks for tweaking some of them.

tests/test_areadetector.py
```diff
@@ -116,6 +117,64 @@ class MyDetector(SimDetector):
...
+
+    assert det.tiff1 is det.get_plugin_by_asyn_port(det.tiff1.port_name.get())
+    assert det.cam is det.get_plugin_by_asyn_port(det.cam.port_name.get())
+    assert det.roi1 is det.get_plugin_by_asyn_port(det.roi1.port_name.get())
+
+
+def test_read_configuration_smoke():
+    class MyDetector(SingleTrigger, SimDetector):
+        stats1 = Cpt(StatsPlugin, 'Stats1:')
+        porc1 = Cpt(ProcessPlugin, 'ROI1:')
```
klauer: Typo here causing test to fail

ophyd/areadetector/plugins.py
```diff
@@ -106,6 +112,24 @@ def array_pixels(self):
 
         return pixels
 
+    def read_configuration(self):
+        ret = super().read_configuration()
+
+        source_port = self.nd_array_port.get()
+        source_plugin = self.root.get_plugin_by_asyn_port(source_port)
```
klauer: `source_plugin` seems like something that'd be handy to have as a property (or a `get_*` as you seem to prefer)

ophyd/areadetector/base.py
```diff
@@ -148,3 +149,110 @@ def default_match(attr, signal, doc):
...
+                    return sig
+        return None
+
+    def get_asyn_port_dictionary(self):
+        '''Return port name : component map
+
+        Returns
+        -------
+        port_map : dict
+            keyed on port name (str) values are OphydObjects
```
klauer: Yes, they are `OphydObj`ects, but I think we can be a bit more specific here

tests/test_areadetector.py
tacaswell: > (the ones you would consider to be enums)

I found them by looking for descriptions that and an `enum_str` field in them.  Is calling them enums not fair?

klauer:
Overall 👍 Great addition to the current functionality

tests/test_areadetector.py
klauer: It's fair given the view ophyd gives - but they're represented by different records at the EPICS level. So from that point of view, calling them all just enums is strange.

danielballan:
+1 for our first usage of networkx

tacaswell:
@klauer The second half of this is ready to review.

ophyd/areadetector/base.py
```diff
@@ -148,3 +150,117 @@ def default_match(attr, signal, doc):
...
+
+        Returns
+        -------
+        ret : ADBase or None
+            Either the requested plugin or None if not found
+
+        '''
+        try:
+            name = self.port_name.get()
+            if name == port_name:
```
klauer: Bike-shedding: move this out of the try/except block, `signames` variable could be easily removed here and below (unless you get a bonus for line counts then by all means leave it in)

ophyd/areadetector/base.py
```diff
@@ -148,3 +150,117 @@ def default_match(attr, signal, doc):
...
+
+        port_map : dict
+            Mapping between port_name and ADBase objects
+        '''
+        port_map = self.get_asyn_port_dictionary()
+        G = nx.DiGraph()
+        for out_port, cpt in port_map.items():
+            try:
+                in_port = cpt.nd_array_port.get()
+                G.add_edge(in_port, out_port)
```
klauer: `G.add_edge(...)` should go in an `else` clause to be clear what the `except AttributeError` is trying to catch

ophyd/areadetector/base.py
```diff
@@ -148,3 +150,117 @@ def default_match(attr, signal, doc):
...
+    def missing_plugins(self):
+        g, port_map = self.get_asyn_digraph()
+        ret = []
+
+        for node in g:
+            if node not in port_map:
+                ret.append(node)
+
+        return ret
+    config_names = Component(AttributeSignal, '_configuration_names',
```
klauer: `configuration_attrs` but `config_names`... probably should keep it consistent?

tacaswell: so consistent long it is

klauer: `add_prefix` is also unnecessary, assuming you use `attr='_configuration_names'` instead of the positional argument here

ophyd/areadetector/plugins.py
```diff
@@ -59,14 +59,30 @@ def register_plugin(cls):
...
+        if configuration_attrs is None:
+            configuration_attrs = self._default_configuration_attrs
         # Turn array callbacks on during staging.
         # Without this, no array data is sent to the plugins.
-        super().__init__(*args, **kwargs)
-
+        super().__init__(*args, configuration_attrs=configuration_attrs,
+                         **kwargs)
+        # make sure it is the right type of plugin
+        if not self.plugin_type.get().startswith(self._plugin_type):
```
klauer: It may be useful to remove this check in some cases. I should be able to use `PluginBase` with any plugin and not always the exact plugin class.

To that end, can you add a `if self._plugin_type is not None and ...` before this?

ophyd/areadetector/plugins.py
```diff
@@ -106,6 +122,43 @@ def array_pixels(self):
...
 
+    def read_configuration(self):
+        ret = super().read_configuration()
+
+        ret.update(self.source_plugin.read_configuration())
+
+        return ret
+
+    @property
+    def source_plugin(self):
```
klauer: A property - great idea! 👍 But please include docstrings...

ophyd/areadetector/plugins.py
```diff
@@ -106,6 +122,43 @@ def array_pixels(self):
...
+        source_port = self.nd_array_port.get()
+        source_plugin = self.root.get_plugin_by_asyn_port(source_port)
+        return source_plugin
+
+    def describe_configuration(self):
+        ret = super().describe_configuration()
+
+        source_port = self.nd_array_port.get()
+        source_plugin = self.root.get_plugin_by_asyn_port(source_port)
```
klauer: Use the source_plugin property here?

ophyd/areadetector/base.py
tacaswell: I changed this, but I do not think in the way you meant.

ophyd/utils/epics_pvs.py
```diff
@@ -428,11 +428,12 @@ def _compare_maybe_enum(a, b, enums, atol, rtol):
 
 
 _type_map = {'number': (float, ),
-             'array': (np.ndarray, ),
+             'array': (np.ndarray, list),
```
klauer: I guess this is fine, but I have a feeling it was working around what `ArrayAttributeSignal` was made for: https://github.com/NSLS-II/ophyd/blob/master/ophyd/signal.py#L941

tacaswell: and I actually bet that on return this is going to be turned into an object array as we upcast everything that is 'dtype' array to `np.array`.

I should either revert these changes and use `ArrayAttributeSignal` or we make list a first class thing in the schema

ophyd/areadetector/base.py
klauer:
```python
        try:
            name = self.port_name.get()
        except AttributeError:
            pass

        if name == port_name:
            return self

        for nm in self._sub_devices:
            cpt = getattr(self, nm)
            if hasattr(cpt, 'get_plugin_by_asyn_port'):
                sig = cpt.get_plugin_by_asyn_port(port_name)
                if sig is not None:
                    return sig
        return None
```

klauer:
Passes travis finally 👍

klauer:
My last remaining quibble: `_default_configuration_attrs` is not being checked for all plugins. There should be a test that ensures all of the default attributes exist.

tacaswell:
Why did we lose codecov in this?

klauer:
👍

PR #339, opened 2016-06-20
CI: auto-build docs

### LINKED ISSUE #298: Auto build and upload ophyd docs

klauer:
Closes #298

cc @ericdill (mind verifying this is correct?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,14 @@ before_install:
   - export EPICS_CA_MAX_ARRAY_BYTES=10000000
   - export DOCKERIMAGE="klauer/epics-docker"
   - export CONDA_ENV="testenv"
+  - export BUILD_DOCS_BRANCH="master"
 
   - perl --version
   - git fetch --unshallow
+
+  # include nsls2's ci scripts
+  - git clone https://github.com/nsls-ii/nsls2-ci $HOME/ci-scripts
+
   - docker pull ${DOCKERIMAGE}
   - docker images
   - docker ps -a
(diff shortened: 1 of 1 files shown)


ericdill:
Thanks for doing this!  Few comments.

1. Consider inserting another step in the build matrix, so you don't build the docs for both versions of python. It becomes difficult to inspect the .travis.yml log file if you build the docs since it adds ~5000 lines of output.  If you do decide to go down this route, you should also consider *not* running the test suite in the same run that you build the docs.  i.e., wrap all the test-related things with 
```bash
if [ BUILDING_DOCS == "true" ]; then
    build_docs_stuff...;
else
   do_automated_testing_stuff...;
fi;
```

2. You need to upload a github authentication token so you can push back to github: https://docs.travis-ci.com/user/environment-variables/. 

tl;dr on the encryption:

```bash
sudo gem install travis
travis encrypt GITHUB_TOKEN=<your github token> -r nsls-ii/ophyd
```
Copy the output of `travis encrypt` into the .travis.yml and then use the `GITHUB_TOKEN` env var as you would normally in the .travis.yml

3. Separate the build step and the upload step.  The build docs step should probably become part of the `script:` section so that we know if it fails. Leave the upload step in the `after_success:` section.

ericdill:
You can get rid of lines 40-52 with `- . ~/scripts/install-miniconda.sh`

ericdill:
Don't need `chmod +x do-the-docs.sh ` since they are already executable in the git repo (I think...)

danielballan:
The /docs and /bluesky repos are now auto-building.

I reused almost all of Eric's nsls2-ci code, but I did take a different approach to auth. There are separate "deploy keys" created on the NSLS-II.github.io repository -- one separate key for each repo that pushes the docs.

I'm going to close this PR and set up ophyd the same way I set up bluesky. Thanks for your efforts here, everyone.

danielballan:
Sweet. The most recent Travis build of ophyd master pushed commit a355f1e to NSLS-II/NSLS-II.github.io. Success.

PR #340, opened 2016-06-22
FIX: untested attributesignal attributeerror exception

klauer:
Don't merge yet - still am not getting coverage

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -918,7 +918,7 @@ class AttributeSignal(Signal):
             try:
                 obj = getattr(obj, part)
             except AttributeError as ex:
-                raise AttributeError('{} ({})'.format(part, ex))
+                raise AttributeError('{}.{} ({})'.format(obj.name, part, ex))
 
         return obj
 
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -918,8 +918,7 @@ class AttributeSignal(Signal):
             try:
                 obj = getattr(obj, part)
             except AttributeError as ex:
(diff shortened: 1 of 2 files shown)


tests/test_device.py
```diff
@@ -235,6 +236,17 @@ def prop(self, value):
...
 
 
+def test_attribute_signal_attributeerror():
+    class MyDevice(Device):
+        sig1 = Component(AttributeSignal, 'unknownattribute')
+        sig2 = Component(AttributeSignal, '__class__.__name')
+
+    dev = MyDevice('', name='mydev')
+    pytest.raises(AttributeError, dev.sig1.get)
+    pytest.raises(AttributeError, getattr(dev.sig1, 'base'))
```
tacaswell: I think this needs to be `pytest.raises(AtributeError, getattr, dev.sig1, 'base')`?

klauer: I do this all the time - it's annoying

klauer:
This took me more (distracted) tries than I'd like to admit. Ready for review.

ophyd/signal.py
```diff
@@ -918,8 +918,7 @@ def base(self):
             try:
                 obj = getattr(obj, part)
             except AttributeError as ex:
-                attr = '.'.join(self.parent_attr[:i + 1])
-                raise AttributeError('{} ({})'.format(attr, ex))
+                raise AttributeError('{} ({})'.format(part, ex))
```
tacaswell: :bike: :house: is it worth including `obj` here as well?

PR #347, opened 2016-07-27
FIX: For QuadEM the port_name must be different

hhslepicka:
for each type of device (NSLS_EM, TetrAMM, APS_EM, etc).


arkilic:
pele!

PR #349, opened 2016-07-29
ENH: Avoid the use of wrong PVs prefix when defining a motor

hhslepicka:
The following situation happened at LIX beamline:
```
sh=EpicsMotor('XF:16IDC-ES:Sol{Enc-Ax:XL}Mtr.VAL', name='hand_low_encl')
```

This resulted on every kind of weird issues and also one accident with a motor.
This simple fix will block this sort of situation.

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -40,6 +40,7 @@ class EpicsMotor(Device, PositionerBase):
     timeout : float, optional
         The default timeout to use for motion requests, in seconds.
     '''
+    record_type = Cpt(EpicsSignalRO, '.RTYP', string=True)
     user_offset = Cpt(EpicsSignal, '.OFF')
     user_readback = Cpt(EpicsSignalRO, '.RBV')
     user_setpoint = Cpt(EpicsSignal, '.VAL', limits=True)
@@ -68,6 +69,9 @@ class EpicsMotor(Device, PositionerBase):
         super().__init__(prefix, read_attrs=read_attrs,
                          configuration_attrs=configuration_attrs,
                          name=name, parent=parent, **kwargs)
+
+        if self.record_type.get() != "motor":
+            raise ValueError("Prefix: {} is not a valid Epics Motor Record.")
 
         # Make the default alias for the user_readback the name of the
         # motor itself.
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -71,7 +71,7 @@ class EpicsMotor(Device, PositionerBase):
                          name=name, parent=parent, **kwargs)
 
         if self.record_type.get() != "motor":
-            raise ValueError("Prefix: {} is not a valid Epics Motor Record.")
+            raise ValueError("Prefix: {} is not a valid Epics Motor Record.".format(prefix))
 
         # Make the default alias for the user_readback the name of the
         # motor itself.
--- a/tests/test_epicsmotor.py
+++ b/tests/test_epicsmotor.py
@@ -41,6 +41,19 @@ def test_timeout(motor):
     assert motor.timeout == 20.0
 
 
+def test_record_type():
+    sim_pv = 'XF:31IDA-OP{Tbl-Ax:X1}Mtr'
+
+    m = TestEpicsMotor(sim_pv, name='epicsmotor', settle_time=0.1,
+                       timeout=10.0)
+
+    sim_pv = 'XF:31IDA-OP{Tbl-Ax:X1}Mtr.VAL'
+
+    with pytest.raises(ValueError):
+        TestEpicsMotor(sim_pv, name='epicsmotor', settle_time=0.1,
+                       timeout=10.0)
+
+
 def test_connected(motor):
     assert motor.connected
 


arkilic:
👍 Checking for motor record

hhslepicka:
Attn. @arkilic @klauer

klauer:
I like this sort of check, but the downside is that now you're blocking in the initializer:
It will slow down motor-record-heavy beamlines and also raise when it can't connect to one.

hhslepicka:
I'm not sure about the amount of added overhead given how small this
verification is. Better safe than sorry, this almost cost us a sample
holder. If the PV name is wrong, the user receives an error in ipython
shell.

klauer:
I don't disagree about the importance of verifying it, but now we have a bunch of composite devices now made up of many `EpicsMotor`s (and otherwise) on almost every beamline. Turn off any one of those IOCs even temporarily, start your collection environment, and all of a sudden your whole data collection environment is broken until you comment out those lines.

There are some alternatives that come to mind - checking the record type in the `raise_if_disconnected` decorator (which calls `EpicsMotor.connected` that could be customized easily) and stashing the result, or maybe even keeping some persistent (i.e., between session) cache of PV name to record type.

My opinion remains that a synchronous check in the `__init__` method is the wrong way to do this in our architecture. Any other opinions for or against, @tacaswell @danielballan?

tacaswell:
I am with @klauer that this should be done as asynchronously as possible.

Something that would also solve this problem is detecting that we have two `.` in a pvname and raising (as epics seems to silently drop everything after the second period).  @hhslepicka Tried to explain the details of this to me over chat today, but I might still not understand.

hhslepicka:
I agree with the check proposed by Tom but I would also like to have the
RTYP check to ensure that it is really a motor.
I agree with you too, @klauer we can talk about he correct way to implement
it on the following week.
Thank you all for your input on this matter.

cowanml:
How did we almost lose a sample holder?

Protections should be at the lowest level feasible, not the top level.
Lower level stuff should be protecting things?

hhslepicka:
@cowanml and @dchabot,
The sample holder is composed by two stages (one for the needles and the
other one for the samples).
The needles were down and I have no idea which command was sent to the
sample motor which reflected on the ".VAL" pv that resulted to a movement
to the side while the needles were inserted.

I know and understand that protection needs to be in the lowest level as
possible (Hardware) but it would also be good to have this sort of
protection on the software as an additional layer.
I'm not saying that it was a fault from the software, it was a human error
and it could be avoided in many ways. =)
I was just trying to give my 2 cents on a way to avoid this type of mistake
again.

klauer:
At which level would you like this fixed?

```sh
$ caget XF:31IDA-OP{Tbl-Ax:X1}Mtr.RBV
XF:31IDA-OP{Tbl-Ax:X1}Mtr.RBV  0
$ caget XF:31IDA-OP{Tbl-Ax:X1}Mtr.RBV.VAL
XF:31IDA-OP{Tbl-Ax:X1}Mtr.RBV.VAL 0
$ caget XF:31IDA-OP{Tbl-Ax:X1}Mtr.RBV.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL
XF:31IDA-OP{Tbl-Ax:X1}Mtr.RBV.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL 0
$ caget XF:31IDA-OP{Tbl-Ax:X1}Mtr.EGU.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL
XF:31IDA-OP{Tbl-Ax:X1}Mtr.EGU.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL.VAL deg
```

hhslepicka:
All, I will just try to explain better here: 
- This was not an issue with motors only
- The problem is that PV_NAME_Mtr.RBV.VAL is valid for Channel Access and connected will be true as in the example given by @klauer.
- The proposal was to ensure that the user is sending a RTYP == motor when creating the EpicsMotor object.
- Yes, I know that the collision of the needle and motor was a fault of not having it protected by hardware.

Since we have now the PR #350 by @tacaswell, which is a more general fix, I'm closing this PR and moving forward! 

Thanks! :+1:

tacaswell:
@hhslepicka I think that having a non-blocking way to verify that a device is talking to the right record is a good idea in general (see https://github.com/NSLS-II/ophyd/blob/master/ophyd/areadetector/plugins.py#L70 (and this is also in the `__init__`, but plugin objects only ever show up an components so are lazily created when you try to access them so this will not blowup IPython start up))

arkilic:
@dchabot we are now on github  👻 
https://github.com/epics-base/epics-base/blob/3.16/src/ioc/dbStatic/dbStaticLib.c#L1515

PR #351, opened 2016-08-01
ENH: Allow stage_sigs keys to be attr names, not attrs.

danielballan:
Speaking of lazy instantiation, devices with `stage_sigs` typically break lazy instantiation by including references to signals in the `__init__`, such as:

```python
class AD(AreaDetector):

    def __init__(self, ...):
        super().__init__(...)
        self.stage_sigs[self.cam.whatever] = 5
```

This PR allows the keys in stage sigs to be strings, like so:

```python
class AD(AreaDetector):

    def __init__(self, ...):
        super().__init__(...)
        self.stage_sigs['cam.whatever'] = 5
```

Here, the attribute lookup / signal connection happens during `stage()` instead of during `__init__`. This should significantly speedup reload times for IPython profile scripts that include many area detectors.

The old usage is still allowed. In cases where the key refers to something that is not an attribute of `self`, it will still be the only option.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -424,8 +424,19 @@ class BlueskyInterface:
         logger.debug("Staging %s", self.name)
         self._staged = Staged.partially
 
+        # Resolve any stage_sigs keys given as strings: 'a.b' -> self.a.b
+        stage_sigs = OrderedDict()
+        for k, v in self.stage_sigs.items():
+            if isinstance(k, str):
+                attr = self
+                for attrname in k.split('.'):
+                    attr = getattr(attr, attrname)
+                stage_sigs[attr] = v
+            else:
+                stage_sigs[k] = v
+
         # Read current values, to be restored by unstage()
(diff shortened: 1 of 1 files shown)


ophyd/device.py
```diff
@@ -424,8 +424,19 @@ def stage(self):
         logger.debug("Staging %s", self.name)
         self._staged = Staged.partially
 
+        # Resolve any stage_sigs keys given as strings: 'a.b' -> self.a.b
+        stage_sigs = OrderedDict()
+        for k, v in self.stage_sigs.items():
+            if isinstance(k, str):
+                attr = self
```
klauer: `Device.__getattr__` takes care of this stuff already - pass it the dotted name directly

danielballan: Thanks. I thought you had done that somewhere, but I was looking under utils and couldn't find it there.

danielballan:
The "landscape" check seems to be hung, but codecov and Travis are happy.

PR #353, opened 2016-08-02
REF: pseudosingle is now a Device

klauer:
Allows `setpoint` and `readback` to be read as normal `Device` `Components`

Closes #344

PR #358, opened 2016-09-02
ENH: pseudo lock

klauer:
Two additional features: 

Per-PseudoSingle, use initial inverse-calculated position as target (setpoint)
Per-PseudoPositioner, auto-target: automatically set the target (setpoint) of all PseudoSingles when a full PseudoPosition is specified (e.g., setting an hkl position would set the targets of the three pseudo axes)

cc @cmazzoli @ambarb


klauer:
Only consideration is that it probably should have been rebased on top of #353 but I forgot it existed....

PR #359, opened 2016-09-02
Enh pseudo lock

danielballan:
#358 rebased on master (with #353 now merged)

### DIFF
--- a/ophyd/pseudopos.py
+++ b/ophyd/pseudopos.py
@@ -45,18 +45,20 @@ class PseudoSingle(Device, SoftPositioner):
     source : str, optional
         Metadata indicating the source of this positioner's position. Defaults
         to 'computed'
+    target_initial_position : bool, optional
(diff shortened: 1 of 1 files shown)


danielballan:
I would like to source-install this somewhere and try it out on a pseudopos before we roll out everywhere.

tacaswell:
Does @NSLS-II/csx volunteer?

ambarb:
If okay with Claudio, we can do it today (thought we both are busy 2:45 - 3:45.  Otherwise, you can install it remotely on Friday, but we will not be able to test it until the facilities work is complete in the user area (probably Tuesday of next week).

danielballan:
If you two can figure out how to source install that branch of ophyd, go for it. If you need help it will have to wait for next week.

danielballan:
Merging (these commits are Ken's, not mine)

PR #360, opened 2016-09-08
FIX: Renaming the read PV that was shadowing the read method.

hhslepicka:


hhslepicka:
Attn. @tacaswell !

danielballan:
Ha. Good catch.

I briefly considered, "Maybe we should ensure, at the metaclass level, that no component names can shadow a required part of the BlueskyInterface." I'm leaning against that... my bias is toward allowing users to shoot themselves in the foot if they want to... thoughts?

danielballan:
Merging as is for the bug fix, but let discussion continue on my last point.

klauer:
@danielballan we... er, you, already do that for Device, but not BlueskyInterface: https://github.com/NSLS-II/ophyd/blob/master/ophyd/device.py#L330

Might be worthwhile to consider extending that... I get confused when I shoot myself in the foot.

hhslepicka:
@danielballan I agree with @klauer... This case was simple but there can be some time wasted on debug of an issue that could be solved with a nice error message!

@klauer Welcome back! 😜

danielballan:
Yeah, fair. I should have gone with my first impulse. Even I had to look up the bluesky interface in the docs to remember it all. See #361.

PR #361, opened 2016-09-08
Do not allow component names to shadow any bluesky interface attrs

danielballan:


hhslepicka:
👍  Sweet!

danielballan:
The logic needs to be reworked a bit. The metaclass is seeing the *actual* read method and rejecting it.

danielballan:
@hhslepicka This works now and I added a test. Ready for review.

PR #363, opened 2016-10-04
ENH: Add utility command for stopping all positioners on Ctrl+C.

danielballan:


danielballan:
untested

tacaswell:
I think the intent is that if _any_ exception makes it out to the user as a traceback, then something has gone horribly wrong and we should stop the ~~presses~~ motors.

klauer:
This... scares me a bit.

tacaswell:
Why?

danielballan:
*runs in fear from the ghost of Ken*

danielballan:
I am not a strong advocate of this PR. It pragmatically solves a problem in an admittedly ham-fisted way and it certainly has the potential for unexpected behavior. I would abandon it without much protest, but I would be interested in hearing more detailed reactions from the ex-pats.

klauer:
Some beamlines have a few hundred motors - stopping all of them instead of just the relevant ones for the scan/method may have some unintended consequences:

Imagine one scientist poking at a CSS screen moving an unrelated motor, another on the data collection command line who just caused a syntax error - now that motor stops (along with every other one).

tacaswell:
Maybe limit to just `KeyboardInterupt`?

BL are using the synchronous versions of `obj.move()` from the repl which makes something like this important.

danielballan:
Maybe move registers itself with a module-level registry of things that are known to be moving because this session told them to move?

Get Outlook for iOS<https://aka.ms/o0ukef>




On Tue, Oct 4, 2016 at 7:36 PM -0400, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


Maybe limit to just KeyboardInterupt?

BL are using the synchronous versions of obj.move() from the repl which makes something like this important.

tacaswell:
The _every_ ophyd object needs to do that, the brain need to live higher, either by providing `RE(mv(mtr, pos))` + stern warnings to to use the `move` method or by something like this.

danielballan:
I think it's reasonable for BLs to write a `emergency_stop()` utility but that it shouldn't be implicitly invoked on Ctrl+C, so it does not need to ship with ophyd.

PR #364, opened 2016-10-14
FIX: Fix a issue when a ADBase device has no port name and validate_a…

hhslepicka:
…syn_port is called with it. E.g., OverlayPlugin and Overlay object, the first one has the port but its child no.

Still need to add tests for this case.

Attn. @danielballan


hhslepicka:
@danielballan Now we also have tests for this! \o
This one is good to go!

PR #367, opened 2016-10-27
FIX: allow set_and_wait to not have a timeout

tacaswell:
Fixes issue at XPD


danielballan:
A little background for anyone reading along: Normally a long set_and_wait is scary because it blocks the RunEngine. But EpicsSignalPositioner calls set_amd_wait in a thread. In that case we want the ability to have no timeout. The default timeout for set_and_wait is still 10 seconds.

PR #376, opened 2017-01-25
ENH: add more fields to scaler

tacaswell:


tacaswell:
This should hold on me fixing the tests.  The fake PVs backing the configuartion settings are randomly updating their values.  It passes when we get lucky.

tacaswell:
This should be merged as-is.  I will have another PR re-arranging the scaler more aggressively.

PR #378, opened 2017-01-26
REF: Make use of #351 to make more devices lazy.

danielballan:
In #351, we added the option to specify stage_sigs as attribute *names* as opposed to attributes. This restores the intended lazy-connecting feature of ophyd objects, where they don't need to connect to all their signals at __init__ time.

This PR takes advantage of that in more places in the codebase. It should improve the startup time of IPython at SIX and others.

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -141,10 +141,10 @@ class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
 class FileStorePluginBase(FileStoreBase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.stage_sigs.update([(self.auto_increment, 'Yes'),
-                                (self.array_counter, 0),
-                                (self.auto_save, 'Yes'),
-                                (self.num_capture, 0),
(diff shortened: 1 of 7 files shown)


ophyd/areadetector/filestore_mixins.py
```diff
@@ -249,15 +249,15 @@ def __init__(self, *args, images_per_set_name='images_per_set',
         self._proc_name = proc_name
         cam = getattr(self.parent, self._cam_name)
         proc = getattr(self.parent, self._proc_name)
-        self.stage_sigs.update([(self.file_template, '%s%s_%6.6d.tiff'),
-                                (self.file_write_mode, 'Single'),
-                                (proc.nd_array_port, cam.port_name.get()),
+        self.stage_sigs.update([('file_template', '%s%s_%6.6d.tiff'),
+                                ('file_write_mode', 'Single'),
+                                (nd_array_port, cam.port_name.get()),
```
tacaswell: did this loose a 'proc'?

danielballan: Oh good catch. Yes. Fixed original commit and force-pushed.

danielballan:
Travis is happy and codecov is partly happy.

PR #380, opened 2017-01-26
MNT: In PVPositioner, make sure stop_signal is connected before moving.

danielballan:


tacaswell:
I'm 90% sure someone knocked the IOC over whilst they were running in the case that prompted this, but seems prudent.

PR #383, opened 2017-02-10
Doc round2

tacaswell:


danielballan:
I realize this is still a WIP but (with @tacaswell's OK via Slack) I merged because it's better than what we currently have published, and we're likely to get some readers from Chicago this week.

PR #386, opened 2017-03-08
CI: test py3.6 instead of py3.4

tacaswell:


danielballan:
@licode Have we built all the deps we need for this to work yet?

danielballan:
Superceded by #391.

PR #387, opened 2017-03-10
FIX: Impose upper limit on poll time back-off.

danielballan:
I randomly remembered that we've been meaning to do this for months. Now back to vacation.

### DIFF
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -414,7 +414,8 @@ def set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
         logger.info("Waiting for %s to be set from %r to %r%s...",
                     signal.name, current_value, val, within_str)
         ttime.sleep(poll_time)
-        poll_time *= 2  # logarithmic back-off
+        if poll_time < 0.1:
+            poll_time *= 2  # logarithmic back-off
         current_value = signal.get()
         if expiration_time is not None and ttime.time() > expiration_time:
             raise TimeoutError("Attempted to set %r to value %r and timed "


ophyd/utils/epics_pvs.py
```diff
@@ -414,7 +414,8 @@ def set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
         logger.info("Waiting for %s to be set from %r to %r%s...",
                     signal.name, current_value, val, within_str)
         ttime.sleep(poll_time)
-        poll_time *= 2  # logarithmic back-off
+        if poll_time < 1:
```
tacaswell: I would cut this off at `0.1` instead

danielballan: Changed and force-pushed.

ophyd/utils/epics_pvs.py
```diff
@@ -414,7 +414,8 @@ def set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
         logger.info("Waiting for %s to be set from %r to %r%s...",
                     signal.name, current_value, val, within_str)
         ttime.sleep(poll_time)
-        poll_time *= 2  # logarithmic back-off
+        if poll_time < 0.1:
+            poll_time *= 2  # logarithmic back-off
```
klauer: Is a `poll_time` of 0 valid?
Perhaps something like this would be advisable:
`poll_time = max(poll_time * 2, min_poll_time)`

danielballan: I would rather blow up if the user passes in a poll time below the legal value. Silently "correcting" (ignoring) user input is bad.

PR #388, opened 2017-03-23
BLD: update versioneer to 0.18

tacaswell:


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [versioneer]
 VCS = git
-style = pep440
+style = pep440-post
```
danielballan: What does this change compared so our current configuration? Does it _always_ include the commit in the version? The corresponding PRs in databroker and bluesky didn't have this change, so if we do this we should go back and do it consistently there too.

stuartcampbell: For the other PRs I let versioneer do it's thing.  @tacaswell did you add this explicitly.

tacaswell: https://github.com/warner/python-versioneer/blob/master/details.md#how-do-i-select-a-version-style

This is `TAG.postDISTANCE+gHASH`, the default is `TAG+DISTANCE.gHASH`

I have had strong opinions about this in the past but do not recall why...

danielballan: I don't recall either. Can you pull the `-post` into a separate commit so we can keep the projects consistent until you decide?

danielballan:
Re-opened to regenerate the Travis build matrix because we no longer test on or support 3.4

danielballan:
FYI, I updated databroker and bluesky to use this `-post` versioning scheme after verbal confirmation from Tom that he definitely prefers that and won't change his mind even though he cannot recall his reasoning. (This was before his honeymoon, don't worry.)

PR #391, opened 2017-03-30
BLD/CI: List requirements descriptively; test pip-based build.

danielballan:
* Add tests on 3.6; drop tests on 3.4 (superceding #386)
* Add a test using pip install and EPICS from Debian (no conda).
* Use requirements.txt and test-requirements.txt.

This applies the same Travis cleanup @tacaswell and I applied to bluesky in https://github.com/NSLS-II/bluesky/pull/660 and https://github.com/NSLS-II/bluesky/pull/668

### DIFF
--- a/install_epics_on_travis.sh
+++ b/install_epics_on_travis.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+set -e
+sudo apt-get update
+sudo apt-get install -yq wget
+wget --quiet http://epics.nsls2.bnl.gov/debian/repo-key.pub -O - | apt-key add -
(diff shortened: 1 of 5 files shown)


danielballan:
Wow I did not expect that work in just three tries.

install_epics_on_travis.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -e
+sudo apt-get update
+sudo apt-get install -yq wget
+wget --quiet http://epics.nsls2.bnl.gov/debian/repo-key.pub -O - | apt-key add -
+echo "deb http://epics.nsls2.bnl.gov/debian/ jessie/staging main contrib" | tee /etc/apt/sources.list.d/nsls2.list
```
arkilic: @danielballan here

PR #392, opened 2017-03-30
ENH: Avoid poll_time=0 bug by setting MIN_POLL_TIME.

danielballan:
attn @klauer, following up on #387

### DIFF
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -20,6 +20,8 @@ import epics
 
 from .errors import DisconnectedError, OpException
 
+MIN_POLL_TIME = 0.0001
+
 __all__ = ['split_record_field',
            'strip_field',
            'record_field',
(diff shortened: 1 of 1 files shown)


ophyd/utils/epics_pvs.py
```diff
@@ -20,6 +20,8 @@
 
 from .errors import DisconnectedError, OpException
 
+MIN_POLL_TIME = 0.0001
```
danielballan: This might be too generous: Our default poll_time, 0.01, is probably a more reasonable lower bound. But given that we back off logarithmically, there's probably no harm is putting a lower value as the hard-coded lower bound.

tacaswell:
What is wrong with a poll time of 0?  Just means we hammer `get`

tacaswell:
Closing as it seems like a feature that the user can force a 0 wait loop.

PR #396, opened 2017-04-03
DOC: Keep extending prose docs

tacaswell:


danielballan:
Latest commit broke the docs build:

https://travis-ci.org/NSLS-II/ophyd/jobs/219278528#L1099

doc/source/device-overview.rst
```diff
@@ -34,15 +36,52 @@ Under the hood, `Device` uses a metaclass to allow for
...
+       yield from bp.mv(robot.execute_cmd, 1)
+
+   def unload_sample(robot):
+       yield from bp.mv(robot.unload_cmd, 1)
+       yield from bp.mv(robot.execute_cmd, 1)
+
+   def robot_plan(list_of_samples):
+       for sample in list_of_samples:
+           # load the sample
+	   yield from load_sample(my_robot, sample)
```
danielballan: The robot is actually probably not a good first example of a device. This skips a couple subtle steps (define `load_sample` plan, register `load_sample` message with RunEngine).

danielballan:
@tacaswell is still working on additional docs, but this is a coherent step forward so we'll publish this as is for now.

PR #403, opened 2017-07-21
MNT: Use the int value at EnableCallbacks instead of the string value

hhslepicka:
This avoid issues when the String representation was changed on the NDPluginBase.template file at areaDetector. The int value never changes.
The current implementation works fine but this change makes it stronger to survive in those cases as mentioned above.
Here at SLAC we faced that and the merge of this PR would avoid the annoyance of creating a class that inherits from PluginBase just to redefine stage.

Att. @teddyrendahl

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -102,7 +102,7 @@ class PluginBase(ADBase):
 
     def stage(self):
         # Ensure the plugin is enabled. We do not disable it on unstage.
-        set_and_wait(self.enable, 'Enable')
+        set_and_wait(self.enable, 1)
         super().stage()
 
     @property
--- a/tests/test_quadem.py
+++ b/tests/test_quadem.py
@@ -44,39 +44,20 @@ def quadem():
             means that set_and_wait() will never be successful for
             EpicsSignalWithRBVs... :-(
     '''
+
     for sig in em.stage_sigs:
         sig = getattr(em, sig)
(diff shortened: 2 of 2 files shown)


hhslepicka:
Weird issue with the test_quadem -> test_scan_point. Tests are passing for area detector.

danielballan:
Sensible enough. I can't help dig into those test failures at this moment but in general I think this is a good change.

tacaswell:
I'm a bit torn, the string version is clearer to readers.  Does upstream thing the int values are also API they need to maintain?  It would be a shame to change to a less readable version _and_ have upstream break us anyway.

danielballan:
I almost suggesting putting 'Enable' in a code comment, but if there are multiple strings that all mean `1` than the comment could accidentally mislead someone into thinking that the string is definitely `'Enable'`.

danielballan:
+1 on your question about whether integers are API stable

hhslepicka:
Integers are API stable as the String can be customized as much as you want!
Please note that I am saying that one can which is different than one
should. ;)

tacaswell:
The failure here is caused by:
 - enable in defined at the ophyd as a string PV (so we got the human-readable version of the enums)
 - the quad_em tests use a the fake_pv classes (as we do not have a 'real' IOC to back the test)
 - 1 is getting string-ified to `'1'` and `set_and_wait` is failing

hhslepicka:
Awesome! Thank you guys for getting it through!

PR #404, opened 2017-07-26
ENH: Move fake_epics into package for external use

teddyrendahl:
We would like to use the `using_fake_epics_pv` wrapper for some external packages containing Ophyd devices. In the current state it is difficult to access these because the tests directory is not inside the package itself.

In the long term it might be nice to just put the tests inside the `ophyd` module much like Bluesky is setup. I figured that for now this would be the less invasive change.


tacaswell:
50/50 to exposing this out to the world.  I would prefer to move away from this in the ophyd test suite as well (and move to using caproto to stand up test servers), but can see why this is useful.

teddyrendahl:
Yeah I've experimented with using pcaspy to simulate the IOC we are testing but you end up dealing with strangeness in both pcaspy and pyepics. It is also a lot of leg work to mock all of the records you need. This seemed like a solid way to smoke test devices, but I could see why it isn't quite polished enough to be part of the main repo.

Any thoughts on sticking the test directory inside ophyd so I can just use:

```python 
from ophyd.tests import using_fake_epics_pv
```
This way it also matches the bluesky structure

tacaswell:
:+1: to moving the tests in, not sure why they are outside here actually...

teddyrendahl:
Closing this PR as #406 voids the need for this change

PR #406, opened 2017-07-30
TST: Internal tests

teddyrendahl:
Longer term solution discussed in #404 . Only sticking point was since `test_signal.py` has `from . import main`, importing `using_fake_epics_pv`  in `__init__ `would create a circular import. My quick solution was to create a `conftest.py` and place the FakeEpicsPV utilities there. This seemed reasonable as they are used in a number of other tests, but if someone has a nicer solution I'm in no means attached.


tacaswell:
Can you investigate why one of the tests (https://codecov.io/gh/NSLS-II/ophyd/src/b185c584d52643fbc72998391fa25098b5a3a859/ophyd/tests/test_flyers.py#L52 ) does not think it is being run?

teddyrendahl:
Okay. Please correct me if I'm mistaken, but it looks like this is skipping the exact tests as master? Looking at the last build log on Travis `test_flyers::test_waveform` and `::test_ad_time_series` were skipped:

https://travis-ci.org/NSLS-II/ophyd/jobs/220559047 

I'm guessing the reason you noticed is that now that I moved the tests directory inside the package it is being included in the coverage report. It looks like these tests have been omitted from the coverage for a while now.

The fixture that is causing the tests to skip is doing so because of a try, except statement watching for `TimeoutError`. If I had to guess there is a mismatch between the IOC created for the tests and the prefix name used in the fixture `'XF:23ID1-ES{Tst-Cam:1}'`. Not entirely confident I understand how you are launching the IOC, but if it is based on the `tests/config.py` file there is no mention of this PV anywhere.

tacaswell:
Oh, because those are aimed at PVs on one of then NSLS-II beamlines.... ( @stuwilkins ).

PR #407, opened 2017-07-31
MNT: work around pyepics returning None

tacaswell:
Just keep trying until it works


hhslepicka:
That is really weird! 😨

tacaswell:
This was more of a way to communicate a patch to @stuartcampbell than a legit PR.

But yes, we had a scaler decide to randomly time out (on a random channel) today which results in `None` getting into the event stream which is all manner of bad.

PR #415, opened 2017-08-03
FIX: ADBase stage should only validate_asyn_ports after staging the signals

hhslepicka:
This allows one to change for instance the nd_array_port in a plugin and have a correct validation on the asyn ports.

Here is the use case for that...

At the normal operation the camera is pointing to a stream of data that is limited in terms of data rate and sometimes with binning
when we want to collect data we want to switch the plugin nd_array_port to the CAM stream which is the raw data coming as fast as the camera can acquire.

Attn @danielballan @tacaswell

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -152,8 +152,13 @@ class ADBase(Device):
                 match_fcn(attr=attr, signal=getattr(self, attr), doc=doc)
 
     def stage(self, *args, **kwargs):
-        self.validate_asyn_ports()
-        return super().stage(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


hhslepicka:
The only thing that I haven't tested is the case of the exception being raised inside Bluesky and what will happen when Bluesky calls unstage again after the exception on validate_asyn_ports.

Any clues?

danielballan:
Just to document this for anyone following along, I'll reiterate what we discussed on Hangouts. The scenario you mentioned is not a problem because (foreseeing this exact problem) unstaging is an indempotent operation: redundant calls have no effect. This is not true of *staging*.

Thanks for the fix!

PR #417, opened 2017-08-06
WIP: Implement commands as magics.

danielballan:
Needs tests and also %log_pos

### DIFF
--- a/ophyd/magics.py
+++ b/ophyd/magics.py
@@ -1,0 +1,35 @@
+from IPython.core.magic import Magics, magics_class, line_magic
+from operator import attrgetter
+from ophyd.utils import DisconnectedError
+
+@magics_class
+class SPECMagics(Magics):
+
+    positioners = []
+
+    @line_magic
+    def wa(self, line):
+        "List positioner info. 'wa' stands for 'where all'."
+        if line.split():
+            raise TypeError("No parameters expected, just %wa")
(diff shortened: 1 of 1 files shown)


ophyd/magics.py
```diff
@@ -0,0 +1,38 @@
+from IPython.core.magic import Magics, magics_class, line_magic
+from operator import attrgetter
+
+@magics_class
+class SPECMagics(Magics):
+
+    positioners = []
+
+    @line_magic
+    def wa(self, line):
```
stuartcampbell: Is it worth adding something saying what `wa` does for the uninitiated? 😄

danielballan: Good suggestion. Done and force-pushed.

danielballan:
I found a nice way to implement `%wa` without an ophyd dependency, so I'm putting this in `bluesky.magics` for now (in https://github.com/NSLS-II/bluesky/pull/720). Separately we can discuss whether we want to pull that out as a separate repo, but it does seem nice to have all the magics in one module -- wherever it lives.

danielballan:
Specifically, it's here: https://github.com/NSLS-II/bluesky/pull/720/commits/3cab61fa830d6e7d3c8ee1c041f010aafd43d2d0#diff-9cbbeaee4ee19a2f0f7fd6cae424b824

PR #420, opened 2017-08-14
ENH: Add hooks useful for progress bars.

### LINKED ISSUE #419: Add API to status objects to support smarter progress bars

danielballan:
Closes #419. ~~Needs tests.~~ Tested.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -206,6 +206,7 @@ class DeviceStatus(StatusBase):
     '''
     def __init__(self, device, **kwargs):
         self.device = device
+        self._watchers = []
         super().__init__(**kwargs)
 
     def _handle_failure(self):
@@ -218,6 +219,17 @@ class DeviceStatus(StatusBase):
                 'success={1.success})'
                 ''.format(self.__class__.__name__, self)
                 )
+
+    def watch(self, func):
+        # See MoveStatus.watch for a richer implementation and more info.
+        if self.device is not None:
+            self._watchers.append(func)
+            func(name=self.device.name)
+
+    def _settled(self):
+        '''Hook for when status has completed and settled'''
+        for watcher in self._watchers:
+            watcher(name=self.device.name, fraction=1)
 
     __repr__ = __str__
 
@@ -272,7 +284,6 @@ class MoveStatus(DeviceStatus):
         self.finish_ts = None
         self.finish_pos = None
 
-        self._watchers = []
         self._unit = getattr(self.pos, 'egu', None)
         self._precision = getattr(self.pos, 'precision', None)
         self._name = self.pos.name
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -321,7 +321,7 @@ class MoveStatus(DeviceStatus):
         # *args and **kwargs catch extra inputs from pyepics, not needed here
         if not self._watchers:
             return
-        current = self.pos.position
+        current = value
         target = self.target
         initial = self.start_pos
         time_elapsed = time.time() - self.start_ts
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -318,7 +318,7 @@ class MoveStatus(DeviceStatus):
         self._watchers.append(func)
 
     def _notify_watchers(self, value, *args, **kwargs):
-        # *args and **kwargs catch extra junk from pyepics, not used here
+        # *args and **kwargs catch extra inputs from pyepics, not needed here
         if not self._watchers:
             return
         current = self.pos.position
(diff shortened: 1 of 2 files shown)


ophyd/status.py
```diff
@@ -264,12 +264,35 @@ def __init__(self, positioner, target, *, start_ts=None,
...
+            Expected signature: ``func(status_obj, fraction_done)``
+        """
+        self._watchers.append(func)
+
+    def notify_watchers(self):
+        try:
+            fraction = ((self.pos.position - self.start_pos) /
+                        (self.target - self.start_pos))
+        except Exception as exc:
+            fraction = exc
```
tacaswell: This seems odd.  Why would you want to pass the exception into the watcher?

danielballan: I am concerned that the subtraction operation above might fail in some situations, and I don't want that to kill a run in progress -- at least until this feature has seen battle-tested on the floor. I thought it might sense to report the exception back to the watcher. In JavaScript callback conventionally have the signature `f(result, err)`, a pattern I thought about adopting here. But in Python we can check `isinstance(res, Exception)` and be on our way.

tacaswell: It makes sense, just surprising.  Maybe just print the traceback and don't call the callback?

danielballan:
Previously, status objects could launch a `_timeout_thread`, which polled at some appropriate interval to check whether the timeout had been exceeded. If the timeout was `None`, no thread was launched.

In this PR, status objects always launch a thread (rechristened the `_heartbeat_thread`) which polls a fixed interval (0.2 seconds) to notify any progress bars. It also checks the timeout, if applicable.

```python
In [3]: m = EpicsMotor('XF:31IDA-OP{Tbl-Ax:X1}Mtr', name='m')

In [4]: s = m.set(5); s.watch(print)

In [5]: 0.04499999999999993
0.14500000000000002
0.2450000000000001
0.3500000000000001
0.5
0.6000000000000001
0.7000000000000002
0.8050000000000002
0.9050000000000002
0.9900000000000002
1.0
1.0
1.0
1.0
```

tacaswell:
:disappointed: for threads.  I would rather subscribe to the proper thing (which is typically the default callback).  That way we get updates on the cadence of the underlying hardware.

danielballan:
Good point, not sure why I didn't think of that first. New commit does it with `subscribe`. Notice we don't get the string of `1.0`s the end:

```python
In [5]: s = m.set(1); s.watch(print)

In [6]: 0.02
0.09
0.19
0.29
0.39
0.5
0.6
0.7000000000000001
0.8
0.9
0.97
1.0
In [6]:
```

ophyd/status.py
```diff
@@ -57,15 +57,19 @@ def __init__(self, *, timeout=None, settle_time=None, done=False,
...
 
         if self.timeout is not None and self.timeout > 0.0:
-            thread = threading.Thread(target=self._timeout_thread, daemon=True)
+            thread = threading.Thread(target=self._wait_and_cleanup,
+                                      daemon=True)
             self._timeout_thread = thread
             self._timeout_thread.start()
 
-    def _timeout_thread(self):
+    def _wait_and_cleanup(self):
```
danielballan: Note: The original name of this method collided with the name of the attribute where the thread was stash. It doesn't matter because the method is only called once, and it might even have been intentional, butI found it confuses so I'm leaving the change in.

ophyd/status.py
```diff
@@ -264,12 +269,47 @@ def __init__(self, positioner, target, *, start_ts=None,
         self.pos = positioner
         self.target = target
         self.start_ts = start_ts
+        self.start_pos = self.pos.position
         self.finish_ts = None
         self.finish_pos = None
 
+        self._watchers = []
+
+        def _notify_watchers(value, *args, **kwargs):
```
tacaswell: you need to hold on to a ref to this function so you can un-subscribe it later

ophyd/status.py
```diff
@@ -264,12 +269,47 @@ def __init__(self, positioner, target, *, start_ts=None,
...
+
+        Parameters
+        ----------
+        func : callable
+            Expected signature: ``func(fraction_done)``
+        """
+        self._watchers.append(func)
+
+    def _notify_watchers(self, value):
+        if not self._watchers:
```
tacaswell: I would just like this take `*args, **kwargs`, subscribe this method and have logic like

```python
if self.done:
    self.pos.clear_sub(self._notify_watchers)
    return 
```

danielballan: Grumble grumble `*args **kwargs` :-P

PR #426, opened 2017-08-18
ENH: default read and configuration attrs

tacaswell:


danielballan:
Tests failures are random failures on unrelated code (homing forward, homing reverse) that we often see. Merging!

PR #438, opened 2017-08-31
FIX: Defer more computation to the progress bar.

danielballan:


danielballan:
The only reason for the status object to be computing fraction, time_remaining, etc. is if it knows about some special case that the progress bar can't know. Otherwise we should just let the progress bar do its naive computation.

PR #442, opened 2017-09-01
ENH: SubscriptionStatus

teddyrendahl:
Created a `SubscriptionStatus` object that uses the built-in Ophyd event system to update itself
## Description
The `SubscriptionStatus` is a subclass of the `DeviceStatus` object that accepts a device, event_type and callback. The callback is registered at initialization and automatically cleared from the device after the status finishes via completion or timeout.
 
## Motivation and Context
While `PositionerBase` handles any device that has a `DMOV` type flag to subscribe to, I found myself rewriting a lot of code for simpler devices that did not have this flag. For instance if you wanted to set signal A, and wait for signal B and C to reach specific values you were stuck implementing this on a class by class basis. Now you can just write a simple function to check the state of these PVs and summarize it in a boolean, then plug it straight into the `SubscriptionStatus`. I also like that this solution avoids unnecessary polling and spawning additional threads.

## How Has This Been Tested?
I wrote a basic test inside `ophyd/tests/test_status.py` but saw afterwards that `ophyd/tests/test_ophydobj.py` has most of the actual `Status` tests. Let me know if you want it moved. I also didn't put a test in where this status fails, tried to avoid long waits and/or race conditions.

## General Thoughts
There are only two ways the status will mark itself finished, either via timeout or a success. I could see this causing some strange behavior, but I don't think there is a clean way around it if the device does not have a DMOV flag you can use. For example, if you create your status with a timeout of `None`, then check it later, you have no idea if it is your request that finished it or a subsequent call or motion of the device triggered the completion. I considered putting in a default `timeout` but decided against it.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -234,6 +234,70 @@ class DeviceStatus(StatusBase):
     __repr__ = __str__
 
 
+class SubscriptionStatus(DeviceStatus):
+    """
+    Status updated via `ophyd` events
+
+    Parameters
+    ----------
+    device : obj
+
+    callback : callable
+        Callback that takes event information and returns a boolean. Signature
+        should be `f(*args, **kwargs)`
+
+    event_type : str, optional
+        Name of event type to check whether the device has finished succesfully
+
+    timeout : float, optional
+        Maximum timeout to wait to mark the request as a failure
+
+    settle_time : float, optional
+        Time to wait after completion until running callbacks
+    """
+    def __init__(self, device, callback, event_type=None,
+                 timeout=None, settle_time=None):
+        #Store device and attribute information
+        self.device    = device
+        self.callback  = callback
+
+        #Start timeout thread in the background
+        super().__init__(device, timeout=timeout, settle_time=settle_time)
+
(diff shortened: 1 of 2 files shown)


tacaswell:
I was just discussing the need for something like this with @Brudhu this week!

Brudhu:
Yes, very nice!

ophyd/status.py
```diff
@@ -234,6 +234,70 @@ def _settled(self):
...
+    timeout : float, optional
+        Maximum timeout to wait to mark the request as a failure
+
+    settle_time : float, optional
+        Time to wait after completion until running callbacks
+    """
+    def __init__(self, device, callback, event_type=None,
+                 timeout=None, settle_time=None):
+        #Store device and attribute information
+        self.device    = device
```
tacaswell: please follow pep8, only one space around `=`

tacaswell:
We really need to start running a linter on travis.

danielballan:
I'm not sure why this got hung up. Apologies, @teddyrendahl.

(The one build failure is the docs thing addressed by #451.)

PR #443, opened 2017-09-06
CI: Update EPICS installation to account for thrashed packaging.

danielballan:


danielballan:
Closing. We should adopt the approach we use in caproto here.

Issue #179, opened 2015-12-15
Sundry cleanup


danielballan:
* [x] Make sure `_sep` is scrubbed out from the codebase.
* [x] Check that `range_def` in device.py is removed.
* [x] Add full docstring for DDC.
* [ ] Consider which fields would be better as configuration fields (e.g., `motor_egu`)
* [x] Check that trigger doesn't accept kwargs.
* [x] Check that the `configuration` property is removed.
* [x] Make sure that `describe()` and `describe_config()` emit all the data key fields we need.

PR #444, opened 2017-09-12
API: Stop acquiring when device is stopped (via pause for example).

danielballan:
The addition of progress bars has made it more obvious that detectors
continue acquiring after the RunEngine is paused. I 'fixed' that
locally at CSX but I *think* it's a sensible default and should move
into ophyd here.

Open for debate!

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -143,6 +143,10 @@ class SingleTrigger(TriggerBase):
             # Negative-going edge means an acquisition just finished.
             self._status._finished()
 
+    def stop(self):
+        self._acquisition_signal.put(0)
+
+
 
 class MultiTrigger(TriggerBase):
     """This trigger mixin class can take multiple acquisitions per trigger.


hhslepicka:
This is a tricky one and I have mixed feelings about it... Sometimes if you put the acquire to 0 you can totally screw the rest of the collection for the detector... some detectors will need a brand new stage to work properly.
In which case you saw it? How was the detector configured?

danielballan:
The report that motivated this was #179 -- it was the Fast CCD at CSX with standard AD configuration. Staff can always customize what happens by defining custom `stop`, `pause` and `resume` methods on their devices. Are there cases where it wouldn't make sense to _at least_ stop acquisition (and maybe do other things as well)? If so, then we shouldn't put this here because it's easier to add functionality in a subclass that to remove it.

hhslepicka:
It also depends on another question... if you do stop what do you want from the current exposure? Some detectors will abort and drop the image/value. Is this the desired behavior?

Standard AD configuration doesn't say much... was it Free Run? Was it External Triggering? Fixed number of images using the internal clock?

You are right, we can always customize the behavior at our subclasses but what I am not sure is if this is the correct default one to have and I am trying to look for cases where it can affect so we can stimulate the discussion on wether or not the default should be to stop. For a ```abort``` case I totally agree that it should stop, for a pause I am not sure... maybe the detector in some cases should be able to just raise an exception and say: "Sorry, I can't pause... the data collection is ruined." if it can't restart.

danielballan:
This does get thorny. My gut feeling was that is was safer to customize on a per-beamline, per-detector basis, and it seems that is the case. This would be worth revisiting if the ophyd side grew more semantics around various AD configurations and refined interruption behavior (like sub-trigger checkpoints).

I think this should not be merged, but I'll leave the issue open for now just to remind us of these ideas.

hhslepicka:
Pinging @klauer @dchabot and @prjemian for more input...

prjemian:
Can this be made an option rather than an absolute decision?

danielballan:
Yes. Currently, it's an option. The way to invoke that option is by extending the `stop()` method via subclassing. It sounds like moving this into the base class opens up a can of worms at best and creates hard-to-unravel problems at worst.

hhslepicka:
@dchabot If I remember it correctly Dectris Pilatus will drop the image acquisition if you write 0 to acquire while it is exposing... it was a while ago... I believe that most of the CCDs will do the same...
@brunoseivam what happens with the Eiger detector if you write 0 while it is exposing? Will it generate the output of the last frame and stop or just drop the procedure and the data?

brunoseivam:
The Eiger driver currently sends an 'abort' to the server if 0 is written to Acquire. This causes the current frame being exposed to be dropped. There's a plan to change from 'abort' to 'cancel', which waits for the current frame to finish before stopping.

Neither 'abort' or 'cancel' will drop *all* data, just potentially the last frame being exposed.

There's no 'pause'. If stopped, the Eiger has to be rearmed.

tacaswell:
I agree with @dchabot , pausing a plan should stop everything involved.

On pause we currently call the `pause` method on each objects which gives them a chance to make them selves safe and to raise a special exception if they can not be replayed.  I suspect we should extend this to allow the object to return a plan that needs to be run in order to resume (is the case where we can stuff all of the logic into the device) and a second Exception class that will be thrown down to the plans to notify them that _they_ must take extra care to clean up.

Alternatively, it might make more sense for the `resume` method to return the plan required to resume the device.

danielballan:
I also agree that pausing a plan should stop everything involved, but if @hhslepicka is right that stopping it safely (without needlessly dumping data) is sometimes device-specific, then `SingleTrigger` is the wrong place for this logic.

danielballan:
?

Get Outlook for iOS<https://aka.ms/o0ukef>



On Thu, Apr 19, 2018 at 8:21 PM -0400, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


Closed #444<https://github.com/NSLS-II/ophyd/pull/444>.

tacaswell:
This just sets the acquisition bit to 0, to stop and resume an area detector in general requires something much more complex.

PR #446, opened 2017-09-27
ENH/API: Make status objects support __and__

danielballan:
I'm reviving this very old branch because I keep seeing places
where it seems useful to me, including most recently a use case
that @swilkins found.

Unlike my first attempt, this one is fully backward-compatible.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -1,6 +1,8 @@
+from collections import deque
 import time
 from threading import RLock
 from functools import wraps
+from warnings import warn
 
 import logging
 import threading
@@ -9,13 +11,17 @@ import numpy as np
 logger = logging.getLogger(__name__)
 
 
+class UseNewProperty(RuntimeError):
+    ...
+
+
 # This is used below by StatusBase.
 def _locked(func):
     "an decorator for running a method with the instance's lock"
     @wraps(func)
     def f(self, *args, **kwargs):
         with self._lock:
-            func(self, *args, **kwargs)
+            return func(self, *args, **kwargs)
 
     return f
 
@@ -38,7 +44,7 @@ class StatusBase:
                  success=False):
         super().__init__()
         self._lock = RLock()
-        self._cb = None
+        self._callbacks = deque()
         self.done = done
         self.success = success
         self.timeout = None
@@ -98,9 +104,9 @@ class StatusBase:
             self.done = True
             self._settled()
 
-            if self._cb is not None:
-                self._cb()
-                self._cb = None
+            for cb in self._callbacks:
+                cb()
+            self._callbacks.clear()
 
     def _finished(self, success=True, **kwargs):
         '''Inform the status object that it is done and if it succeeded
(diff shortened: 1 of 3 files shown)


danielballan:
To spell out the consequences a little more:

Currently, status objects can only be set with _one_ callback that they will call when they are complete:

```python
status.finished_cb = cb
status.finished_cb is cb
```

If we want to make it possible to do something like `status1 & status2`, then status objects need to support multiple callbacks, like this:

```python
status.add_callback(cb1)
status.add_callback(cb2)
status.callbacks == set(cb1, cb2)
```

 because of this scenario:

```python
a = status1 & status2
b = status1 & status3
# status1 has to notify both a and b when it is done -- hence needs to support multiple callbacks.
```

We can support the legacy API (illustrated in the first code block) in the special case where there is one callback attached. See `ophyd/tests/test_status.py::test_status_legacy_finished_cb` for a demonstration of this.

Thoughts on this extension of the status API -- @teddyrendahl, @ZLLentz, @hhslepicka, @prjemian?

ophyd/status.py
```diff
@@ -98,9 +103,9 @@ def _settle_then_run_callbacks(self, success=True):
             self.done = True
             self._settled()
 
-            if self._cb is not None:
-                self._cb()
-                self._cb = None
+            for cb in list(self._callbacks):
```
tacaswell: This is locked so you can just loop through them and then clear the callbacks I think.

ophyd/status.py
```diff
@@ -38,7 +43,7 @@ def __init__(self, *, timeout=None, settle_time=None, done=False,
                  success=False):
         super().__init__()
         self._lock = RLock()
-        self._cb = None
+        self._callbacks = set()
```
tacaswell: Probably better to use a dequeue or list so we can make guarantees about order the callbacks are run in.  I am not sure I _like_ it being used that way, but cpython makes these sort of garuntees on `atexit` and `weakref` callbacks.

also I am channeling @klauer

prjemian:
My initial take was "sounds good".

On second look, the handling of response status with possible conditional callbacks (do cb2 if cb1 succeeds but otherwise cb3) makes my head spin.  Is there a simple way to express this without inventing a language?

tacaswell:
I am in principle in favor of this, but left some implementation comments

tacaswell:
@prjemian I do not understand your comment.

prjemian:
In https://github.com/NSLS-II/ophyd/pull/446#issuecomment-332827465, there was the example

```
a = status1 & status2
b = status1 & status3
```

which led me to believe there might be some conditional situation involving callbacks.  Maybe the status is just a simple `set(cb1, cb2)`?

tacaswell:
These objects are what are used to communicate to the RunEngine that various asynchronous tasks are done.  In almost all cases the only callback that is installed onto them is the `asyncio.Event.set` method to act as a bridge between an epics monitor firing to releasing the RunEngine. 

We have also talked about making status objects for complex/python pseudo devices smarter, for example the object returned by a energy axis, which has an undulator and a mono, could effectively expose 3 status objects, one for the total motion being done, the mono being done and the undulator being done

```python
class E(Device):
    mono = Cpt(Mono)
    gap = Cpt(Gap)

    def set(self, E):
        st_gap = self.gap.set(E_to_gap(E))
        st_mono = self.mono.set(E_to_bragg(E))
        st_E = st_gap & st_mono
        st_E.children = {'mono': st_mono, 'gap': st_gap}
        return st_E
```

We can then add a Msg like `Msg('wait_for_status', None, status)` so a plan could now wait for all three parts independently (for example if you don't care if gap is done, only that the mono is done).  To support this sort of thing we need to be able to support multiple callbacks as implemented here.

I think an alternate approach to this would be to put chaining logic in the setter

```python

if self._cb is not None:
    old_cb = self._cb
    def inner():
        old_cb()
        cb()
    self._cb = inner
else:
    self._cb = cb

```

but maintaining the list of callbacks via a deque / list seems easier to understand than maintaining that state via nested closures.

danielballan:
Let me try to answer @prjemian's comment another way:

> On second look, the handling of response status with possible conditional callbacks (do cb2 if cb1 succeeds but otherwise cb3) makes my head spin. Is there a simple way to express this without inventing a language?

Nobody here likes logic expressed in callback chains -- for sure! That's not what this is accomplishing. This is letting us express logic like, "Wait until three components have all finished triggering/moving." There is no "if/else" logic here. There is only "wait for all", like this:

```python
combined_status = a_status & b_status & c_status
```

We already had a way to express "wait for all" in a _plan_:

```python
import bluesky.plans as bp

def plan():
    yield from bp.trigger(device.a, group='foo')  # Each of these hands the RunEngine a status object.
    yield from bp.trigger(device.b, group='foo')
    yield from bp.trigger(device.c, group='foo')
    yield from bp.wait('foo')  # Tell RunEngine to wait for all status objects to complete.
```

This PR gives us the option of encapsulating that logic in a device, which may lend itself to better reusability in some circumstances -- circumstances where we don't want to leak device-specific concerns into the plan.

```python
class Device:
    ... # define components a, b, c

    def trigger(self):
        return self.a.trigger() & self.b.trigger() & self.c.trigger()

device = Device(...)

def plan():
   yield from bp.trigger(device, wait=True)
```

ophyd/status.py
```diff
@@ -77,11 +77,15 @@ def _wait_and_cleanup(self):
...
-            logger.debug('Status object %s timed out', str(self))
-            try:
-                self._handle_failure()
-            finally:
-                self._finished(success=False)
+            with self._lock:
+                if self.done:
+                    # Avoid race condition with settling.
+                    return
+                logger.debug('Status object %s timed out', str(self))
```
tacaswell: Does all of this need to be in the lock?

tacaswell:
Do we need a lock around `self.done` in `_finished` as well?

ophyd/status.py
danielballan: Yes. We can't have settling start running while failure is being handled.

tacaswell: right, because we set `done` pretty late

danielballan: Exactly

danielballan:
Where?

danielballan:
I think we do not because we are not setting it there.

prjemian:
re https://github.com/NSLS-II/ophyd/pull/446#issuecomment-337259759: ok with that clarification, my head stops spinning

danielballan:
@prjemian Great! Thanks for making us pause to explain ourselves. It's useful.

tacaswell:
I have a suspicion that there are still some thread-related edge cases in here, but they were here _before_ this change.

PR #451, opened 2017-10-18
CI: Turn off parallel building for now because it is broken.

danielballan:
There seems to be an issue with sphinx's parallelization feature
that is not specific to our projects. This PR removes the `-j` option
to get things working again until it is resolved.


danielballan:
The failure is random and not related. Self-mering this because it only touches `doc/Makefile` and it's blocking everything else.

PR #452, opened 2017-10-18
Mnt callbacks

tacaswell:


tacaswell:
Why did circle come back?

klauer:
I'll assume this adds some actual benefit over the original code other than making it Tomish. Enjoy the merge! 👍 

(inside joke - don't mind me)

ophyd/ophydobj.py
```diff
@@ -31,21 +45,28 @@ class OphydObject:
...
-        self._sub_cache = defaultdict(lambda: None)
+        self.subscriptions = {getattr(self, k)
+                              for k in dir(type(self))
+                              if (k.startswith('SUB') or
+                                  k.startswith('_SUB'))}
+
+        # dictionary of wrapped callbacks
+        self._callbacks = {k: {} for k in self.subscriptions}
+        # this is to maintain api on clear_sub
+        self._unwrapperd_callbacks = {k: {} for k in self.subscriptions}
```
danielballan: I had to pick on private variable names, but can we do "unwrapped" here instead of "unwrapperd"?

danielballan:
I believe @stuartcampbell is adding Circle-CI in some repos. The barrier to switching is getting lower as our .travis.yml files have become much simpler. I'm working on doctr support for Circle.

ophyd/ophydobj.py
tacaswell: I no longer remember if that is a typo that editor helpful complete propogated or if it was intentional.

tacaswell:
We may not be testing many of those places...

tacaswell:
Do we want to walk back that warning on using the default subscription?

danielballan:
We probably shouldn't risk adding a bunch of internally-tripped warnings in 1.0. Let's consider adding the warning in a future, less rushed release when we have time to catch the warnings on the floor.

danielballan:
Some context for any remote collaborators following along: this should make the internals of OphydObj, particularly the subscription system, easier to follow. It should not break any API.

PR #455, opened 2017-10-19
API: make the name keyword to device a required keyword only arg

tacaswell:
This ensures that the names that appear in the read dictionary are
always human-readable.

In practice if this is left out it is accidental and results in saved
data with keys that are a concatenation of the PV prefix with the
names of the children in the python class.  This is neither human
understandable nor a valid / existing PV.  Neither humans nor machines
can read this.

Keeping keyword only to preserve the ability to have sub-classes with
more than one required positional arguments.

Do not want to make it the first positional argument to avoid massive
API breakage and the convention of 'last positional argument' is
brittle.

A keyword-only argument also clearly disambiguates the name (for
humans) from the prefix (for hardware).

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -676,13 +676,13 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     ----------
     prefix : str
         The PV prefix for all components of the device
+    name : str, keyword only
+        The name of the device
     read_attrs : sequence of attribute names
         the components to include in a normal reading (i.e., in ``read()``)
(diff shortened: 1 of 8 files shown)


tacaswell:
attn @prjemian

prjemian:
+1 
Did not realize this construction (`, *, name, `) was possible:

    def __init__(self, prefix, *, name,
                 read_attrs=None, configuration_attrs=None,
                 parent=None, **kwargs):

tacaswell:
That is a new feature in python3 and one of the ones I am most excited about :smile:

PR #457, opened 2017-10-19
API: Make Component lazy by default.

danielballan:
Closes #443. It looks like our friends at SLAC are
[already doing this locally](https://github.com/slaclab/pcds-devices/blob/780117057595b906f813e171f2efd676d48eb1e8/pcdsdevices/component.py#L10).

I originally reported (in #443) that I thought this might cause
trouble but I was seeing an unrelated issue. I think this is pretty
safe to change, and it should improve IPython load times in a big way.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -71,7 +71,7 @@ class Component:
         string to attach to component DvcClass.component.__doc__
     '''
 
-    def __init__(self, cls, suffix=None, *, lazy=False, trigger_value=None,
+    def __init__(self, cls, suffix=None, *, lazy=True, trigger_value=None,
                  add_prefix=None, doc=None, **kwargs):
         self.attr = None  # attr is set later by the device when known
         self.cls = cls


danielballan:
attn @prjemian

teddyrendahl:
So just a heads up. This does decrease our load times substantially, but it means that it is pretty easy to initialize a device and not realize it is broken before you use it. We usually try and call `.wait_for_connection` before we trust the device. Maybe this is even something we should consider baking into `stage` if everything is lazy?

The most obnoxious side affect is if you press tab in an IPython shell the misconfigured device times out on all your signal attributes, locking up the shell ....

klauer:
`lazy=True` means now all pyepics connection requests will be done in a blocking fashion one-by-one, slowing you down at some point later. In my opinion, a slower startup time is more acceptable, hence the current default.

prjemian:
Ok

ZLLentz:
Some questions about `lazy=True`:
https://github.com/NSLS-II/ophyd/blob/7d1347f4d4761de4850924faba4101fdf87f2c91/ophyd/device.py#L132-L133
At this point in the `create_component` routine, we're waiting for the connection on all lazy components. Why is this done? Non-lazy component connections are never waited for at any point. It seems like the differences between lazy and non-lazy are:

- Non-lazy components get their `__init__` statements called at the end of the parent device `__init__`
- Lazy components get their `__init__` statements called at first access
- Lazy components are forced to wait for all connections in series at first access

If I'm interpreting this right, then the startup difference between lazy and non-lazy components is largely just slow calls in the subcomponent `__init__` statements, rather than EPICS slowness, and the tab-completion slowdown Teddy and I see is just the `wait_for_connections` call that lazy components are forced through. Is there a reason why lazy components should be so proactive in waiting for connections before the signals are needed? To me, the keyword `lazy` implies that we start connections but do not wait for them, and then trust our `Device` methods to wait for connections as needed.

klauer:
@ZLLentz you are correct in that the delay for non-lazy signals is not from `wait_for_connection` but slow initialization/calls to ctypes/etc. The assumption made is that you can easily do a massive set of connection requests at startup in parallel and mostly everything will be ready once you hit the IPython prompt.

The idea behind Components was that they should be ready to use on first access. As a user, I don't want to have to know if a component is lazy or not - littering code with `wait_for_connection()` here and there. As `Device` doesn't really know the intention of the caller (will they access one component or all of them?) it does its best effort to ready each component on access.

I didn't see a better way to do it then and I still don't now, but I'm open to suggestions.

danielballan:
As an aside, I am convinced that we should not throw this in for 1.0. I think this I prefer lazy default -- amortizing the connection cost instead of paying it all at startup time -- but I think working `wait_for_connection` into stage and exploring better parallelism of channel creation may need to be done first.

ZLLentz:
My suggestion is to remove the first-access `wait_for_connection()` call from `create_component` and then, in `EpicsSignalBase`, rather than putting a `@raise_if_disconnected` decorator around everything, use a decorator with a `wait_for_connection()` call that will be tried exactly once for each signal. Now you don't have to worry about handling this at the `Device` or `Component` level, instead you trust that the `Signal` subclasses will try to ensure they are ready at access time if they haven't finished establishing their connections for some reason.

danielballan:
I like that suggestion.

klauer:
That seems like a valid alternative. You'll of course have to implement it in a way that connections are done in parallel (otherwise you'll end up with the exact same scenario as exists currently). There's more complexity to this than meets the eye, but you'll find that out soon enough as you move to implement the solution.

klauer:
Did not intend for my comments to make progress on this stagnate! Hopefully it's easier (and with less complications in existing beamlines) than I'm imagining. By all means move forward with it and 🤞 it will work out.

danielballan:
Good to know. I thought you guys were still using lazy-by-default.

As an aside, we have discussed ways of making ophyd devices connect in bulk. We would need some new interface between ophyd and the control layer to say, "Tell me all of your signals," rather than letting signals connect one at a time. Bulk-connection looks to be _significantly_ (~100X) faster, and that would remove some of the benefit of lazy connection in the first place.

teddyrendahl:
Yup no issues since we discovered `setup_ophyd`. 👍 to the idea of bulk connections. We have a few applications that connect to ~100 Ophyd devices, it would be handy to speed startup of these up.

ZLLentz:
I'm always :+1:  for changes that speed up the initialization process, though I was never under the impression that the one-at-a-time connections were the limiting factor here.

danielballan:
Seems like we're all for closing this then. Please reopen if I have misinterpreted the state of the consensus.

PR #462, opened 2017-10-27
ENH: teach PseudoPosition and RealPosition {+, -} and other useful lessons

tacaswell:
It turns out even in python we are writing a vector library.

We may not need this...

### DIFF
--- a/ophyd/pseudopos.py
+++ b/ophyd/pseudopos.py
@@ -354,13 +354,31 @@ class PseudoPositioner(Device, SoftPositioner):
 
         def __add__(self, other):
             if not isinstance(other, type(self)):
-                return NotImplemented
+                try:
+                    for k in self._fields:
+                        other.setdefault(k, 0)
+                    other = type(self)(**other)
+                except (TypeError, AttributeError):
+                    try:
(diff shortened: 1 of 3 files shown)


klauer:
> We may not need this...

As if that's enough to stop us! Looks good to me.

ophyd/pseudopos.py
```diff
@@ -327,6 +331,29 @@ class PseudoPositioner(Device, SoftPositioner):
...
+
+        def __add__(self, other):
+            if not isinstance(other, type(self)):
+                raise NotImplemented
+            return type(self)(*(s + o for s, o in
+                                zip(self, other)))
+
+        def __sub__(self, other):
+            if not isinstance(other, type(self)):
+                raise NotImplemented
```
klauer: Refer to https://docs.python.org/3.6/library/constants.html#NotImplemented

I believe the proper way is to either return `NotImplemented` or raise `NotImplementedError` - they aren't interchangeable.

tacaswell: Got that I wanted `NotImplemented` but not that I wanted to return it :disappointed:

PR #463, opened 2017-10-30
MNT: Do not directly import pyepics anywhere but control_layer

tacaswell:


tacaswell:
```
✔ ~/source/bnl/ophyd [master {NSLS-II/master}|✚ 17…3⚑ 2] 
13:39 $ git grep 'from epics\>'
ophyd/epics_compat.py:from epics import get_pv as _get_pv, caget, PV, caget, caput
✔ ~/source/bnl/ophyd [api_support_multiple_epics_libraries L|…2⚑ 2] 
13:45 $ git grep 'import epics\>'
.travis.yml:  - python -c "import epics; print(epics.__version__)"
.travis.yml:  - python -c "import epics.ca; print(epics.ca.find_libca())"
.travis.yml:  - python -c "import epics; print(epics.caget('XF:31IDA-OP{Tbl-Ax:X1}Mtr'))"
ophyd/epics_compat.py:import epics
ophyd/epics_compat.py:    import epics
ophyd/epics_compat.py:    import epics
✔ ~/source/bnl/ophyd [api_support_multiple_epics_libraries L|…2⚑ 2] 
13:45 $ 

```

ophyd/epics_compat.py
```diff
@@ -0,0 +1,236 @@
...
+                      epics.__version__,
+                      ImportWarning)
+        return 'native'
+    else:
+        return 'time'
+
+
+pv_form = get_pv_form(epics.__version__)
+
+_exposed_names = ['caput', 'caget', 'thread_class', 'get_pv', 'setup', 'PV',
```
danielballan: How is this different from `__all__`?

tacaswell: I have change it to `__all__` in uncommited changes already, this is notes to my self

tacaswell:
@teddyrendahl @ZLLentz @hhslepicka @klauer @prjemian   This is a pretty major re-organization of imports. 

The idea is to push all of the imports into `_LIB_shim.py` which are imported into `epic_compat.py` which is what is actually used by the code.  This should allow ophyd to switch between pyepics, caproto, pyca,etc assuming the library can provide a small handful of functions and a `PV` class that duck-types to the pyepics PV class.

We could probably get the required list down to {`get_pv`, `threading_class`, `setup`} if we tried.

ophyd/epics_compat.py
```diff
@@ -0,0 +1,29 @@
...
+def fmt_time(tstamp=None):
+    "simple formatter for time values"
+    if tstamp is None:
+        tstamp = time.time()
+    tstamp, frac = divmod(tstamp, 1)
+    return "%s.%5.5i" % (time.strftime("%Y-%m-%d %H:%M:%S",
+                                       time.localtime(tstamp)),
+                         round(1.e5*frac))
+
+if os.environ.get('CAPROTO', False):
```
stuartcampbell: Do we want to make the environment variable more descriptive ?  (e.g. `USE_CAPROTO`) or 'namespace' it to ophyd (e.g. `OPHYD_CAPROTO`?  

Just throwing thoughts out!

ophyd/control_layer.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+__all__ = ['setup',
+           'caput', 'caget',
+           'get_pv',
+           'pv_form',
+           'thread_class']
+
+if os.environ.get('CAPROTO', False):
```
danielballan: Re-upping @stuartcampbell's comment: something more specific, like:

```python
if os.environ.get('OPHYD_CONTROL_LAYER').lower() == 'caproto':
```

stuartcampbell: I like it, much better than my suggestions.

tacaswell: and that gives a nice hook to a future extension where `ophyd.setup(...)` gives you a code-level (rather than ENV level) hook to control this.

tacaswell:
@stuartcampbell done

danielballan:
This is exciting! @ZLLentz Any interest in adding a PyCA compat layer?

ZLLentz:
@danielballan I'll write a pyca compat layer at some point, but it's not super high on the list of priorities right now. I'm glad that the option is available in case I find a pressing need to switch. To be fully open, I'm much more excited by the prospects of a future caproto than by the prospects of revisiting pyca.

PR #468, opened 2017-11-01
FIX: add mover

### LINKED ISSUE #467: DOC: Mover still in docstrings
It seems to `Mover` got lost in the move from bluesky to ophyd. It is still referenced in all the examples though. Can we please get a shim for this?

CJ-Wright:
Closes #467


danielballan:
See discussion on #467. This code was intentionally removed for reasons explained there.

PR #476, opened 2017-11-02
Random number with given seed

licode:
also fixed bug in poisson noise.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -331,11 +331,13 @@ class SynGauss(SynSignal):
         max intensity of peak
     sigma : number, optional
         Default is 1.
-    noise : {'poisson', 'uniform', None}
+    noise : {'poisson', 'uniform', None}, optional
         Add noise to the gaussian peak.
-    noise_multiplier : float
+    noise_multiplier : float, optional
         Only relevant for 'uniform' noise. Multiply the random amount of
         noise by 'noise_multiplier'
+    random_state : numpy random state object, optional
+        np.random.RandomState(0), to generate random number with given seed
 
     Example
     -------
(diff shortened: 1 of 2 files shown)


licode:
@danielballan

ophyd/sim.py
```diff
@@ -336,6 +336,8 @@ class SynGauss(SynSignal):
     noise_multiplier : float
         Only relevant for 'uniform' noise. Multiply the random amount of
         noise by 'noise_multiplier'
+    random_state : numpy random state object
```
danielballan: Can you add `, optional` after the type here? While you are at it, the parameters `noise` and `noise_multiplier` should be marked optional as well.

ophyd/sim.py
```diff
@@ -344,18 +346,19 @@ class SynGauss(SynSignal):
...
         self._motor = motor
-
+        if random_state is None:
+            random_state = np.random
         def func():
             m = motor.read()[motor_field]['value']
             v = Imax * np.exp(-(m - center) ** 2 / (2 * sigma ** 2))
             if noise == 'poisson':
-                v = int(np.random.poisson(np.round(v), 1))
+                v += int(random_state.poisson(np.round(v), 1))
```
danielballan: Actually, the existing code a correct implementation of Poisson noise. You can just rebase and remove the commit 28c64e6 where this was done.

ophyd/sim.py
```diff
@@ -402,22 +407,23 @@ class Syn2DGauss(SynSignal):
...
-                 **kwargs):
+                 random_state=None, **kwargs):
 
         if noise not in ('poisson', 'uniform', None):
             raise ValueError("noise must be one of 'poisson', 'uniform', None")
         self._motor = motor0
         self._motor1 = motor1
-
+        if random_state is None:
+            random_state = np.random
```
danielballan: :+1:

ophyd/sim.py
licode: good catch. I will add optional after the type.

ophyd/sim.py
licode: 👍

danielballan:
Note to reviewers: ensure that the stray 'tag v0.6.1' in this PR is stripped out before merging!

jrmlhermitte:
maybe add a test as well?

licode:
sounds good. I will add a test.

ophyd/tests/test_sim.py
```diff
@@ -0,0 +1,37 @@
...
+    for i in range(2):
+        s = np.random.RandomState(0)
+        noisy_det = SynGauss('noisy_det', motor, 'motor', center=0, Imax=1,
+                             noise='uniform', sigma=1, noise_multiplier=0.1,
+                             random_state=s)
+        noisy_det.trigger()
+        d = noisy_det.read()['noisy_det']['value']
+        dlist.append(d)
+    assert dlist[0] == dlist[1]
+
```
jrmlhermitte: I would try a test without random_state and run
```python
for i in range(2):
    np.random.seed(0)
    noisy_det = SynGauss(...)
```
so basically try it without `random_state` flag.

licode: done.

danielballan: Good suggestion, @jrmlhermitte. Looks good, @licode.

danielballan:
I'm re-running Travis. Once it's happy this is good to merge.

jrmlhermitte:
looks good!

PR #478, opened 2017-11-03
Fix fake PV

### LINKED ISSUE #477: FakeEpicsPV Callbacks
I am currently in the process of migrating SLAC's devices for compatibility with latest (we had previously frozen the version for a bit while we worked on other things). I'm having trouble migrating our unit tests, which previously used a modified version of `FakeEpicsPV` but now include the class from ophyd's `test/conftest.py` due to all the changes with the `control_layer`.

My issue is that the class's callbacks are stored in `WeakValueDictionary`:

https://github.com/NSLS-II/ophyd/blob/master/ophyd/tests/conftest.py#L51-L54

Keeping this as a `WeakValueDictionary`, all of my callbacks are being instantly garbage collected, so I can't write unit tests for things like returning a working `Status` object after a `set` call if they depend on an EPICS subscription.

My questions are:
1. What is the purpose of using a `WeakValueDictionary` here?
2. What are the negative consequences of changing this to a normal `dict` object? (my tests and the ophyd tests pass with this change)
3. If there are consequences, do I need to handle this by keeping a reference to every callback in all of my device implementations?

tacaswell:
I sneaked in some style and warning changes as well.

closes #477

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -1,8 +1,6 @@
 import asyncio
-import copy
 import time as ttime
 from collections import deque, OrderedDict
-from threading import RLock
 import numpy as np
 import random
 import threading
@@ -697,6 +695,7 @@ class InvariantSignal(SynSignal):
         for k in res:
             res[k]['timestamp'] = 0
         return res
+
     def __repr__(self):
         return "<INVARIANT REPR>"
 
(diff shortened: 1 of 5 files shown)


tacaswell:
@teddyrendahl I think you have merge power ;)

teddyrendahl:
Thoughts on eliminating the lock in the `.put` method per https://github.com/NSLS-II/ophyd/issues/477#issuecomment-341784217?
https://github.com/NSLS-II/ophyd/blob/d63d22789dbbf6fad36312008df982e267b190c7/ophyd/tests/conftest.py#L209-L214

tacaswell:
Eliminating locks that cause deadlocks is usually not the right solution as the deadlock is telling you here is a race condition that needs to be ironed out.

danielballan:
@ZLLentz @teddyrendahl If you think this is worth merging as is into 1.0.0 (final), which I plan to tag late in the day on Monday, please merge.

It seems like we should take our time ironing out the deadlock / race condition carefully. We'll be tagging a bug-fix release in 2-3 weeks anyway, so the stakes are low.

PR #479, opened 2017-11-03
ENH: Add OphydObj.unsubscribe_all()

danielballan:
Suggested by @stuartcampbell during an ophyd tutorial today

### DIFF
--- a/ophyd/tests/test_ophydobj.py
+++ b/ophyd/tests/test_ophydobj.py
@@ -116,6 +116,27 @@ def test_unsubscribe():
     # check multi unsubscribe
     test_obj.unsubscribe(cid)
     test_obj.clear_sub(increment)
+    assert hit == 1
+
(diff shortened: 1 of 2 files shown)


danielballan:
Added a test.

ophyd/tests/test_ophydobj.py
```diff
@@ -116,6 +116,27 @@ def increment(**kwargs):
     # check multi unsubscribe
     test_obj.unsubscribe(cid)
     test_obj.clear_sub(increment)
+    assert hit == 1
```
jrmlhermitte: why this extra line?

danielballan: Just a little extra paranoia, to be sure that unsubscribing doesn’t fire the subscription one last time. (It shouldn’t.)

PR #480, opened 2017-11-03
pedant modification

jrmlhermitte:
hopefully travis stays happy

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -560,7 +560,7 @@ class BlueskyInterface:
                 stage_sigs[k] = v
 
         # Read current values, to be restored by unstage()
-        original_vals = {sig: sig.get() for sig, _ in stage_sigs.items()}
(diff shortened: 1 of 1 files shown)


ophyd/device.py
```diff
@@ -560,7 +560,7 @@ def stage(self) -> List[object]:
                 stage_sigs[k] = v
 
         # Read current values, to be restored by unstage()
-        original_vals = {sig: sig.get() for sig, _ in stage_sigs.items()}
+        original_vals = {sig: sig.get() for sig in stage_sigs.keys()}
```
danielballan: In fact, we don't even need `.keys()` here!

jrmlhermitte:
haha yes!

ophyd/device.py
jrmlhermitte: haha yes!

PR #481, opened 2017-11-07
API: Add empty hints for SynSignalWithRegistry.

danielballan:
This makes BestEffortCallback work until LiveTable can be fixed.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -595,6 +595,21 @@ class SynSignalWithRegistry(SynSignal):
         self._path_stem = None
         self._result = {}
 
+        self._hints = None
+
+    @property
+    def hints(self):
+        if self._hints is None:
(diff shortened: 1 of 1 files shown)


tacaswell:
Have been using 

```python
@property
def hints(self):
    if self._hints is None:
         return {'fields': []}
    return self._hints

@hints.setter
def hints(self, val):
    self._hints = dict(val)
```

on many of the objects that have a default hints.

danielballan:
I see. That approach makes it easier to revert to the default after making some change. Force-pushed.

tacaswell:
Is there already a default value for `self._hints` in the init?

danielballan:
Oof! No. Fixed and force-pushed. Ran Tim's bug report script against this. Indeed it fails on my previous commit but now works.

PR #482, opened 2017-11-07
DOC: Use autosummary the same way we do in bluesky.

danielballan:
- Nicer URL (`_as_gen` -> `generated/`)
- Cleanup generated rst files as part of `make clean`

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -55,6 +55,7 @@ help:
 
 clean:
 	rm -rf $(BUILDDIR)/*
+	rm -rf source/generated/*.rst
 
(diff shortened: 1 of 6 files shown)


tacaswell:
Can you add the generated path to `.gitignore`?

danielballan:
No need to add it -- it's [already there](https://github.com/danielballan/ophyd/blob/044c3408f7aae8b4fd9ccf08491f81601d3eb429/.gitignore#L69). We must have generated the gitignore from a template that catches both `_as_gen/` and `generated/`.

Issue #459, opened 2017-10-23
Test against multiple versions of EPICS


danielballan:
We should follow the pattern we use in caproto .travis.yml.

danielballan:
The blocking issue here is getting a script to build Area Detector. If I understood right, @gwbischof has been playing around with this.

PR #484, opened 2017-11-10
DOC: add documentation on how to stand up docker for testing

tacaswell:


danielballan:
Nice! Glad to see that on-boarding is forcing us to write down our institutional knowledge. It's likely we will blow this Docker stuff away with #459 but that could take awhile.

doc/source/docker.rst
```diff
@@ -0,0 +1,65 @@
...
+   #!/usr/bin/bash
+
+   export DOCKER0_IP="172.17.0.1"
+   export EPICS_CA_ADDR_LIST=$( echo $DOCKER0_IP | sed -e 's/^\([0-9]\+\)\.\([0-9]\+\)\..*$/\1.\2.255.255/' )
+   export EPICS_CA_AUTO_ADDR_LIST="no"
+   export EPICS_CA_MAX_ARRAY_BYTES=10000000
+
+
+and to run docker with the correct images (assuming the preceding code
+block is saved in :file:`epics_export`) ::
```
danielballan: Would it be convenient to add such a file to the repo itself? The web community conventionally names this file `env` or `.env`.

tacaswell:
Even if we blow away docker on travis, I still prefer it for local testing purposes.

jrmlhermitte:
worked for me. Just had to wring out some docker setup details. I can list them if you think they are good to add.

danielballan:
Please do! It would be great to have a complete set of instructions for getting a working test setup.

jrmlhermitte:
Actually, I followed this documentation starting from scratch and it works just fine, so it may be better to leave as is:
1. [Installing docker on ubuntu](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#uninstall-docker-ce)
2. [Configuring docker](https://docs.docker.com/engine/installation/linux/linux-postinstall/)
(Allowing to run as non-root, running at startup etc)

tacaswell:
To some degree we run this on CI all the time (as we run against docker for ophyd (for now)) and these instructions were extracted from the `.travis.yml` file.

PR #489, opened 2017-11-29
FIX: Reorder set_and_wait to avoid race condition

ZLLentz:
`set_and_wait` currently accesses `signal.tolerance` on the first line. This is a problem because the `tolerance` property is wrapped with a `raise_if_disconnected` decorator, which prevents the whole method from being run if the `EpicsSignal` object has not finished connecting.

This happened to me with the `AreaDetectorComponent` objects. Since these are the only `Component` objects that are set as `lazy=True`, there was no guarantee that the connection finished in time for my `set_and_wait` call.

I expect that these `epics_pvs` utils should work even if the connection hasn't finished yet, so here I've reordered the method so that the `put` and `get` calls happen first, which call `wait_for_connection` on the `_write_pv` and `_read_pv` attributes respectively. This guarantees that the connection has happened in time for the tolerance check.

### DIFF
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -225,14 +225,15 @@ def set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
     ------
     TimeoutError if timeout is exceeded
     """
+    signal.put(val)
+    expiration_time = ttime.time() + timeout if timeout is not None else None
+    current_value = signal.get()
+
     if atol is None and hasattr(signal, 'tolerance'):
         atol = signal.tolerance
     if rtol is None and hasattr(signal, 'rtolerance'):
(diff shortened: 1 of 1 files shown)


tacaswell:
Ah!  This is probably the source of the failures we saw when we flipped the default, everything blew up, and then we flipped the default back with out chasing it to ground.

The test failures are failure to import pytest in the conda envs, restarted to see if it is a transient thing or of we need to dive down a packaging rabbit hole.

danielballan:
I think accidentally let this languish because some Travis builds were randomly failing. I re-ran the builds, and one more passed; the other failed on something unrelated.

Thanks @ZLLentz and sorry for the hold-up. Feel free to kick the Travis builds if this happens again. Do you have the permissions to do so?

PR #491, opened 2017-12-07
DOC: added all projects toctree to ophyd

mrakitin:
## Description

Add all projects toctree to the ophyd documentation.

## Motivation and Context

Motivated by https://github.com/NSLS-II/docs/issues/73.

## How Has This Been Tested?

Tested locally by building the documentation (required to do `sudo apt-get install graphviz`).

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -69,3 +69,34 @@ other control systems.
    :caption: API
 
    api
+
+.. toctree::
+   :hidden:
+   :caption: Data Collection
+
+   bluesky <https://nsls-ii.github.io/bluesky>
+   ophyd <https://nsls-ii.github.io/ophyd>
+
+.. toctree::
+   :hidden:
+   :caption: Data Access and Management
+
+   databroker <https://nsls-ii.github.io/databroker>
+   amostra <https://nsls-ii.github.io/amostra>
+   datamuxer <https://nsls-ii.github.io/datamuxer>
+   suitcase <https://nsls-ii.github.io/suitcase>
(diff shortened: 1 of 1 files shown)


jrmlhermitte:
this looks good to me, but will sphinx fail then if you had to install graphviz?
I did some digging around, the travis file builds the docs, clones an nsls2-ci repo and eventually runs this file:
https://github.com/NSLS-II/nsls2-ci/blob/master/make-docs.sh

does this mean we should add `sudo apt-get install graphviz` to the travis.yml file or something similar?
@danielballan may have suggestions?

jrmlhermitte:
nevermind, i see it in `DOCS_PIP_DEPS` in travis.yml file. looks good!

doc/source/index.rst
```diff
@@ -69,3 +69,27 @@ other control systems.
...
+
+   bluesky <https://nsls-ii.github.io/bluesky>
+   ophyd <https://nsls-ii.github.io/ophyd>
+
+.. toctree::
+   :hidden:
+   :caption: Data Access and Management
+
+   databroker <https://nsls-ii.github.io/databroker>
+   amostra <https://nsls-ii.github.io/amostra>
```
danielballan: As amostra and datamuxer and both unmaintained, I think we should not advertise them so prominently (for now). Would you mind removing them?

mrakitin:
I updated doctree by excluding unsupported codes and adding caproto docs link (see https://github.com/NSLS-II/docs/issues/82 for motivation).

PR #493, opened 2017-12-12
Ad queue draining

tacaswell:
This is no way ready to merge yet and I will rebase / squash / force-push at will.

The end goals of this work are:

 1. be sure there are no images in the processing pipelines we care about when.
 2. enable running collection with blocking callbacks disabled.
 3. lay ground work for sophisticated 'modes' management of AD devices.

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -208,8 +208,14 @@ class ADBase(Device):
 
         return ret
 
-    def get_asyn_digraph(self):
+    def get_asyn_digraph(self, *, enable_rule=None):
         '''Get the directed graph of the ASYN ports
+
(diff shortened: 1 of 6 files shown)


danielballan:
@tacaswell Safe to close this now, given fixes to AD?

tacaswell:
I think we will eventually want to salvage most of this (for example, only caring about plugins that are enabled when walking the processing tree).

Having this for old versions of AD is also not a bad thing.

klauer:
@tacaswell do you have details on what's left to fix/salvage from this?

tacaswell:
I think that this can be abandoned due to the changes Mark put into 3.3 for us.

PR #499, opened 2017-12-22
Mnt sundry

tacaswell:


ophyd/utils/__init__.py
```diff
@@ -75,3 +75,46 @@ def wrapper(f):
...
+
+
+def ducks_from_namespace(attrs, *, ns=None):
+    '''Get instances that have all of attributes.
+
+    "Ducks" is a reference to "duck-typing." If it looks like a duck....
+
+    Parameters
+    ----------
+    attr : Union[str, Iterable[str]]
```
danielballan: I don't like that types are randomly creeping into the codebase without a plan. I think "string or iterable of strings" is better documentation for humans than `Union[str, Iterable[str]]`, at least today.

tacaswell: I am resisting (on Steve's advice) putting them in as annotations, but I like having them in the docstrings as we could pragmatically extract them if want to.

It also gives us a consistent vocabulary which has been thought through carefully.

danielballan: Do you mean "programatically" or "pragmatically"? If "pragmatically", then I would say just stash them in a comment. I think we should either adopt this or not adopt it; I don't like the randomness of it.

tacaswell: 'programitically'.

I don't think it is worth going through and changing existing docstrings to use typing-style types, but is worth using for new docstrings and as we touch things.

danielballan: OK, I won't litigate this here.

PR #501, opened 2018-01-03
DOC: update dispatch docstring to match the code

tacaswell:


ophyd/areadetector/detectors.py
```diff
@@ -47,17 +47,17 @@ class DetectorBase(ADBase):
     def dispatch(self, key, timestamp):
         """Notify plugins of acquisition being complete.
 
-        When a new acquisition is finished, this method is called with a
+        When a new acquisition is started, this method is called with a
         key which is a label like 'light', 'dark', or 'gain8'.
```
mrakitin: Everything looks good to me. I only have a question regarding `gain8` - is it a valid label?

danielballan: Yes, any string is a valid label. The first application of this ability to use multiple labels was a detector that captured images with a variety of gain settings, all as part of one "trigger".

mrakitin: Thanks for the explanation, Dan!

PR #506, opened 2018-01-31
FIX: Andor does not support 'Multiple' mode.

danielballan:
Kaz pointed out that, per the documentation, Andor only supports
'Fixed' and 'Continuous'.

https://millenia.cars.aps.anl.gov/software/epics/andor3Doc.html

(This site is down currently but the cached Google version is available.)

This change will take precedence over ``SingleTrigger``'s default via MI.

Before merging we should survey other detectors' docs to see if any others don't support 'Multiple'.

### DIFF
--- a/ophyd/areadetector/detectors.py
+++ b/ophyd/areadetector/detectors.py
@@ -92,10 +92,24 @@ class AndorDetector(DetectorBase):
     _html_docs = ['andorDoc.html']
     cam = C(cam.AndorDetectorCam, 'cam1:')
 
+    def __init__(*args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # The default cam.image_mode imposed by SingleTrigger is 'Multiple'
+        # but Andor does not support that mode, so we override that default
+        # here.
+        self.stage_sigs['cam.image_mode'] = 0  # 'Fixed'
+
 
 class Andor3Detector(DetectorBase):
     _html_docs = ['andor3Doc.html']
     cam = C(cam.Andor3DetectorCam, 'cam1:')
+
+    def __init__(*args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # The default cam.image_mode imposed by SingleTrigger is 'Multiple'
+        # but Andor does not support that mode, so we override that default
+        # here.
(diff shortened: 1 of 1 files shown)


danielballan:
There was a miscommunication. It's not _Andor_ that fails to implement 'Multiple' mode -- it's Area Detector v2.3. I'm told NSLS-II is in the process of moving to the present, to AD v3.x, so I'll close this issue.

As always, this problem can be addressed in configuration at runtime by changing the contents of `your_detector.stage_sigs` on the instance.

kgofron:
I meant AreaDetector Release 3.x is using Fixed/Continuous, and no 'Multiple' data acquisition modes. The change was probably introduced earlier.

danielballan:
@kgofron Are you saying that AD 3.x _removed_ support for 'Multiple' mode? (I don't think that's the case, but we have so few AD 3.x on the floor at NSLS-II that I'm not _certain_ I'm right about that....) For all detectors? For Andor detectors?

hhslepicka:
@kgofron I am not sure where is that since from ADCore Master we have: https://github.com/areaDetector/ADCore/blob/master/ADApp/Db/ADBase.template#L317

Would you mind pointing to where you saw it?

hhslepicka:
At Andor ImageMode is defined as: https://github.com/areaDetector/ADAndor/blob/master/andorApp/Db/andorCCD.template#L46

And Andor3 uses the definitions from ADCore:
https://github.com/areaDetector/ADAndor3/blob/master/andor3App/Db/andor3.template

kgofron:
I checked more carefully, and my initial statement was incorrect. Andor3  is is an exception.
The Fixed/Continous mode is specific to Andor3 (Andor Neo, Zila detectors) AD module, supported by Andor SDK3.  Only these detectors do not have 'Multiple' ImageMode. 
Line 1393 of andor3.cpp defines the Andor3 modes as 'Fixed/Continuous' only.

image

danielballan:
@kgofron Would you post a GitHub link? It is easier to explore than a screenshot of your desktop.

kgofron:
https://github.com/areaDetector/ADAndor3/blob/master/andor3App/src/andor3.cpp#L1393
The GitHub link to the driver setting

PR #508, opened 2018-02-04
For all plugin sub-devices, include all attrs in default_read_attrs.

danielballan:
As pointed by @jrmlhermitte in an internal document, settting

```py
cam05.stats1.read_attrs = ['centroid']
```

does not have any effect on what is read because

```py
cam05.stats1.centroid.read_attrs == []
```

I think all plugin subdevices should include all attrs as read_attrs
by default. The effective default behavior will not change because

```py
cam05.stats1.read_attrs == []
```

but if ``'centroid'`` is added, its x and y components should come
along for the ride.

- Extend DynamicDeviceComponent to accept optional default_read_attrs and
  default_configuration_attrs arguments, which is will assign as class
  attributes on the class it dynamically creates.
- Systematically add default_read_attrs=(...) to every DDC on every plugin.

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -193,7 +193,8 @@ class PluginBase(ADBase):
                               (('height', 'ArraySize1_RBV'),
                                ('width', 'ArraySize0_RBV'),
                                ('depth', 'ArraySize2_RBV'))),
-                     doc='The array size')
+                     doc='The array size',
+                     default_read_attrs=('height', 'width', 'depth'))
 
     bayer_pattern = C(EpicsSignalRO, 'BayerPattern_RBV')
     blocking_callbacks = C(SignalWithRBV, 'BlockingCallbacks',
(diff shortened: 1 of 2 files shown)


jrmlhermitte:
great! How scalable is this? Do we have to manually add defaults for each new device we make?

danielballan:
You need to manually add defaults only if you want to override the _default_ default (ha) of empty read_attrs and configuration_attrs. I think that default default still makes sense: given the sprawling number of signals on most real-life devices, we need the device implementator to tell us which signals/subdevices are probably important.

danielballan:
Having just read your related comment on BC, I’ll elaborate a bit.

Alternatives:

- Make the default default “put everything in read_attrs”. This would make the out-of-the-box Devicw very slow to read and would generate a ton of data that no one actually needs.
- Provide some sentinel that means “all”, as in default_read_attrs=ophyd.ALL or something like that. I considered this option but thought spelling out the attrs was more obvious and not too tedious. That’s debatable for sure.

In the case of plugins, we happen to either want all the attrs in by default or None, but in other modules we select a hand-picked few. I guess I consider it good to encourage that thoughtful moment by making people spell out which attrs to include, but I’m open to other ideas that might make this more convenient.

jrmlhermitte:
Sounds good. I like your approach.
Yes, adding all attrs by default might be too messy and is seldom the configuration scientists want.
I agree about `all`, I prefer being more explicit rather than using some `ophyd.ALL` flag.

Beamline scientists should also be aware of how to debug this, since likely they want customization at this level. So i'd say this is good in that we make our best effort but we should not go any further.

I think this is good to go. Should we ask someone else's blessing? (Also, I restarted the py36 build which erred on a timeout.)

danielballan:
I'm not too worried that this will be controversial, but I'd feel better if we got @klauer's blessing.

klauer:
I didn't check all of the defaults that you've specified, but I agree with this in principle. 👍

PR #509, opened 2018-02-12
ENH: only let channels which have a name into read/config list

tacaswell:


ophyd/scaler.py
```diff
@@ -123,8 +123,14 @@ def __init__(self, *args, **kwargs):
...
-        self.channels.read_attrs = ['chan01']
-        self.channels.configuration_attrs = ['chan01']
+        active_channels = []
+        for s in self.channels.component_names:
+            ch_name = getattr(self.channels, s).name
+            if ch_name:
+                active_channels.append(ch_name)
+
+        self.channels.read_attrs = list(active_channels)
+        self.channels.configuration_attrs = list(active_channels)
```
mrakitin: Is `list` needed in this case? It's already defined as list above

tacaswell: Yes, if you just assigned then `self.channels.read_attrs is self.channels.configuration_attrs` would be `True` and appending or popping from one list would affect the other.  That may actually be sensible in this case, but would be _very_ surprising.

PR #510, opened 2018-02-14
Fix version

jrmlhermitte:
fix the case where `epics.__version__` may not be a version string

### DIFF
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -206,7 +206,7 @@ def get_pv_form(version):
 
     try:
         version = parse_version(_fix_git_versioning(version))
-    except ValueError:
+    except TypeError:
         version = None
 
     if version is None:
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -204,7 +204,10 @@ def get_pv_form(version):
     except ImportError:
         parse_version = _naive_parse_version
 
-    version = parse_version(_fix_git_versioning(version))
+    try:
+        version = parse_version(_fix_git_versioning(version))
+    except ValueError:
+        version = None
 
     if version is None:
         warnings.warn('Unrecognized PyEpics version; using local timestamps',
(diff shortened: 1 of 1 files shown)


jrmlhermitte:
issue happened here: https://github.com/NSLS-II/databroker/blob/master/doc/source/tutorial.rst

where the output was:
```
/home/julienl/research/projects/ophyd/ophyd/_pyepics_shim.py in get_pv_form(version)
   206
   207     print("version is {}".format(version))
--> 208     version = parse_version(_fix_git_versioning(version))
   209
   210     if version is None:

/home/julienl/anaconda3/lib/python3.6/site-packages/pkg_resources/__init__.py in parse_version(v)
   209 def parse_version(v):
   210     try:
--> 211         return SetuptoolsVersion(v)
   212     except packaging.version.InvalidVersion:
   213         return SetuptoolsLegacyVersion(v)

/home/julienl/anaconda3/lib/python3.6/site-packages/pkg_resources/_vendor/packaging/version.py in __init__(self, version)
   198     def __init__(self, version):
   199         # Validate the version and parse it into pieces
--> 200         match = self._regex.search(version)
   201         if not match:
   202             raise InvalidVersion("Invalid version: '{0}'".format(version))

TypeError: expected string or bytes-like object
```

The variable `epics.__version__` was:
`<MyMock object at 0x7f0d7f23d860>`

This should likely be fixed upstream, but probably good to handle here, as @danielballan 's suggestion

tacaswell:
It is probably better to fix the Mock object that is getting passed in than to put the cut-out in ophyd.

danielballan:
@tacaswell That sounds right, but a defensive cut-out here (and maybe a GH issue reminding us to remove it in a year) is good defense in depth -- aiming at old versions of pyepics.

jrmlhermitte:
I like @tacaswell . Since the issue was fixed in bluesky [here](https://github.com/NSLS-II/databroker/pull/346), perhaps we can close this for now and re-open if we find another issue?

danielballan:
Works for me. Feel like submitting your first pyepics PR and fixing this upstream?

PR #513, opened 2018-02-21
enable plugin before staging

bsobhani:


ophyd/areadetector/plugins.py
```diff
@@ -97,6 +98,8 @@ def __init__(self, *args, **kwargs):
     port_name = C(EpicsSignalRO, 'PortName_RBV', string=True)
 
     def stage(self):
+        if self._enable_on_stage:
+           set_and_wait(self.enable,1)
```
danielballan: For PEP8 compliance, there should be a space after the comma here.

ophyd/areadetector/plugins.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, *args, **kwargs):
                                           self._plugin_type,
                                           self.plugin_type.get(), self.prefix))
 
+        self._enable_on_stage=True
```
danielballan: For PEP8 compliance, there should be spaces around the `=` sign when it's used for variable assignment -- here and a couple places below.

bsobhani:
Sure, I will make these changes.

tacaswell:
I think this will also need a change to `unstage` to disable an plugins that were enabled during stage (to clean up after our selves).

I am confused why this fixes the problem as via https://github.com/NSLS-II/ophyd/blob/13ecb70943f5d074cce525af449d9131acbc756b/ophyd/device.py#L588-L602 the values in stage_sigs are run through `set_and_wait` in order so how is this functionally different than the current behavior?

danielballan:
> I think this will also need a change to unstage to disable an plugins that were enabled during stage (to clean up after our selves).

I recall that this came up in the past and we decided that "enabling" is allowed to be special and that we should not disable.

> How is this functionally different than the current behavior?

This ensures that enabling _definitely_ happens first. It seems like the enable signal was not landing at the top of the OrderedDict, and thus ophyd was hanging on trying to set a PV on a disabled plugin.

bsobhani:
I did notice on Friday that in addition to enabling HDF5 I had to set some other parameters to get the camera to work. I think NumCaptures (or something similar) was initially 0 and I had to set it to 1, or I would get the same error. I might have had to make some other changes as well. Could this be why the tests are failing?

danielballan:
Could be. Worth trying some things and once we get the tests to pass we can revisit what the proper behavior should be. For example, should we just raise a helpful error if NumCaptures is 0 rather than quietly setting it to 1?

bsobhani:
What is the current behavior of the code - does it set NumCaptures to 1 already?

danielballan:
I don't think so.

ophyd/areadetector/plugins.py
```diff
@@ -105,15 +108,16 @@ def enable_on_stage(self):
 
         a convenience method for adding ('enable', 1) to stage_sigs
         """
-        self.stage_sigs['enable'] = 1
+        self.enable_on_stage = True
```
ZLLentz: Is this a typo? It looks to me like this method just overrides itself with `True`.

bsobhani: I think I might have forgotten an underscore

bsobhani:
I think helpful error messages would be a good idea if it's complicated to do automatically. But that was a huge mistake I made, maybe the build will work this time.

klauer:
Have you considered keeping the original `stage_sigs`-based functionality, but re-ordering the `OrderedDict` keys such that `Enable` comes first? That way, all of your staging information remains in `stage_sigs`.

I did something similar in [hxntools](https://github.com/NSLS-II-HXN/hxntools/blob/master/hxntools/detectors/utils.py#L26) (in-place move key to beginning - without a docstring, of course, sorry 'bout that).

bsobhani:
I will give it a try.

bsobhani:
I tried moving enable to the front of stage_sigs in the PluginBase init method. I am not sure if this will work because I do not see when and where the stage_sigs of the derived classes are being added, so I am not certain if 'enable' will be moved to the front after all the other stage_sigs have been added, or if it will be before.

ophyd/areadetector/plugins.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, *args, **kwargs):
                                           self.plugin_type.get(), self.prefix))
 
         self.enable_on_stage()
+        self.stage_sigs.move_to_end('enable', last=False)
```
klauer: Nice find on `move_to_end` (I'll pretend I had a reason not to use that when I came up with the hxntools solution above...)

danielballan:
Staging _generally_ restores the device to the state that it found it in, but there are exceptions. For example, when we set a new filename on stage, we don't reset the _previous_ filename on unstage; that would not make sense.

My original thinking was, this should be an exception too. If ophyd needs a plugin enabled, it should enable it and leave it enabled when it is done. Disabling a plugin can be a destructive action: some plugins, like the HDF5 plugin in streaming capture mode, need to be "primed", so I would suggest that ophyd should avoid disabling plugins, ever. Within the plugin, there are other ways to deactivate it (AutoSave off, Capture off, etc.).

ophyd/areadetector/plugins.py
bsobhani: I didn't notice hxntools was a blue link or I would have used your function. To be honest I googled how to move an OrderedDict item to the beginning.

bsobhani:
Yes, that makes sense, and for the sake of curiosity too I would like to see how it is implemented. It is strange to me that set_and_wait(self.enable, 1) causes failures when I put it in PluginBase, but putting 'enable' in stage_sigs doesn't, when it seems like that is making BlueskyInterface.stage() also do set_and_wait(self.enable, 1).

danielballan:
Thanks, @bsobhani. This will make it onto the floor this week.

PR #516, opened 2018-02-28
ENH: allow 'derived_from' input do DerivedSignal to be a string

tacaswell:
This defers grabbing the actual Signal object until init time.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -245,15 +245,17 @@ class DerivedSignal(Signal):
 
         Parameters
         ----------
-        derived_from : Signal
-            The signal from which this one is derived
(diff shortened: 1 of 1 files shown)


danielballan:
Context: without this change, `DerivedSignal` is unusable inside a `Device` because it needs a reference to a `Signal` instance at init time.

klauer:
👍 (aside: why is doc building failing, though?)

danielballan:
> why is doc building failing

The new version of sphinx is fussier. This update broke bluesky as well. I have pinned it to 1.6.7 for now (in master, as part of my doctr work). I opened an #517 to remind us to address this at some point.

PR #519, opened 2018-03-01
ENH: Implement collect_asset_docs in backward-compatible way.

danielballan:
This implements ophyd's piece of https://github.com/NSLS-II/databroker/pull/322.

Ophyd Area Detector FilePlugins no longer need a Registry object but (for now)
they still accept one. If the plugin has a Registry, it registers the
Resource/Datum with that registry, as before.  It also caches the Resource and
Datum documents and provides them through a new method, ``collect_asset_docs``,
which the upcoming release of bluesky will call. It does this whether or not
it has a Registry.


tacaswell:
We apparently use `.reg` alot in the tests...

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +247,66 @@ def write_path_template(self, val):
...
         self._locked_key_list = False
         self._datum_uids.clear()
         super().stage()
         if self.enable.get() and self._reg is None:
             raise RuntimeError('The plugin {!r} is enabled, but '
                                'the Registry (self._reg) is `None`')
+        # Subclasses will assemble resource_kwargs and call
+        # self._generate_resource(resource_kwrags)
+
+    def _generate_resource(self, resource_kwargs):
```
tacaswell: do we want to move this method and all it's state to a stand-alone class in databroker?

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +247,66 @@ def write_path_template(self, val):
...
-        reading = {'value': uid, 'timestamp': timestamp}
+        # This is temporarily more complicated than it will be in the
+        # future.  It needs to support old configurations that have a
+        # registry.
+        datum = {'resource': self._resource_uid,
+                 'datum_kwargs': datum_kwargs}
+        if self._reg is not None:
+            # If a Registry is set, we need to allow it to generate the
+            # datum_id for us.
+            datum_id = self._reg.register_datum(
```
jrmlhermitte: I like how by introspecting the `datum_id`, we can have a hint about how it was generated for free. That will be useful.
I would have suggested passing a new `version` kwarg in. However, this would require changing the function in `databroker`. I don't think it's worth the risk for the minimal gain it would give.

ophyd/areadetector/filestore_mixins.py
jrmlhermitte: I agree and would say the same for datum, if I understand you @tacaswell .
In my opinion, this would help a newcomer better understand the document structure (by inspecting the generation functions). It won't look nice, but might help when we have libraries other than ophyd which may generate documents.

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +247,66 @@ def write_path_template(self, val):
...
+        else:
+            self._resource_uid = new_uid()
+        resource['uid'] = self._resource_uid
+        self._asset_docs_cache.append(('resource', resource))
 
     def generate_datum(self, key, timestamp, datum_kwargs):
         "Generate a uid and cache it with its key for later insertion."
+        i = next(self._datum_counter)
         datum_kwargs = datum_kwargs or {}
+        datum_kwargs.update({'point_number': i})
```
jrmlhermitte: wait, **every** file handler will get a `point_number` now? Is that safe?
For example, here in `eiger_io`:
https://github.com/NSLS-II/eiger-io/blob/master/eiger_io/fs_handler.py#L142

If I had to vote, I'd say it's worth the risk, but it's something we're going to have to bring attention to the whole group, and monitor these next few weeks (upgrade handlers to take in a `point_number`). I may have misunderstood something though. It might be good for another fresh pair of eyes to review this.
Here is Eiger's file plugin, inheriting from `FileStoreBase` [here](https://github.com/NSLS-II-CHX/profile_collection/blob/master/startup/20-area-detectors.py#L67).

ophyd/areadetector/filestore_mixins.py
tacaswell: You definitely should not depend on anything about the structure of the datum stings for _anything_.

This is the code path that is going to get eliminated eventually.

ophyd/areadetector/filestore_mixins.py
tacaswell: Good catch!  This should be moved back to the sub-class it was pulled up out of.  The eiger filestore plugin must inherit from the base class rather than `FileStoreIterativeWrite` like most of the other plugins.

ophyd/areadetector/filestore_mixins.py
tacaswell: ah, good point about not wanting to have a databroker dependency in ophyd.  Should probably be factored out into a helper class though.

jrmlhermitte: actually i was saying the opposite, it should use a databroker function since it is databroker, but you're right, that adds a dependency. in the end, that's one reason there are schemas for these i guess. i think we've just both self-countered our arguments.

ophyd/areadetector/filestore_mixins.py
jrmlhermitte: that could be a problem since there will be two dependencies on the iterator `datum_counter`.
Actually, I don't know if I like `{resource_id}/{point_number}` as the uid. Can we just have it be a uid?
What if a user decides they want  `_datum_counter` to be a sequence id or do something else funny?
Or what if the user finds a way to use the same resource for multiple `stage/unstage` runs?

I understand we want to keep the structure of the read datums, so how about:
`{resource_uid}/{datum_uid}` where `datum_uid=str(uuid.uuid4()[:10])` or something?

tacaswell: The current 'status-quo' is just `str(uuid.uuid4())` but as part of adding the 'column_hdf5' asset registry and the performance work around that (https://github.com/NSLS-II/databroker/pull/216 for the code, did not keep good notes on the performance work) we are starting to move to `datum_id = {resourc_uid}/{N}` to a) support random and bulk access in column case and b) generating the uuids for each datum was starting to contribute significantly to run time for lots of datum (on my laptop it is 6 µs per uuid which for a 2k by 2k scan is ~24 seconds total time)

ophyd/areadetector/filestore_mixins.py
```diff
@@ -158,10 +161,13 @@ def __init__(self, *args,
...
             raise ValueError("write_path_template is required")
         self.reg_root = root
+        self.path_semantics = path_semantics
         self.write_path_template = write_path_template
         self.read_path_template = read_path_template
 
-        self._point_counter = None
+        self._resource_uid = None
+        self._datum_counter = None  # becomes an itertools.count() in stage()
+        self._asset_docs_cache = deque()
```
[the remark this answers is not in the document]
danielballan: Good idea. Done: #532

danielballan:
It _seems_ like these failures are random timeouts, but the failure rate is high, particularly of `caproto/tests/test_examples.py::test_curio_server_example`, so I have reservations.

FWIW, locally, that runs and hangs at the end of the test for me -- but it also did that in the last PR (now `master`) for which we got 8 all builds to pass.

danielballan:
Ignore previous (now deleted) comment. Was in wrong PR tab.

danielballan:
Thanks for the fix, @tacaswell. @mrakitin or @jrmlhermitte, would you take one more look at this and do the honors?

mrakitin:
Sure, on it.

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +292,66 @@ def write_path_template(self, val):
...
+            spec=self.filestore_spec,
+            root=str(self.reg_root),
+            resource_path=str(fn),
+            resource_kwargs=resource_kwargs,
+            path_semantics=self.path_semantics)
+
+        # If a Registry is set, we need to allow it to generate the uid for us.
+        if self._reg is not None:
+            logger.debug("Inserting resource with filename %s", self._fn)
+            # register_resource has accidentally different parameter names...
```
jrmlhermitte: thanks for adding that comment. at least we can remember which one we're aware of that we want to move away from :-D

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +292,66 @@ def write_path_template(self, val):
...
+        fn = PurePath(self._fn).relative_to(self.reg_root)
+        resource, self._datum_factory = resource_factory(
+            spec=self.filestore_spec,
+            root=str(self.reg_root),
+            resource_path=str(fn),
+            resource_kwargs=resource_kwargs,
+            path_semantics=self.path_semantics)
+
+        # If a Registry is set, we need to allow it to generate the uid for us.
+        if self._reg is not None:
```
jrmlhermitte: We should also note with a comment that we are eventually getting rid of this correct?
(just like further below)

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +292,66 @@ def write_path_template(self, val):
...
+            self._resource_uid = self._reg.register_resource(
+                rpath=resource['resource_path'],
+                rkwargs=resource['resource_kwargs'],
+                root=resource['root'],
+                spec=resource['spec'],
+                path_semantics=resource['path_semantics'])
+            resource['uid'] = self._resource_uid
+        # If a Registry is not set, we need to generate the uid.
+
+        self._resource_uid = resource['uid']
```
jrmlhermitte: It's done twice when `_reg` is present but i really like this clear separation (which I think was intentional :), no sarcasm! )

ophyd/areadetector/filestore_mixins.py
```diff
@@ -341,6 +440,9 @@ def stage(self):
             raise IOError("Path %s does not exist on IOC."
                           "" % self.file_path.get())
 
+    def unstage(self):
```
jrmlhermitte: is this necessary?

ophyd/areadetector/filestore_mixins.py
```diff
@@ -241,25 +292,66 @@ def write_path_template(self, val):
...
         self._locked_key_list = False
         self._datum_uids.clear()
         super().stage()
-        if self.enable.get() and self._reg is None:
-            raise RuntimeError('The plugin {!r} is enabled, but '
-                               'the Registry (self._reg) is `None`')
+        # Subclasses will assemble resource_kwargs and call
+        # self._generate_resource(resource_kwrags)
+
+    def _generate_resource(self, resource_kwargs):
```
jrmlhermitte: why an underscore (compared to `generate_datum`)?

ophyd/areadetector/filestore_mixins.py
```diff
@@ -294,6 +392,7 @@ def __init__(self, *args, **kwargs):
                                 ('num_capture', 0),
                                 ])
         self._fn = None
+        self._fp = None
```
jrmlhermitte: what is the motivation between `_fn` and `_fp`? I see `_fp` looks like the template and `_fn` the read path after the template. just curious!

tacaswell: Because mulit-file tiff datasets.

ophyd/areadetector/filestore_mixins.py
tacaswell: no, it is not..

ophyd/areadetector/filestore_mixins.py
tacaswell: generate_datum is part of the public API and the interface that AD uses to find it's children than emit AssetRegistry documents, this is a helper method to centralize some code.

If we could do it again I would call `generate_datum` something like `notify_children_AD_emitted_an_array_that_wrote_to_disk` (but that is a bit wordy....)

ophyd/areadetector/filestore_mixins.py
jrmlhermitte: thanks!

ophyd/utils/__init__.py
```diff
@@ -118,3 +118,8 @@ def ducks_from_namespace(attrs, *, ns=None):
         ns = _filtered_ip_ns()
     return [val for val in ns.values()
             if all(hasattr(val, attr) for attr in attrs)]
+
+
+class DO_NOT_USE:
+    "sentinel value"
+    ...
```
mrakitin: That's neat!

ophyd/areadetector/filestore_mixins.py
```diff
@@ -158,10 +206,13 @@ def __init__(self, *args,
         if write_path_template is None:
             raise ValueError("write_path_template is required")
         self.reg_root = root
+        self.path_semantics = path_semantics
         self.write_path_template = write_path_template
         self.read_path_template = read_path_template
 
-        self._point_counter = None
+        self._resource_uid = None  # wiil be removed
+        self._datum_factory = None
```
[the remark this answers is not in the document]
tacaswell: because we need to generate a new factory everytime we create a new resource which happens everytime we stage.

PR #520, opened 2018-03-07
Enh dummy cl

tacaswell:


danielballan:
In case "Enh dummy cl" isn't self-evident (ahem), this is about making ophyd usable without pyepics installed, particular for doing stuff with `ophyd.sim` in the context of analysis or prototyping.

PR #523, opened 2018-03-23
FIX: fixing caproto integration with ophyd

mrakitin:
An attempt with @tacaswell to better integrate caproto with ophyd.
Work in progress...


mrakitin:
Thanks for merging, @ZLLentz!

PR #526, opened 2018-04-06
ENH: Individual Device logging

### LINKED ISSUE #525: Individual Device Logging
## Description
Putting a proposal out there for attaching individual loggers to Ophyd devices. This allows you to track all the log messages emitted by an individual instantiation of a device. In fact, if you are clever about naming the loggers with their parent names you can log messages all the way down to the signal level. This would be a huge help with debugging as you can watch a single device `DEBUG` mode during a scan without taking the full firehose of DEBUG logs.

This has precedent within the software stack as the `RunEngine` prints to its own `.log`

## Implementation
```python
class Device:
    def __init__(self, prefix, * , **kwargs):
        super().__init__(prefix, **kwargs)
        if self.parent:
               self.log = logging.getLogger(self.parent.log.name + '.' + self.name)
         else:
               self.log = logging.getLogger(self.__class__.__module__ + '.' + self.name)

# Setup logging
m  = ophyd.EpicsMotor('Tst:Mtr:1', name='test_motor')
logging = getLogger('ophyd') # Still gets everything it did prior
m.log.addHandler(StreamHandler(level=logging.DEBUG)) # Debug from only my test_motor
```

## Consequences
On the top level this is harmless as `ophyd.Device` will have `.log`  that people can choose to use or not. However, as part of this I would like to go through the whole library and have devices write to their own logs. This will be a large PR that touches a lot of files that I wanted to get approval for ahead of time.

Original discussion https://github.com/pcdshub/pcdsdevices/issues/193

teddyrendahl:
## Description
Reroute logging messages through loggers attached to instantiations of devices. This allows more detailed logging of individual device behaviors through complex scanning routines. Component devices use their `parent` log so that logic can be tracked through often complex nested component architecture. The highest level logger in the device will still start with `ophyd.` so `logging.getLogger('ophyd')` still gets all the log messages you received previously. 

I avoided changing any class that doesn't explicitly inherit from `OphydObject` so any log messages in the interface classes still go to module level logger. It would be nice to get these as well but I didn't want to break anything by assuming all implementations of the interface are on `OphydObj`. 

## Motivation
Closes #525

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1025,7 +1025,7 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
             except ExceptionBundle as ex:
                 exc_list.extend([('{}.{}'.format(attr, sub_attr), ex)
                                  for sub_attr, ex in ex.exceptions.items()])
-            except Exception:
+            except Exception as ex:
                 exc_list.append((attr, ex))
                 self.log.exception('Device %s (%s) stop failed', attr, dev)
 
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -563,7 +563,7 @@ class BlueskyInterface:
                                    "Maybe the most recent unstaging "
                                    "encountered an error before finishing. "
                                    "Try unstaging again.".format(self))
-        logger.debug("Staging %s", self.name)
+        self.log.debug("Staging %s", self.name)
         self._staged = Staged.partially
 
         # Resolve any stage_sigs keys given as strings: 'a.b' -> self.a.b
@@ -586,7 +586,7 @@ class BlueskyInterface:
         devices_staged = []
         try:
             for sig, val in stage_sigs.items():
-                logger.debug("Setting %s to %r (original value: %r)",
+                self.log.debug("Setting %s to %r (original value: %r)",
                              self.name,
                              val, original_vals[sig])
                 set_and_wait(sig, val)
@@ -601,7 +601,7 @@ class BlueskyInterface:
                     device.stage()
                     devices_staged.append(device)
         except Exception:
-            logger.debug("An exception was raised while staging %s or "
+            self.log.debug("An exception was raised while staging %s or "
                          "one of its children. Attempting to restore "
                          "original settings before re-raising the "
                          "exception.", self.name)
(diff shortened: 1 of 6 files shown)


ophyd/device.py
```diff
@@ -1027,7 +1027,7 @@ def stop(self, *, success=False):
                                  for sub_attr, ex in ex.exceptions.items()])
             except Exception as ex:
                 exc_list.append((attr, ex))
-                logger.error('Device %s (%s) stop failed', attr, dev,
+                self.log.error('Device %s (%s) stop failed', attr, dev,
```
klauer: Since you're touching these lines, mind switching `log.error(..., exc_info=ex)` -> `log.exception`?

ophyd/ophydobj.py
```diff
@@ -217,7 +222,7 @@ def inner(*args, **kwargs):
                     cb(*args, **kwargs)
                 except Exception as ex:
                     sub_type = kwargs['sub_type']
-                    logger.error('Subscription %s callback exception (%s)',
+                    self.log.error('Subscription %s callback exception (%s)',
```
klauer: Same

danielballan:
> I did notice the module in general was a little light on logging. I'll probably come back and add a few messages if I feel like there is some common sticking points.

Great. When we get the caproto 'control layer' working, we'll have to think about how to make that logging integrate nicely with this. That should improve the quality of error messages as well, but more can be done from there.

ophyd/ophydobj.py
```diff
@@ -60,6 +58,13 @@ def __init__(self, *, name=None, parent=None):
         self._args_cache = {k: None for k in self.subscriptions}
         # count of subscriptions we have handed out, used to give unique ids
         self._cb_count = count()
+        # Create logger name from parent or from module class
+        if self.parent:
+            base_log = self.parent.log.name
+        else:
+            base_log = self.__class__.__module__
+        # Instantiate logger
+        self.log = logging.getLogger(base_log + '.' + self.name)
```
tacaswell: Won't this lead to geometrically growing set of names as the child name is `'{parent.name}_{attr}'`?

tacaswell: :sheep: never mind, missed the `.log.` in the line where we set `base_log`.

PR #528, opened 2018-04-12
Added OEP for read_mode

jrmlhermitte:
WIP discussion

### DIFF
--- a/doc/source/OEP/OEP_002.rst
+++ b/doc/source/OEP/OEP_002.rst
@@ -54,13 +54,13 @@ For reading, one would just call each objects ``read(**kwargs)`` method:
     read_dict = dict()
     # first read the streaming devices
     for device in devices:
-        read_dict.update(cam1.read(streaming='always'))
+        read_dict.update(device.read(streaming='always'))
 
     # if there were any streaming devices, then re-read
     # devices who will supply streaming on demand
     if read_dict():
         for device in devices:
-            read_dict.update(motor1.read(streaming='ondemand'))
+            read_dict.update(device.read(streaming='ondemand'))
 
 
  - in bluesky (optional):
--- a/doc/source/OEP/OEP_002.rst
+++ b/doc/source/OEP/OEP_002.rst
@@ -1,5 +1,5 @@
 ===============================
-OEP 3 : Streaming Capability
+OEP 2 : Streaming Capability
 ===============================
 
 
--- a/doc/source/OEP/OEP_002.rst
+++ b/doc/source/OEP/OEP_002.rst
@@ -11,6 +11,7 @@ Status
 
 **Discussion**
 
+
 Branches and Pull requests
 ==========================
 
@@ -40,10 +41,26 @@ For example, let's imagine a case where we want to read streaming
 devices from a ``cam`` device and, if streaming devices present, also
 from a motor ``motor1``. The following information would be sufficient
 to perform such a scan:
+.. code-block:: python
 
     cam.roi1.read_mode = {'streaming' : 'never',}
     cam.hdf5.read_mode = {'streaming' : 'always',}
     motor1.read_mode = {'streaming' : 'ondemand',}
+
+For reading, one would just call each objects ``read()`` method:
+
+.. code-block:: python
+
+    read_dict = dict()
+    # first read the streaming devices
+    for device in devices:
+        read_dict.update(cam1.read(streaming='always'))
+
+    # if there were any streaming devices, then re-read
+    # devices who will supply streaming on demand
+    if read_dict():
+        for device in devices:
+            read_dict.update(motor1.read(streaming='ondemand'))
 
 
  - in bluesky (optional):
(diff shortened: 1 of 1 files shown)


tacaswell:
Can you also write up the change to the `read` method signature?

tacaswell:
I had a slightly different understanding from our conversation, more along the lines of

```python
may_stream = [d for d in devices 
              if d.read_mode.get('streaming', 'never') in ('always', 'ondemand')]
if any(d.read_mode['streaming'] == 'always'):
    for d in may_stream:
        yield from bps.read(d, read_mode='streaming')


```

jrmlhermitte:
Hmm yeah, thanks for the clarification. Here are my thoughts.

The pro of your suggestion is it removes the necessity of `read()` to handle arbitrary `kwargs`.

The con of that is it forces `read_mode` to be used in plan logic which may complicate bluesky (i.e. we now have to remember to first check `read_mode` before reading). I also worry about that usage as it might make it tricky to decide how to read from a device whose parent doesn't stream but child does. Do you have any nice idea on how to solve that?

tacaswell:
> I also worry about that usage as it might make it tricky to decide how to read from a device whose parent doesn't stream but child does

I do not think this is issue that we need to sort out automatically, it is up to the ophyd devices to sort this out. 

> The con of that is it forces read_mode to be used in plan logic which may complicate bluesky (i.e. we now have to remember to first check read_mode before reading).

In the case of streaming, we need to touch the plan logic _anyway_ to emit the extra event for streaming.  I think this simplifies bluesky as it only has to know about 1 kwarg ('read_mode') rather than many (streaming + what ever else we think of).

doc/source/OEP/OEP_002.rst
```diff
@@ -0,0 +1,239 @@
...
+any one of the three methods then it must accept it on all of them.
+The values accepted for ``read_mode`` should be strings.
+
+Objects much have a 'default' read mode so ``read()`` works.
+
+``trigger`` may dis-regard ``read_mode``, but must accept it if
+``read`` does.
+
+An object may require a matched ``trigger(read_mode=FOO)`` to go with
+a ``read(read_mode==FOO)``.
```
jrmlhermitte: `read_mode=FOO`

doc/source/OEP/OEP_002.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Changes to the RunEngine
+------------------------
+
+The descriptor cache will now need to be by keyed by ``(obj,
+msg.kwargs.get('read_mode', None))`` rather than just ``obj`` and the
```
jrmlhermitte: My initial feeling was to hash obj and `stream_name` in the descriptor cache. This is much better, I like it (the hash identifying the object should know *nothing* of the stream). Great catch!  :+1:

doc/source/OEP/OEP_002.rst
```diff
@@ -0,0 +1,239 @@
...
+descriptor building logic will have to be updated to account for both
+the object and read mode involved.  Reading the same device twice with
```
jrmlhermitte: "the descriptor building logic will have to be updated to account for both the object and read mode involved"
About implementing that. Do we want bluesky to know about `read_mode` or just pass kwargs (should we explicitly pass `read_mode` into the descriptor)? I would be tempted to say we should pass `read_kwargs` to the descriptor. What do you think?

doc/source/OEP/OEP_002.rst
```diff
@@ -0,0 +1,239 @@
...
+           yield Msg('checkpoint')
+           yield Msg('set', motor, step, group=grp)
+           yield from take_dark()
+           yield Msg('wait', None, group=grp)
+
+       yield from move()
+       return (yield from trigger_and_read(list(detectors) + [motor]))
+
+
+The ``is_dark_stale`` function is doing alot of work, we may need to
```
jrmlhermitte: Great example. As you suggested before in a previous conversation, read_mode can also be a function.
```
get_dark = det.getattr('read_mode', {}).get('dark', lambda : False)
if get_dark():
    # take dark
```
where `get_dark` could contain some of the logic.
This would at least move the function from some global function to an object member, which is safer. What do you think?

doc/source/OEP/OEP_002.rst
tacaswell: I can see arguments either way.  Currently the `_read` coro passes all the args and kwargs into `obj.read` in case we need to, on the fly, write a deeply coupled custom device and plan we can.

If we put _all_ of the args/kwargs in the hash key we may be cutting off some flexibility (ex, if you have weird device that needs some kwarg passed to it that is different every time).  If every key went in then you would end up with a different descriptor every time which is not what we want.  But, the next time we want to add a 'standard' kwarg to 'read' and friends we would not have to change any of the descriptor logic.

If we put only read_mode we keep the flexibility to do really ad-hoc custom stuff if needed, but the next time we want to add a new 'standard' kwarg we will have to make bigger changes.

I was primarily thinking of the logic required to determine if for the set of objects read in this event if we need to construct a new descriptor or if we have a cached one that works or we need to fail, but these things should also be put into the descriptor document.  Any suggestions as to where?

I lean towards making 'read_mode' special to preserve the 'do something really wacky' flexibility.

doc/source/OEP/OEP_002.rst
tacaswell: I don't remember saying that, but it is a good idea!

Slightly concerned about the wildly varying semantics on the keys....

doc/source/OEP/OEP_002.rst
jrmlhermitte: That's a good point and something we should discuss first: how to know when a generated descriptor is stale.

As far as I understand, this is done by sending the `configure` message. Since the logic supplying the `kwargs` for read is at the same level as the logic for flagging a descriptor as stale, is this enough? I would be more tempted to say that such flagging should remain on the bluesky side, and the run engine shouldn't handle this.

doc/source/OEP/OEP_002.rst
jrmlhermitte: you mentioned it when trying to emphasize the power of having a dictionary of keys pointing to arbitrary object (as opposed to `bool`s), that they could also be functions :-D

klauer:
Is this still something under consideration?

danielballan:
I think it's worth preserving this until we get a chance to make a new issue or new PR. The are some ideas and discussion here that we don't want to lose.

tacaswell:
Yes, this I think this is something we still need.

danielballan:
To be continued in https://github.com/bluesky/bluesky-enhancement-proposals/pull/7

PR #531, opened 2018-04-19
ENH: Flexible Alarm Handling in EpicsMotor

### LINKED ISSUE #530: EpicsMotor error level is unconfigurable
## Description
After a move, `EpicsMotor` checks the alarm state of the motor and raises an exception if there is an issue. Unfortunately at SLAC we have a had a few instances of motors being set up in haste and throwing minor errors on a high percentage of moves. I also think our implementation of the motor record might be a little more liberal with generating alarms. This means they have a difficult time being used in `bluesky` routines as one failed move (even when the motor reached the target and will accept another move), brings the scan to a grinding halt. The only work around is to subclass `EpicsMotor` and either catch the exception or duplicate the logic but tolerating a `MINOR` alarm state.

 https://github.com/NSLS-II/ophyd/blob/d170de078d229650a61a8f6aca81ecb68e1ff094/ophyd/epics_motor.py#L255

## Proposed Solution
Allow this variable to be configurable so that adjusting a constant doesn't requires creating a subclass. Also noting that I'm not in love with this solution. We'll probably implement something at SLAC but if you do not have issues and do not want to sully your `EpicsMotor` implementation I understand.

```python
class EpicsMotor(PositionerBase, Device)
    ...
    tolerated_severity = AlarmSeverity.NO_ALARM
    ...
    
        if severity != AlarmSeverity.NO_ALARM:
            if severity > self.tolerated_severity:
                self.log.error(...)
                success = False
            else:
                self.log.warning(...)
```

teddyrendahl:
## Description
Compare the alarm state at the conclusion of a move against the `tolerated_alarm` instead of a fixed `AlarmSeverity.NO_ALARM`. The operator will always be warned of an alarm but this change just modifies whether or not the alarm is of a high enough level to mark the move as failed. This allows alarm conditions to be ignored that do not affect the motor's ability to scan.

## Motivation
Closes #530

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -70,6 +70,9 @@ class EpicsMotor(Device, PositionerBase):
     home_forward = Cpt(EpicsSignal, '.HOMF')
     home_reverse = Cpt(EpicsSignal, '.HOMR')
 
+    # alarm information
+    tolerated_alarm = AlarmSeverity.NO_ALARM
+
     def __init__(self, *args, **kwargs):
 
         self._hints = None
@@ -250,15 +253,22 @@ class EpicsMotor(Device, PositionerBase):
                 if self.high_limit_switch.get() == 1:
                     success = False
 
+            # Check the severity of the alarm field after motion is complete.
+            # If there is any alarm at all warn the user, and if the alarm is
+            # greater than what is tolerated, mark the move as unsuccessful
             severity = self.user_readback.alarm_severity
 
             if severity != AlarmSeverity.NO_ALARM:
                 status = self.user_readback.alarm_status
-                self.log.error('Motion failed: %s is in an alarm state '
-                               'status=%s severity=%s',
-                               self.name, status, severity)
-                success = False
-
+                if severity > self.tolerated_alarm:
+                    self.log.error('Motion failed: %s is in an alarm state '
+                                   'status=%s severity=%s',
+                                   self.name, status, severity)
+                    success = False
+                else:
+                    self.log.warning('Motor %s raised an alarm during motion '
+                                     'status=%s severity %s',
+                                     self.name, status, severity)
             self._done_moving(success=success, timestamp=timestamp,
                               value=value)
 
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -236,7 +236,7 @@ class EpicsMotor(Device, PositionerBase):
             started = self._started_moving = (not was_moving and self._moving)
 
         self.log.debug('[ts=%s] %s moving: %s (value=%s)', fmt_time(timestamp),
-                     self, self._moving, value)
+                       self, self._moving, value)
 
         if started:
             self._run_subs(sub_type=self.SUB_START, timestamp=timestamp,


danielballan:
LGTM. Is it worth making this an argument to `__init__` as well?

Pros:
* symmetry/consistency with `timeout` and `settle_time`, which are both arguments to `__init__` at settable attributes
* increased discoverability (it will show up in `EpicsMotor?`)

Cons:
* argument bloat

teddyrendahl:
@danielballan I considered it. Figured I'd take the temperature of the room before adding keywords.

danielballan:
I guess it partly comes down to how much ophyd should let the user work around flaky hardware, a common point of discussion around here. Does ophyd force users to fix problems at a lower level or leave it up to them to decide when to apply a hack-ish solution like ignoring minor alarms?

teddyrendahl:
I am fine with leaving it as an attribute but not a keyword. Things like `timeout` and `settle_time` come down to taste, and I could see why you would want different values for the same motor in different applications.

Motor alarms seems like something that should be slightly more hidden. I am already worried about a user ignoring all alarms by setting the value to `AlarmSeverity.MAJOR_ALARM`. The main motivation for this PR was just that it seemed like a lot of machinery needed to be repeated for an expert to change this, not so much that we want novice beamline operators playing with it.

klauer:
I'm not necessarily opposed to this, but would it not be better as a subclass of `EpicsMotor`? The more modifications we make to `EpicsMotor` to make it fit mutant `EpicsMotor`-like records, the the more confusing the original gets.

So:
1. Has anyone had alarm-related issues with the real motor record?
2. For/against: `TolerantEpicsMotor` (derived from `EpicsMotor`) with a severity kwarg instead?

tacaswell:
I could be sold on making this a private flag and then providing a sub-class where you can control it an init (to help move configuration out of scripts).

ZLLentz:
I'm against the idea of a subclass for features like this because it scales poorly for the user as we add more features. I don't want to sift through `TolerantEpicsMotor`, `SettlingEpicsMotor`, etc. and decide which set of them to inherit from as I implement my `EpicsMotor` subclass that needs some combination of standard ophyd extra features.

I think concern about mutating the `EpicsMotor` class for weird motor-record-like records is valid, and if this is the case then PRs like this should be about refactoring `EpicsMotor` so that pieces of the functionality can be easily overridden. This PR is a result of it being awkward to override this behavior in a subclass as written, and an expectation that other users are probably having a similar problem.

ZLLentz:
@klauer I don't know if they are, but the expectation is where this PR comes from.

danielballan:
I think @ZLLentz makes a good point. Before we merge (hopefully today!) one last thought -- should this be configurable per-motor, an instance attribute rather than a class attribute? I would imagine you would apply this to select motors with known issues, not globally to all motors.

teddyrendahl:
I thought keeping it as a class attribute kept it generic, allowing either mode of operation. Maybe this isn't desirable.

```python
mot = EpicsMotor(...)
mot.tolerated_alarm = ...   # Only for a single instance (How most users will set this I'm guessing)
EpicsMotor.tolerated_alarm = ...  # All motors who don't have tolerated alarm explicitly set
```

danielballan:
Ah, I was confused: I had it my head that the instances would be _coupled_. But of course that is only a danger when the value of the class attribute is _mutable_ (e.g. `_default_read_attrs = []`) which is not the case here.

Merging! Thanks for sharing this work and especially thanks for entertaining our woah-woah-lets-consider-many-possible-alternatives review process. :- )

teddyrendahl:
😆 No worries. I think this one really toed the line between something the class should be able to handle and something we should just demand that lower level software does a better job on. Certainly worth discussing.

Issue #428, opened 2017-08-19
Update usage of networkx to work with their 2.0 release


danielballan:
I'm not sure how hard it is to support both versions of networkx. Maybe we can require 2.0 for the 1.0 release of ophyd.

http://networkx.readthedocs.io/en/latest/release/migration_guide_from_1.x_to_2.0.html

jrmlhermitte:
The [changes](https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html) are not that bad. I only see three lines of code that would be affected in `areadetector/base.py`: [here](https://github.com/NSLS-II/ophyd/blob/master/ophyd/areadetector/base.py#L229), [here](https://github.com/NSLS-II/ophyd/blob/master/ophyd/areadetector/base.py#L252) and [here](https://github.com/NSLS-II/ophyd/blob/master/ophyd/areadetector/base.py#L267).

As far as I can tell, these look okay since the main changes are a switch from lists to views for nodes (so updating a graph after creating the view still contains the updated graph). I don't see any need to change the code.

If that's right, then I think we just need to:
- update [lightsource recipes](https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/ophyd/meta.yaml#L21)
- update [ophyd requirements](https://github.com/NSLS-II/ophyd/blob/master/requirements.txt#L1)
- update [.travis.yml](https://github.com/NSLS-II/ophyd/blob/master/.travis.yml)

Does that sound about right?

tacaswell:
If we aren't pinning does that mean we are supporting networkx2 already?

tacaswell:
or put another way, what version of networkx do we install on travis currently?

CJ-Wright:
From your most recent commit on master you are running against networkx 2.1 https://travis-ci.org/NSLS-II/ophyd/jobs/374059268#L2004

jrmlhermitte:
closed by #533

PR #533, opened 2018-05-03
first test travis

jrmlhermitte:
Don't merge. We will add strict dependence on networkx 2.0.

First, I have added `conda list` so we know what we're running on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,6 +93,8 @@ script:
   - python -c "import epics; print(epics.caget('XF:31IDA-OP{Tbl-Ax:X1}Mtr'))"
 
 
+  # list packages
+  - conda list
   # running tests
   - python run_tests.py --cov=ophyd --cov-report term-missing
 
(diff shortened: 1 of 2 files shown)


jrmlhermitte:
for issue #428 
ready to merge

tacaswell:
I am ok with merging this, but if we currently don't use any API that is not supported by both, why pin the version?

jrmlhermitte:
This is my misunderstanding from #428. I thought we'd just switch to 2.0.

If we don't, should we keep testing for both versions, so that we're aware of it the day it breaks?
Doing so could be tricky if any dependent library requires 2.0.

I'd vote we don't merge this, and don't add a test for 2.0. Does that sound fair?

tacaswell:
@jrmlhermitte Please open PRs from your fork rather than working on upstream branches.

klauer:
Now that there are many 1000s of commits following the standard pattern of "TAG: message", why are PRs being merged that do not follow this format?

jrmlhermitte:
thanks. ill use my branch from now on.
thanks @klauer for the feedback. will try to follow the convention

Issue #527, opened 2018-04-08
QuadEM currentAll


danielballan:
From an email from Lin Yang:

> Could you add currentAll as one of the attributes of the quadem class, when you update Bluesky/Ophyd? The PV does exist.

We should look at the QuadEM variant used at ESM and BMM to make sure this hasn't already been done there.

jrmlhermitte:
ESM [here](https://github.com/NSLS-II-ESM/profile_collection/blob/master/startup/30-detectors.py)
BMM [here](https://github.com/NSLS-II-BMM/profile_collection/blob/master/startup/40-quadem.py)

doesn't look either has done this? I think PR in #537 would do it, but it might break if this PV doesn't exist

We could remove the attr after the fact if that's the case. So what's your opinion on this?

jrmlhermitte:
closed by #537

PR #537, opened 2018-05-10
ENH: Add sumAll to QuadEM

jrmlhermitte:
From #527

### DIFF
--- a/ophyd/quadem.py
+++ b/ophyd/quadem.py
@@ -87,6 +87,8 @@ class QuadEM(SingleTrigger, DetectorBase):
     current3 = ADCpt(StatsPlugin, 'Current3:')
     current4 = ADCpt(StatsPlugin, 'Current4:')
 
+    sumAll = ADCpt(StatsPlugin, 'SumAll:')
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


jrmlhermitte:
Good question. I assumed this was a typo because I didn't see sumAll. I'll ask Lin to double check.
[Here](https://github.com/NSLS-II-LIX/profile_collection/blob/master/startup/20-bpm.py#L34) is the usage.

mrakitin:
I see, thanks. The tests are still failing, probably `read_attrs` of `FakeQuadEM` should be updated with `sumAll`.

jrmlhermitte:
Lin confirmed that it should be sumAll and he likely made a mistake.
ill check soon, thanks!

jrmlhermitte:
I think we have to override the `FakeQuamEM` with a fake version of `sumAll`. waiting on Travis...

tacaswell:
You probably need to make our fake QuadEM smarter to get the tests to pass.

jrmlhermitte:
passes now but py3.6 fails:
```

E           TimeoutError: Server at ('172.17.0.2', 5064) did not respond to attempt to read channel named 'XF:31IDA-OP{Tbl-Ax:X1}Mtr.VAL' within 1.9999873499999694-second timeout.
```

I think this is unrelated. Is this good to go?

jrmlhermitte:
thanks @tacaswell !

PR #538, opened 2018-05-10
ENH: Added labels to ophyd devices

jrmlhermitte:
WIP: From this issue [here](https://github.com/NSLS-II/bluesky/pull/1020).

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -33,11 +33,11 @@ class OphydObject:
 
     _default_sub = None
 
-    def __init__(self, *, name=None, parent=None, device_labels=None):
+    def __init__(self, *, name=None, parent=None, labels=None):
         super().__init__()
-        if device_labels is None:
-            device_labels = []
-        self._ophyd_labels_ = device_labels
+        if labels is None:
+            labels = set()
+        self._ophyd_labels_ = set(labels)
 
         # base name and ref to parent, these go with properties
         if name is None:
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -34,10 +34,11 @@ class OphydObject:
     _default_sub = None
 
     def __init__(self, *, name=None, parent=None, labels=None):
-        super().__init__()
         if labels is None:
             labels = set()
         self._ophyd_labels_ = set(labels)
+
+        super().__init__()
 
         # base name and ref to parent, these go with properties
         if name is None:
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -33,8 +33,11 @@ class OphydObject:
 
     _default_sub = None
 
-    def __init__(self, *, name=None, parent=None):
+    def __init__(self, *, name=None, parent=None, device_labels=None):
         super().__init__()
+        if device_labels is None:
+            device_labels = []
+        self._ophyd_labels_ = device_labels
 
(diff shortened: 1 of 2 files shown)


ophyd/ophydobj.py
```diff
@@ -33,8 +33,11 @@ class OphydObject:
 
     _default_sub = None
 
-    def __init__(self, *, name=None, parent=None):
+    def __init__(self, *, name=None, parent=None, labels=None):
         super().__init__()
+        if labels is None:
```
tacaswell: I would move this above the super call so we can expect this to be available.

ophyd/ophydobj.py
```diff
@@ -33,11 +33,11 @@ class OphydObject:
 
     _default_sub = None
 
-    def __init__(self, *, name=None, parent=None, device_labels=None):
+    def __init__(self, *, name=None, parent=None, labels=None):
```
mrakitin: The changes here are opposite than the statement in the commit title. I'm a bit confused - do we eventually prefer `labels` over `device_labels`?

ophyd/ophydobj.py
jrmlhermitte: good idea thanks!

ophyd/ophydobj.py
jrmlhermitte: oops yes wrote too fast, went from `device_labels` to `lablels` will edit commit
thanks!

jrmlhermitte:
The failures on Travis don't seem to be specifically related to this. It only fails py3.6 and nightly:

```
>       self.assertEqual(m.position, 0.1)
>       assert a == b
E       assert 0.099999999999999645 == 0.1
ophyd/tests/conftest.py:314: AssertionError
```

and
```

```

jrmlhermitte:
@tacaswell do we have final blessing?
@klauer as well?

klauer:
Overall 👍 

Is there a reason not to keep the attribute easily readable by the user? If not just as a public attribute, I'd think there should at least be a property. (These can be likened to hints, which are easily accessible by the user.)

tacaswell:
We were channelling the IPython style `_repr_html_` naming scheme to indicate that this is part of the framework (which is something we have done no where else yet).

danielballan:
Yeah. Two things:

1. I'm not opposed to a property that reflects this up as `.label` in a future release. Keeping it private for now makes it easier to undo this if it's has bad or unintended consequences.
2. This is different from other public bluesky/ophyd API because we're using it in the context of scraping a namespace _across libraries_. Underscore-named interfaces are one established way to do this, such as the way pandas talks IPython (and other consumers) via `_repr_html_`. A public attribute (`label`) or a private one with an unscoped name (`_label`) risks scooping unrelated objects with the same name that aren't trying to participate in this interface. And like `_repr_html_`, its intended use is UI/display-oriented.

Anyway, I think we should consider this an experimental feature.

danielballan:
Taking in the discussion on related issues, I am hearing that people are generally buying that `_ophyd_labels_` is an acceptable experiment. I don't think we've established yet whether this information should be reflected in another, public attribute, and if so where. We have room to continue to discuss that while we try out this (via  https://github.com/NSLS-II/bluesky/pull/1020, https://github.com/NSLS-II/bluesky/pull/1031) at the beamlines and see how it goes.

But the related discussion around `*_attrs` prompted by #539 is very interesting and needs further discussion.

PR #539, opened 2018-05-10
ENH: move hints to Device

### LINKED ISSUE #487: Make hints ubiquitious
## Motivation
Since the `hints` feature looks like it is here to stay, I would like to add an empty implementation to `ophyd.Device`.  Then subclasses of Device would only need to fill a `_hints` class attribute instead of rewriting boilerplate property code. 

## Implementation
It looks like the standard structure of the hints is:
```python
def hints(self):
      return {'fields': [...]}
```
I think that the default can be an empty dictionary if the subclass does not reimplement `_hints`. This way you can check with just `if device.hints`, but I do see the benefit of defining the structure in the base implementation so it would look something like this:
```python
_hints = {'fields' : list()}
```
Just curious, is there anything else we use in the `hints` dictionary besides "fields"? 

## Implications
Any code that doesn't check for an empty dictionary explicitly but calls `hasattr(device, 'hints')` might run into issues.

### LINKED ISSUE #502: Add hints to Device
@danielballan @mrakitin 

> Long-term fix: move `hints` into the `Device` class so that this doesn’t keep happening, as we discussed yesterday at ISR, with a property, a setter, and a class attribute `_default_hints`. Short-term fix: copy the pattern from ISR.

Current fix at ISR is [here](https://github.com/NSLS-II-ISR/isr_profiles/blob/master/profile_collection/startup/20-area-detectors.py#L105). This should be eventually moved to `Device` class
  
Sample usage that needs this feature (at SMI)
https://github.com/NSLS-II-SMI/ipython_ophyd/blob/master/profile_collection/startup/11-energy.py#L184

mrakitin:
Resolves #502 
Resolves #487

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1015,7 +1015,7 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
         else:
             return {'fields': [getattr(self, component_name).name
                                for component_name in
-                                   self._default_hints['fields']]}
+                                   self._default_hints.get('fields', [])]}
 
     @hints.setter
     def hints(self, val):
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -706,6 +706,8 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     configuration_attrs : sequence of attribute names
         the components to be read less often (i.e., in
         ``read_configuration()``) and to adjust via ``configure()``
+    hints : dict, optional
+        the information used to help Bluesky to provide best effort feedback
     parent : instance or None
         The instance of the parent device, if applicable
     """
@@ -719,9 +721,11 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     _default_read_attrs = None
     # If `None`, defaults to `[]`
     _default_configuration_attrs = None
+    # If `None`, defaults to `{'fields': []}`
+    _default_hints = None
 
     def __init__(self, prefix='', *, name,
-                 read_attrs=None, configuration_attrs=None,
+                 read_attrs=None, configuration_attrs=None, hints=None,
                  parent=None, **kwargs):
         # Store EpicsSignal objects (only created once they are accessed)
         self._signals = {}
@@ -746,6 +750,7 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
 
         self.read_attrs = list(read_attrs)
         self.configuration_attrs = list(configuration_attrs)
+        self.hints = hints
 
         # Instantiate non-lazy signals
         [getattr(self, attr) for attr, cpt in self._sig_attrs.items()
@@ -967,6 +972,24 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
             with the ``event_model.event_descriptor.data_key`` schema.
         """
         return self._describe_attr_list(self.configuration_attrs, config=True)
+
+    @property
+    def hints(self):
+        if self._hints is not None:
+            return self._hints
+        else:
+            return {'fields': [getattr(self, component_name).name
+                               for component_name in self._default_hints]}
+
+    @hints.setter
+    def hints(self, val):
+        if val is not None:
+            read_keys = list(self.describe())
+            for key in val.get('fields', []):
+                if key not in read_keys:
+                    raise ValueError("{} is not allowed -- must be one of {}"
+                                     .format(key, read_keys))
+        self._hints = val
 
     @property
     def trigger_signals(self):
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -752,7 +752,13 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
 
         self.read_attrs = list(read_attrs)
         self.configuration_attrs = list(configuration_attrs)
-        self.hints = hints
+        self._hints = None
+        # Some subclasses override the hints property, rendering it
+        # un-settable. (This is probably why you shouldn't override properties
+        # in a subclass, but it happens....) For this reason, we only attempt
+        # to set hints if the user passed in something other than None.
+        if hints is not None:
+            self.hints = hints
 
         # Instantiate non-lazy signals
         [getattr(self, attr) for attr, cpt in self._sig_attrs.items()
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -701,14 +701,16 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
         The PV prefix for all components of the device
     name : str, keyword only
         The name of the device
-    read_attrs : sequence of attribute names
+    read_attrs : sequence of attribute names, optional
         the components to include in a normal reading (i.e., in ``read()``)
-    configuration_attrs : sequence of attribute names
+    configuration_attrs : sequence of attribute names, optional
         the components to be read less often (i.e., in
         ``read_configuration()``) and to adjust via ``configure()``
-    hints : dict, optional
-        the information used to help Bluesky to provide best effort feedback
-    parent : instance or None
+    hints : dict or None, optional
+        May be used to help downstream consumers infer interesting keys.
+        Example: ``{'fields': ['motor_readback']}``. If `None`, a default
+        is derived from the class attribute `_default_hints`
+    parent : instance or None, optional
         The instance of the parent device, if applicable
     """
 
@@ -721,7 +723,7 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     _default_read_attrs = None
     # If `None`, defaults to `[]`
     _default_configuration_attrs = None
-    # If `None`, defaults to `{'fields': []}`
+    # If `None`, defaults to `{}`
     _default_hints = None
 
     def __init__(self, prefix='', *, name,
@@ -977,6 +979,8 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     def hints(self):
         if self._hints is not None:
             return self._hints
+        elif self._default_hints is None:
+            return {}
         else:
             return {'fields': [getattr(self, component_name).name
                                for component_name in
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -979,7 +979,8 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
             return self._hints
         else:
             return {'fields': [getattr(self, component_name).name
-                               for component_name in self._default_hints]}
+                               for component_name in
+                                   self._default_hints['fields']]}
 
     @hints.setter
     def hints(self, val):
@@ -989,6 +990,8 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
                 if key not in read_keys:
                     raise ValueError("{} is not allowed -- must be one of {}"
                                      .format(key, read_keys))
+            val = dict(val)
+
         self._hints = val
 
     @property
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -725,6 +725,31 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     _default_configuration_attrs = None
     # If `None`, defaults to `{}`
     _default_hints = None
+    # WARNING -- POTENTIALLY CONFUSING DETAIL:
+    # Given a Device subclass like:
+    #
+    # class Thing(Device):
+    #     a = Component(Signal)
+    #     b = Component(Signal)
+    #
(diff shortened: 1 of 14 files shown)


ophyd/device.py
```diff
@@ -719,9 +721,11 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     _default_read_attrs = None
     # If `None`, defaults to `[]`
     _default_configuration_attrs = None
+    # If `None`, defaults to `{'fields': []}`
+    _default_hints = None
 
     def __init__(self, prefix='', *, name,
-                 read_attrs=None, configuration_attrs=None,
+                 read_attrs=None, configuration_attrs=None, hints=None,
```
danielballan: The docstring needs an entry for Parameters. Suggestion:

```
hints: dict or None
    May be used to help downstream consumers infer interesting keys.
    Example: ``{'fields': ['motor_readback']}``. If `None`, a default
    is derived from the class attribute `_default_hints`.
```

ophyd/epics_motor.py
```diff
@@ -281,64 +281,13 @@ def report(self):
...
-        self._hints = val if val is None else dict(val)
-
 
 class MotorBundle(Device):
     """Sub-class this to device a bundle of motors
 
     This provides better default behavior for ``hints``,
     ``read_attrs`` and ``configuration_attrs``
     """
-    _hints = None
```
danielballan: MotorBundle merges the hints of its child motors. This behavior is still desirable and different from the behavior that has been moved up to `Device`. This block should not be deleted.

ophyd/positioner.py
```diff
@@ -355,10 +355,6 @@ def read(self):
                         'timestamp': time.time()}
         return d
 
-    @property
```
danielballan: This, like `Signal` has a special value (enshrined as `self.position`). Therefore, I think it should keep this hint.

ophyd/pseudopos.py
```diff
@@ -675,14 +665,6 @@ def real_position(self):
         '''Real motor position namedtuple'''
         return self.RealPosition(*self._real_cur_pos.values())
 
-    @property
```
danielballan: I think this should be saved. Pseudopositioners are potentially multi-dimensional, with a dynamic number of dimensions, and this handles that specially.

ophyd/device.py
```diff
@@ -968,6 +973,27 @@ def describe_configuration(self) -> OrderedDictType[str, Dict[str, Any]]:
         """
         return self._describe_attr_list(self.configuration_attrs, config=True)
 
+    @property
+    def hints(self):
+        if self._hints is not None:
+            return self._hints
+        else:
+            return {'fields': [getattr(self, component_name).name
```
tacaswell: This needs an extra cut-out if `_default_hints is None`

ophyd/device.py
```diff
@@ -719,9 +721,11 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
     _default_read_attrs = None
     # If `None`, defaults to `[]`
     _default_configuration_attrs = None
+    # If `None`, defaults to `{'fields': []}`
```
tacaswell: I thought we agreed that the default behaviour would be `{}` in the case of no hints?

ophyd/quadem.py
```diff
@@ -22,6 +22,8 @@ class QuadEM(SingleTrigger, DetectorBase):
     _default_configuration_attrs = ('integration_time', 'averaging_time')
     _default_read_attrs = ('current1.mean_value', 'current2.mean_value',
                            'current3.mean_value', 'current4.mean_value')
+    _default_hints = {'fields': ['current1', 'current2',
```
danielballan: Looking at the deleted block, I see that the original default hints had the form `self.current1.mean_value.name`. Therefore, these entries should have the form `'current1.mean_value'`.

ophyd/scaler.py
```diff
@@ -160,5 +161,3 @@ def select_channels(self, chan_names):
                                    "{}".format(ch, tuple(name_map)))
         self.channels.read_attrs = list(read_attrs)
         self.channels.configuration_attrs = list(read_attrs)
-        self.hints = {'fields': [getattr(self.channels, ch).s.name
```
danielballan: This is even trickier than the case above. I think that instead of delete this, it's probably easier to change `self.hints` -> `self._hints` and treat this as if it's a user-provided value.

ophyd/sim.py
```diff
@@ -649,19 +638,6 @@ def __init__(self, *args, reg=DO_NOT_USE, save_path=None, save_func=np.save,
         else:
             self.reg = None
 
-    @property
-    def hints(self):
-        if self._hints is None:
-            # Since data is external, hint that it should not be printed or
-            # plotted. Relax this when LiveTable etc. get smarter about
```
danielballan: Per this comment, I _think_ LiveTable and friends are now smart enough to deal with this.

ZLLentz:
Related question: why is the `hints` attribute always a `dict` with a key `"fields"`? I'd naively assume that we should be providing this as a simple list, or at least providing `_default_hints` as a simple list.

danielballan:
@ZLLentz Fair question. We are providing for the possibility that we will want to hint about other things in the future. On the bluesky side (in plans) we have accumulated a couple different fields, and I think in time Devices will grow more, too. It's a bit of extra scoping, but it seems better than accidentally painting ourselves into a corner, or growing a bunch more attributes (`Device.hints_about_another_thing_besides_fields`). Reasonable?

ophyd/device.py
```diff
@@ -968,6 +1006,29 @@ def describe_configuration(self) -> OrderedDictType[str, Dict[str, Any]]:
...
+    @property
+    def hints(self):
+        if self._hints is not None:
+            return self._hints
+        elif self._default_hints is None:
+            return {}
+        else:
+            return {'fields': [getattr(self, component_name).name
+                               for component_name in
+                                   self._default_hints['fields']]}
```
tacaswell: This probably should be `self._default_hints.get('fields', [])` to be extra paranoid.

tacaswell: do we want to do the read_attrs filtering / validation here as well?

ophyd/sim.py
```diff
@@ -768,9 +744,7 @@ class ABDetector(Device):
     def trigger(self):
         return self.a.trigger() & self.b.trigger()
 
-    @property
-    def hints(self):
-        return {'fields': [self.a.name]}
+    _default_hints = {'fields': ['a']}
```
tacaswell: minor style point: can you move this up to the top of the class definition?

ophyd/sim.py
```diff
@@ -793,9 +767,7 @@ def __init__(self, *args, **kwargs):
     def trigger(self):
         return self.a.trigger() & self.b.trigger()
 
-    @property
-    def hints(self):
-        return {'fields': [self.a.name, self.b.name]}
+    _default_hints = {'fields': ['a', 'b']}
```
tacaswell: same as above, move to top.

teddyrendahl:
Just playing devil's advocate here. Why if `hints` is supposed to be general and someday flushed out into a more rich feature set, are we adding a second place to put this stuff called `labels`? Shouldn't it be `hints : {'fields': [], 'labels': []}`? This also allows the possibility for the labels to be used nicely in `bluesky.callbacks` e.t.c

tacaswell:
One last minute wrench: do we always want the user interface to be setting the name of the components to use (as with `_default_hints`, `read_attrs`, and `configuration_attrs`)?  It would break the the property round-trip nature, but would bring the API more in line with our existing APIs and deal with the following evil examples:

```python
class Dongle(Device):
     a = Cpt(Signal)
     b = Cpt(Signal)

dev = Dongle(name='dev')
dev.hints = {'fields': ['dev_a']}
dev.a.name = 'bob'

# or

dev.read_attrs = ['b']
```

danielballan:
Then again, to play my own "devil's advocate"...

This inconsistency is a bit of trap. It starts off OK...

```python
class Thing(Device):
    _default_read_attrs = ['a', 'b']
    _default_hints = {'fields': ['a']}
    a = Component(Signal)
    b = Component(Signal)
```

But this might be surprising:

```python
thing = Thing(name='thing')
thing.read_attrs = ['b']  # cool cool cool
thing.read().keys() == ['thing_b']   # just a reminder about how this looks
thing.hints = {'fields': ['b']} # ValueError!
thing.hints = {'fields': ['thing_'b']}  # correct
```

Would it be better to do:

```python
class Thing(Device):
    _default_read_attrs = ['a', 'b']
    _default_hints_fields_attrs = ['a']
    a = Component(Signal)
    b = Component(Signal)

thing = Thing(name='thing')
thing.hints_fields_attrs = ['b']
thing.hints  # read-only property, equal to {'fields': ['thing_b']}
```

That does seems more self-consistent, if more complex in implementation. That option occurred to me during this development process, but until now I didn't fully realize its advantages. Thoughts?

tacaswell:
the biggest downside is the profferation of `thing.*_attrs` 'magic' attributes.  We should think about how to get those under control...

Would have to support the `child.child.child` implicit nesting we support else where as well.

I am leaning towards a long-ish term move to `_ophyd_attrs_` + validating none-underscored setters + a better summary method?

danielballan:
Oh, due to the fickle nature of websockets, I did not see the last three comments before posting mine. Anyway it seems like we are thinking about similar lines.

@tacaswell I can't quite tell if what you are suggesting is the same as what I am suggesting above. Is it?

@ZLLentz Any thoughts on whether the suggestion in my previous comment would make things better or worse?

@teddyrendahl 

> Shouldn't it be hints : {'fields': [], 'labels': []}?

That does look nice, and I'm glad it was raised as possibility because I admit I hadn't considered it. I still feel that the [analogy to interfaces I argued in this comment](https://github.com/NSLS-II/ophyd/pull/538#issuecomment-388196961) makes labels special. That said...

> This also allows the possibility for the labels to be used nicely in bluesky.callbacks e.t.c

No matter what we call labels, we can teach the RunEngine to capture them (if they exist) and put them in descriptors, just as we now do with hints and object names. If we want, we could put the labels list in _both_ places and enforce that they be synced by virtue of `hints` being a property.

tacaswell:
@danielballan Yes



> That does look nice, and I'm glad it was raised as possibility because I admit I hadn't considered it. I still feel that the analogy to interfaces I argued in this comment makes labels special. That said...

They feel different to me too.

Agree we can always teach the RE to capture more things.

ophyd/sim.py
mrakitin: Sure, done for this case, and for a few other similar cases.

ophyd/device.py
mrakitin: Done for the first comment. Which king of filtering would you suggest?

teddyrendahl:
@danielballan Does that mean we were independently devil's advocates? 👿 

I like the analogy to interfaces as well, I think `labels` will be super interesting to use in conjunction with `typhon`, e.t.c. I do however not really see why these are separate from `"hints"` which seem like a specification for interfaces as well. Hard to say without talking to you guys in person about the direction you see these going, just curious why they need a separate stream when you seem to have left `"hints"` intentionally ambiguous to capture this sort of thing.  Is it that `"hints"` are specifically for `bluesky.callbacks` and 'labels' are for the world at large?

I don't doubt your ability to capture these in the `RunEngine`, I guess similar to the issue of proliferation of `_attr` class attribute, it is a burden to remember to properly configure too many different lists and dictionaries. That is the major reason I'd favor a consolidation.

One suggestion that is actually more relevant to this PR, is it would be nice to pick these up automatically in the Device.hints implementation as @danielballan suggested. 

```python
def hints(self):
     ....
     return {'labels': self._labels or [], ...}
```

ZLLentz:
My assumptions:
- `hints` are intended as a summary to inform other code about the most important details of devices and plans.
- `labels` are intended as a way for users to apply "tags" to their devices, which can be used by other code to group them together, etc.

If this is true, the two fulfill separate purposes and their implementations seem fine. If there is overlap, then perhaps this isn't true and `labels` should just be a special category of `hints`.

The asymmetry trap you pointed out is a problem, and your solution is better than the downsides because we should avoid confusing the user. It seems like the `hints['fields']` is a special case of hints that should be symmetrical with the other `_default_*_attrs` implementations, so the user can minipulate it in a naive way without breaking the class and provide reasonable subclass defaults without overriding `hints`.

I think the name is a bit long, I suggest `_default_hints_attrs` or `_default_hints_fields`. Then the user can manipulate `thing.hints_fields` and `thing.hints` would update with the proper names.

I expect we won't find ourself doing too many of these for `Device`, so I'm not worried about the proliferation. We've covered all reasonable catagories of attrs at least. Perhaps `_default_hints_methods` could be useful at some point (for autogenerated GUI), but I can't think of anything else.

danielballan:
These commits have been incorporated into #541.

PR #544, opened 2018-05-14
API: Follow PEP8 naming (sumAll -> sum_all).

danielballan:
I noticed this while writing release notes. Since it hasn't been in a release yet I think
we should fix the name.

### DIFF
--- a/ophyd/quadem.py
+++ b/ophyd/quadem.py
@@ -92,7 +92,7 @@ class QuadEM(SingleTrigger, DetectorBase):
     current3 = ADCpt(StatsPlugin, 'Current3:')
     current4 = ADCpt(StatsPlugin, 'Current4:')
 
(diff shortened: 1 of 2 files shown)


tacaswell:
Mixed on this, we do consistently use `_` in component names elsewhere, it has been pointed out by Claudio that we also use `_` as the joiner to replace '.' in the default names so `parent_long_child_name_grand_child` is impossible to tell where Device edges are.

danielballan:
This would be the first place we break PEP8 conventions for component names and the ~100th place we reinforce the problem flagged by Claudio. He raises a good point, but I'm not inclined to spend long musing on this on release day.

danielballan:
Thanks, @stuartcampbell.

@tacaswell I made a separate issue about Claudio's point so we can discuss it in a less rushed moment.

PR #545, opened 2018-05-14
DOC: Release notes for v1.2.0

danielballan:


doc/source/release_notes.rst
```diff
@@ -1,6 +1,93 @@
...
+Features
+********
+
+* On each Signal or Device, attach a Python logger attribute named ``log``
+  with a logger name scoped by module name and the ophyd ``name`` of the
+  parent Device.
+* Signals and Devices now accept ``labels`` argument, a set of labels
+  --- presumed but not (yet) forced to be strings --- which the user can use
+  for grouping and displaying available hardware. The labels are accessible via
+  a new attribute ``_ophyd_labels_``, so name to facilitate duck-typing across
```
mrakitin: Was it originally suggested as `__ophyd_labels_` (double underscore)?

danielballan: No, Python reserves those for its own use. When a library adds an interface, it is supposed to use only one underscore, such as IPython's `_repr_html_`.

mrakitin: I see, thanks!

danielballan:
Do not merge. Expanding on 'kind'.

doc/source/release_notes.rst
```diff
@@ -1,6 +1,135 @@
...
+  unavoidable.
+* Add ``sum_all`` component to QuadEM.
+
+Bug Fixes
+*********
+
+* Allow ``DerivedSignal`` to accept a string name as its target component so
+  that it can be used inside Device, where it must defer grabbing its target to
+  initialization time.
+* Signals that start with underscores are now not renamed by ``namedtuple``.
```
teddyrendahl: Might be worth mentioning these Components are completely omitted from the tuple.

PR #548, opened 2018-05-14
FIX: first pass at keeping lazy components lazy

tacaswell:
Does not seem to be quite enough yet.

### DIFF
--- a/ophyd/tests/conftest.py
+++ b/ophyd/tests/conftest.py
@@ -29,9 +29,6 @@ class FakeEpicsPV(object):
                  auto_monitor=True, enum_strs=None,
                  **kwargs):
 
-        # callbacks mechanism copied from pyepics
-        # ... but tweaked with a weakvaluedictionary so PV objects get
-        # destructed
         self.callbacks = dict()
 
         global _FAKE_PV_LIST
--- a/ophyd/tests/conftest.py
+++ b/ophyd/tests/conftest.py
@@ -92,9 +92,7 @@ class FakeEpicsPV(object):
         if self._pvname in ('does_not_connect', ):
             return
 
-        self._connected = True
         last_value = None
-
         while self._running:
             with self._lock:
                 if self._update:
@@ -106,6 +104,7 @@ class FakeEpicsPV(object):
                     last_value = self._value
 
                 time.sleep(self._update_rate)
+            self._connected = True
 
             time.sleep(0.01)
 
--- a/ophyd/tests/conftest.py
+++ b/ophyd/tests/conftest.py
@@ -29,6 +29,11 @@ class FakeEpicsPV(object):
                  auto_monitor=True, enum_strs=None,
                  **kwargs):
 
+        # callbacks mechanism copied from pyepics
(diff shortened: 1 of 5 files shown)


tacaswell:
This seems to have cut the time of the test suite down by a factor of 2 so I guess that is good?  Need to have a think about how to test and bench mark this for real.

tacaswell:
Master branch `============= 470 passed, 22 skipped, 6 warnings in 686.92 seconds =============`

This PR `======== 1 failed, 469 passed, 22 skipped, 6 warnings in 359.06 seconds ========`

So this has taken ~330s off the test suite run time, that seems like a success to me!

tacaswell:
On my home machine, running just the pyepics version of the tests

current master: 311.89 seconds 
**master right before the kind merge: in 130.26 seconds**
gentle_gentle branch: 168.52 seconds (but 20 more tests than baseline)
**gentle_gentle branch: 126.75 seconds (not hints or kind test, 5 less tests that paseline)**

I think it is safe to say that this works.

ophyd/tests/conftest.py
```diff
@@ -29,6 +29,11 @@ def __init__(self, pvname, form=None,
                  auto_monitor=True, enum_strs=None,
                  **kwargs):
 
+        # callbacks mechanism copied from pyepics
+        # ... but tweaked with a weakvaluedictionary so PV objects get
```
klauer: Looks like an outdated comment

tacaswell:
```
________________________________________________________________________________________________________________ test_mca_read_attrs[pyepics] ________________________________________________________________________________________________________________

    @using_fake_epics_pv
    def test_mca_read_attrs():
        # default read_attrs
        mca = EpicsMCA(devs[0], name='test')
        default_normal_kind = ['preset_real_time', 'elapsed_real_time', 'spectrum']
>       assert set(default_normal_kind) == set(mca.read_attrs)
E       AssertionError: assert {'elapsed_rea...', 'spectrum'} == {'spectrum'}
E         Extra items in the left set:
E         'elapsed_real_time'
E         'preset_real_time'
E         Full diff:
E         - {'elapsed_real_time', 'preset_real_time', 'spectrum'}
E         + {'spectrum'}

ophyd/tests/test_mca.py:35: AssertionError
```

I can see this locally, but re-running it works.  Have only seen this with pyepics (not with caproto).  It is really weird that the fake pv has issues...

tacaswell:
Please don't merge yet, this has a nasty mutating-shared-global-state bug.

tacaswell:
This is ready to go.

tacaswell:
both failures are caproto talking to real epics motor PVs.

tacaswell:
The diff on this looks bigger than it really is because I moved the list proxy to be inside of Device, reviewing with white space changes suppressed may be helpful.

danielballan:
All passed except the nightly build. Thanks, @tacaswell for indulging me with these new tests. I am now satisfied that we have not sacrificed the laziness feature accidentally in the course of adding `kind`.

PR #553, opened 2018-05-18
FIX: remove erroneous configuration in ScalerCH init

tacaswell:
This is now handled by the components and this particular code causes
exceptions on init because it is using the wrong names.


klauer:
I didn't catch that on the `kind` PR that this was handled already. Very nice!

PR #554, opened 2018-05-18
ENH: make_fake_device factory function

ZLLentz:
There isn't a convenient way in the library to test logic provided by `Device` subclasses. It can be done with `using_fake_epics_pv` but it tends to be awkward.

This PR introduces `make_fake_device`, which inspects a device class and creates a "fake" version of it with the same constructor, call signatures, etc. It does this by copying the `Component` objects and swapping the `cls` fields from `EpicsSignal` to `FakeEpicsSignal` and from `EpicsSignalRO` to `FakeEpicsSignalRO`, making a subclass with the fake signals. The fake signals are intended to do as little as possible to avoid adding failure points to tests unrelated to the real devices classes we are spoofing.

This is an idea I was throwing around at SLAC in https://github.com/pcdshub/pcdsdevices/pull/229.

I also make some minor tweaks to `sim.py` in the process.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -627,8 +627,8 @@ class SynSignalWithRegistry(SynSignal):
 
     """
 
-    def __init__(self, *args, reg=DO_NOT_USE, save_path=None, save_func=np.save,
-                 save_spec='NPY_SEQ', save_ext='npy',
+    def __init__(self, *args, reg=DO_NOT_USE, save_path=None,
+                 save_func=np.save, save_spec='NPY_SEQ', save_ext='npy',
                  **kwargs):
         super().__init__(*args, **kwargs)
         self.save_func = save_func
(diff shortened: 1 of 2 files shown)


ophyd/sim.py
```diff
@@ -128,16 +133,16 @@ def get(self):
         return super().get()
 
 
-class SynSignalRO(SynSignal):
+class SignalRO(Signal):
     def put(self, value, *, timestamp=None, force=False):
-        raise NotImplementedError("The signal {} is readonly."
-                                  "".format(self.name))
+        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
```
danielballan: Good call.

tacaswell:
I have not reviewed this in detail, but am :+1: on merging it in principle

PR #556, opened 2018-05-22
REF: make Device.hints as lazy as possible.

danielballan:


danielballan:
Added test and force-pushed to put the test commit first so I could verify that it successfully exercises the bug.

PR #558, opened 2018-05-23
Lazier

tacaswell:
This cuts down the number of pvs that QuadEM connects to on init from 217 to 72.

### DIFF
--- a/ophyd/__init__.py
+++ b/ophyd/__init__.py
@@ -8,7 +8,7 @@ logger.addHandler(logging.NullHandler())
 cl = None
 
 
-def set_cl(control_layer=None):
+def set_cl(control_layer=None, *, pv_telemetry=False):
     global cl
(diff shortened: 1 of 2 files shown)


danielballan:
To elaborate a bit, this PR optionally monkey-patches a `counter` onto the `get_pv` function, which keeps a dict mapping each PV name to the number of times it has been passed to `get_pv` (hopefully 1, if things are well configured). This answers the questions:

* How many PVs am I connected to in my session?
* Do I have more than one instance connected to the same PV (which would be considered misconfiguration)?

Note that PVs can never be totally removed from a Context -- only made to go idle -- so the counts will only ever monotonically increase.

tacaswell:
The more important change in here is making all of the children of DDC created by `ad_group` to be lazy.

danielballan:
How many signals did QuadEM connect to in the previous release? Is the high number (217) due to `kind` or had it always been high?

PR #559, opened 2018-05-24
Fix attr error message

tacaswell:


ophyd/device.py
```diff
@@ -820,18 +820,24 @@ def __attr_list_helper(self, val, set_kind, unset_kind, recurse_name):
...
         # now look at everything else, presumably things with dots
         extra = val - cn
-        if any(c for c in extra if '.' not in c):
-            raise ValueError("You asked to set {c} as a configuration_attr "
-                             "on {self}, but there is no such (grand) child."
-                             .format(c=c, self=self))
+        fail = set(c for c in extra if '.' not in c)
+        if fail:
+
+            raise ValueError("You asked to add the components {fail} "
```
danielballan: `"` -> `f"` here and below

tacaswell: I thought we were still support 3.5 on ophyd.

danielballan: You are right. I missed `.format(...)` below. Already so used to f-strings I expect them everywhere...

PR #560, opened 2018-05-24
Remove read_attrs and config_attrs from __repr__

awalter-bnl:
The __repr__ method returns a string that is to long and confusing, I have removed read_attrs and config_attrs from the string in order to reduce the complexity.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -157,8 +157,10 @@ class Component:
         return '\n'.join(doc)
 
     def __repr__(self):
-        kw_str = ', '.join('{}={!r}'.format(k, v)
-                           for k, v in self.kwargs.items())
+      
+        kw_str = ', '.join('{}={!r}'.format(k, v) for k, v in self.kwargs.items() \
+                           if k not in ['read_attrs','configuration_attrs'])
+                           
         if self.suffix is not None:
             suffix_str = '{!r}'.format(self.suffix)
             if self.kwargs:


danielballan:
Thanks for the fast turnaround, @awalter-bnl!

For anyone following alone, here is some sample output from current master, for context. It makes exceptions way too verbose.

```
QuadEM(prefix='qem:', name='qm', read_attrs=[], configuration_attrs=['integration_time', 
'averaging_time', 'current_names', 'current_offsets', 'current_offset_calcs', 'current_scales', 'current1', 
'current1.configuration_names', 'current1.port_name', 'current1.asyn_pipeline_config', 
'current1.blocking_callbacks', 'current1.enable', 'current1.nd_array_port', 'current1.plugin_type', 
'current1.bgd_width', 'current1.centroid_threshold', 'current1.compute_centroid', 
'current1.compute_histogram', 'current1.compute_profiles', 'current1.compute_statistics', 
'current1.hist_max', 'current1.hist_min', 'current1.hist_size', 'current1.profile_cursor', 
'current1.profile_size', 'current1.ts_control', 'current1.ts_num_points', 'current2', 
'current2.configuration_names', 'current2.port_name', 'current2.asyn_pipeline_config', 
'current2.blocking_callbacks', 'current2.enable', 'current2.nd_array_port', 'current2.plugin_type', 
'current2.bgd_width', 'current2.centroid_threshold', 'current2.compute_centroid', 
'current2.compute_histogram', 'current2.compute_profiles', 'current2.compute_statistics', 
'current2.hist_max', 'current2.hist_min', 'current2.hist_size', 'current2.profile_cursor', 
'current2.profile_size', 'current2.ts_control', 'current2.ts_num_points', 'current3', 
'current3.configuration_names', 'current3.port_name', 'current3.asyn_pipeline_config', 
'current3.blocking_callbacks', 'current3.enable', 'current3.nd_array_port', 'current3.plugin_type', 
'current3.bgd_width', 'current3.centroid_threshold', 'current3.compute_centroid',
'current3.compute_histogram', 'current3.compute_profiles', 'current3.compute_statistics', 
'current3.hist_max', 'current3.hist_min', 'current3.hist_size', 'current3.profile_cursor', 
'current3.profile_size', 'current3.ts_control', 'current3.ts_num_points', 'current4', 
'current4.configuration_names', 'current4.port_name', 'current4.asyn_pipeline_config', 
'current4.blocking_callbacks', 'current4.enable', 'current4.nd_array_port', 'current4.plugin_type', 
'current4.bgd_width', 'current4.centroid_threshold', 'current4.compute_centroid', 
'current4.compute_histogram', 'current4.compute_profiles', 'current4.compute_statistics', 
'current4.hist_max', 'current4.hist_min', 'current4.hist_size', 'current4.profile_cursor', 
'current4.profile_size', 'current4.ts_control', 'current4.ts_num_points'])
```

After a little discussion on NSLS-II Slack, we arrived at the consensus that alternatives, like showing only a _partial_ list in the repr is more potentially confusing than helpful. The most important information in a Device repr is the class, name, and parent.

awalter-bnl:
I agree @danielballan that we need to think about this (which I didn't). perhaps we can separate out __str__ and __repr__ which are currently set to the same. and have one with the read/ config attrs and one without?

danielballan:
@awalter-bnl Not a bad idea in principle. I think a big wall of text for either `print(...)` or `repr(...)` could be surprising at times, so given the string of approvals here I'm inclined to merge. OK by you?

awalter-bnl:
@danielballan Ok with me, I was just suggesting it as a way around a problem that people think is such a problem

klauer:
I've always been a big fan of having objects be represented in a way that you could recreate them through their `repr`, and against things that break it. That said, I don't disagree that a 2000 character repr string is unwieldly and ugly.

My initial thought: if the user leaves the configuration/hints as in the class definition, then they could be removed from the repr without loss of information. It's when that list starts getting customized in different ways that this information is now important. Perhaps eventually something could be done with this down the line in a separate PR.

danielballan:
Thanks @klauer. I think that is worthwhile and feasible. I created an issue: https://github.com/NSLS-II/ophyd/issues/561

@awalter-bnl I suspect that Ken defended his eval-able reprs with a test, which is why all of the builds have failed twice in a row. If I'm right about that, this PR will need to update or remove that test. Can you go digging for it in the Travis logs?

ophyd/device.py
```diff
@@ -157,8 +157,10 @@ def make_docstring(self, parent_class):
         return '\n'.join(doc)
 
     def __repr__(self):
-        kw_str = ', '.join('{}={!r}'.format(k, v)
-                           for k, v in self.kwargs.items())
+        for k, v in self.kwargs.items():
```
danielballan: This logic isn't quite right, and that is the source of Travis failure. In this implementation, `kw_str` gets re-defined on every pass through the loop.

I think it would be sufficient to go back to the original implementation and add `if k not in ['read_attrs','configuration_attrs']` after `items()`. That is, this kind of construction:

```python
# Add the even numbers from 0 to 9.
sum(i for i in range(10) if i % 2 == 0)
```

awalter-bnl: I made this change as I think it is a neater way of solving the problem, although I am not sure why this would give the failure error that Travis gave. Anyway lets see if this solves the issue.

danielballan: I think the approach that triggered the error kept re-defining `kw_str` each time, blowing away contents from the previous iteration of the loop. The do this without a list comprehension, you'd have to define some sort of accumulator variable _before_ the loop. I think that's the approach you were going for here, which would also be fine. :- )

danielballan:
All builds passed except the nightly. Thanks, @awalter-bnl!

danielballan:
Rebased to squash into one commit.

ophyd/device.py
awalter-bnl: thanks @danielballan for the explanation, I now realize why what I did didn't work. Still not sure how that generated the error message travis returned, but I guess I am missing something in how/when this requires importing of ophyd.utils.paths or datetime which is the command that travis complained about. Anyway not important now, it is all good.

danielballan: That ophyd.paths failure might have been a red herring, a random hiccup. The error that tipped me off to the `kw_str` issue was a `NameError`.

awalter-bnl: thanks @danielballan, I think I took a wrong turn at the ophyd.paths failure and got stuck in the mud at the end !-). I didn't even notice the NameError until you mentioned it now and I took a look.

PR #567, opened 2018-06-04
time estimates for plans

awalter-bnl:
This is part of a system that allows for time estimation for plans. It implements a telemetry and time estimation process at the ophyd object level to estimate the time to make a 'cmd' on an object. 'cmd' is the commands past in the plan messages.

An est_time('plan') function (see [NSLS-II/Bluesky PR # 1048](https://github.com/NSLS-II/bluesky/pull/1048) ), similar to summarise_plans('plan'), then steps through each message in order to estimate the time to perform each step and hence the entire plan. 
This fixes  [NSLS-II/Bluesky issue # 713](https://github.com/NSLS-II/bluesky/issues/713), in combination with [NSLS-II/Bluesky PR # 1048](https://github.com/NSLS-II/bluesky/pull/1048)

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -1,9 +1,8 @@
 import functools
 from itertools import count
-
+from .EstTime import DefaultEstTime
 import time
 import logging
-
 from enum import IntFlag
 
 
@@ -77,7 +76,7 @@ class OphydObject:
     _default_sub = None
 
     def __init__(self, *, name=None, attr_name='', parent=None, labels=None,
-                 kind=None):
+                 kind=None, est_time=DefaultEstTime):
         if labels is None:
             labels = set()
         self._ophyd_labels_ = set(labels)
@@ -95,6 +94,7 @@ class OphydObject:
             raise ValueError("name must be a string.")
         self._name = name
         self._parent = parent
+        self.est_time = est_time(self.name)
 
         self.subscriptions = {getattr(self, k)
                               for k in dir(type(self))
@@ -124,7 +124,7 @@ class OphydObject:
 
     def __init_subclass__(cls, version=None, version_of=None,
                           version_type=None, **kwargs):
-        'This is called automatically in Python for all subclasses of OphydObject'
+        'Called automatically in Python for all subclasses of OphydObject'
         super().__init_subclass__(**kwargs)
 
         if version is None:
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -79,6 +79,9 @@ class SynSignal(Signal):
     """
     # This signature is arranged to mimic the signature of EpicsSignal, where
     # the Python function (func) takes the place of the PV.
+
+    settle_time = Component(Signal, value=0, kind='config')
+
     def __init__(self, func=None, *,
                  name,  # required, keyword-only
                  exposure_time=0,
@@ -147,6 +150,17 @@ class SynSignal(Signal):
         # Get a new value, which allows us to synthesize noisy data, for
         # example.
         return super().get()
+
+
+class SynAD_det(SynSignal):
+    '''A SynSignal class with some additional AD related attributes.
+    '''
+
+    trigger_mode = Component(Signal, value=1, kind='config')
+    num_images = Component(Signal, value=1, kind='config')
+    acquire_period = Component(Signal, value=1, kind='config')
+    acquire_time = Component(Signal, value=1, kind='config')
+
 
 
 class SignalRO(Signal):
@@ -299,6 +313,7 @@ class SynAxisNoHints(Device):
 
     velocity = Component(Signal, value=1, kind='config')
     acceleration = Component(Signal, value=1, kind='config')
+    settle_time = Component(Signal, value=0, kind='config')
 
     unused = Component(Signal, value=1, kind='omitted')
 
@@ -437,7 +452,7 @@ class SynAxis(SynAxisNoHints):
     readback = Component(ReadbackSignal, value=None, kind=Kind.hinted)
 
 
-class SynGauss(SynSignal):
+class SynGauss(SynAD_det):
     """
     Evaluate a point on a Gaussian based on the value of a motor.
 
@@ -492,7 +507,7 @@ class SynGauss(SynSignal):
         super().__init__(func=func, name=name, **kwargs)
 
 
-class Syn2DGauss(SynSignal):
+class Syn2DGauss(SynAD_det):
     """
     Evaluate a point on a Gaussian based on the value of a motor.
 
@@ -1278,10 +1293,6 @@ def hw(save_path=None):
     for axis in [motor, motor1, motor2, motor3, jittery_motor1,
                  jittery_motor2]:
         axis.est_time = EpicsMotorEstTime(axis.name)
-        axis.velocity = SynAxisNoHints(name='velocity')
-        axis.velocity.put(1)
-        axis.settle_time = SynAxisNoHints(name='settle_time')
-        axis.settle_time.put(0)
 
     noisy_det = SynGauss('noisy_det', motor, 'motor', center=0, Imax=1,
                          noise='uniform', sigma=1, noise_multiplier=0.1,
@@ -1301,19 +1312,6 @@ def hw(save_path=None):
     det5 = Syn2DGauss('det5', jittery_motor1, 'jittery_motor1', jittery_motor2,
                       'jittery_motor2', center=(0, 0), Imax=1,
                       labels={'detectors'})
-
-    for detector in [noisy_det, det, identical_det, det1, det2, det3, det4,
-                     det5]:
-        detector.acquire_time = SynAxisNoHints(name='acquire_time')
-        detector.acquire_time.put(1)
-        detector.acquire_period = SynAxisNoHints(name='acquire_period')
-        detector.acquire_period.put(1)
-        detector.num_images = SynAxisNoHints(name='num_images')
-        detector.num_images.put(1)
-        detector.trigger_mode = SynAxisNoHints(name='trigger_mode')
-        detector.trigger_mode.put(1)
-        detector.settle_time = SynAxisNoHints(name='settle_time')
-        detector.settle_time.put(0)
 
     flyer1 = MockFlyer('flyer1', det, motor, 1, 5, 20)
     flyer2 = MockFlyer('flyer2', det, motor, 1, 5, 10)
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -16,8 +16,9 @@ from collections import deque, OrderedDict
 from tempfile import mkdtemp
 
 from .signal import Signal, EpicsSignal, EpicsSignalRO
+from .areadetector.trigger_mixins import ADTriggerStatus
 from .areadetector.base import EpicsSignalWithRBV
-from .status import DeviceStatus, StatusBase
+from .status import DeviceStatus, StatusBase, MoveStatus
 from .device import (Device, Component, Component as C,
                      DynamicDeviceComponent as DDC, Kind)
 from types import SimpleNamespace
@@ -26,6 +27,7 @@ from .pseudopos import (PseudoPositioner, PseudoSingle,
 from .positioner import SoftPositioner
 from .utils import DO_NOT_USE, ReadOnlyError, LimitError
 
+from .EstTime import EpicsMotorEstTime, ADEstTime
 logger = logging.getLogger(__name__)
 
 
@@ -101,7 +103,8 @@ class SynSignal(Signal):
                          parent=parent, labels=labels, kind=kind, **kwargs)
         self._metadata.update(
             connected=True,
-        )
+            )
+        self.est_time = ADEstTime(self.name)
 
     def describe(self):
         res = super().describe()
@@ -132,7 +135,13 @@ class SynSignal(Signal):
             return st
         else:
             self.put(self._func())
-            return NullStatus()
+            return ADTriggerStatus(self, done=True, success=True)
+
+    def stage(self):
+        pass
+
+    def unstage(self):
+        pass
 
     def get(self):
         # Get a new value, which allows us to synthesize noisy data, for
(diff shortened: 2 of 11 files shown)


awalter-bnl:
@danielballan in relation to your big picture thoughts.

1.  I only included these as I thought it might be useful to tell when something is going wrong if we have the telemetry from these. Before we remove the telemetry on these we should do some testing to work out if your statement regarding performance overhead of the telemetry is true or not.

2. Trigger, set, etc are modified to 'record' the telemetry not 'report' the telemetry. I do not have any ideas on how that can be done other than what I have done. that is by modifying the action function to track the start time, stop time and any other required info (like start and stop position) send this to the device to record at the end. If you have better ideas on how to generate this information without touching the action functions I am open to it.

3. I see them as 2 distinct processes: 'recording and providing telemetry' and 'time estimation' which is why they are separated. In fact the two can work independently of one another: with the time estimation only relying on the config attrs of the device and so do not want to mix them up in one class.

danielballan:
> Before we remove the telemetry on these we should do some testing to work out if your statement regarding performance overhead of the telemetry is true or not.

I would say the opposite: before _adding_ telemetry to these we should think about why we need it. Under the hood it is doing (1) fast dictionary manipulation and (2) network calls to IOCs to ask them to send back information that they already have ready. Adding telemetry would add a second network hit and database insertion.

> Trigger, set, etc are modified to 'record' the telemetry not 'report' the telemetry.

Sorry, I was unclear. I meant "report" in the sense of "report it to the central database where it is recorded," not "report it to the user to provide at ETA."  I agree that "record" is a clearer word choice here.

>  [...] modifying the action function to track the start time, stop time and any other required info (like start and stop position)

Status objects in fact already track all of this in order to report it to progress bars. (Start reading about [here](https://github.com/NSLS-II/ophyd/blob/master/ophyd/status.py#L450) in the code.) I think adding the code that you need is as simple as adding a couple lines to the definition of the `_finished` method to push that information into globally-defined database. The trigger and set methods shouldn't need to be touched.

> I see them as 2 distinct processes: 'recording and providing telemetry' and 'time estimation' which is why they are separated.

Fully agree that they should be separated. But I'm proposed that `OphydObj` does not need to hold a reference to any objects related to the recording side of things. To re-quote myself from above:

>  I was picturing the Status object reporting telemetry to a module singleton, which does not need to be configurable per OphydObj instance.

If we can agree that Status objects can do all the reporting themselves, via reference to a globally-defined `record_telemetry` function, then `OphydObj` doesn't need to know about the recording side. It just need to know about the time estimation side.

danielballan:
FYI I heavily edited my previous comment for clarity, so anyone following along via email may wish to read the comment on GitHub.

awalter-bnl:
you have convinced me about read etc. I will remove telemetry on that.

I agree to this part, 'if we can do the telemetry on status objects then we don't need it on ophyd objects'. And I agree if this is possible it greatly improves the code. However I am not convinced that you do not need extra information from the ophyd object to record the telemetry. 

From what I read I think status objects do not track start and stop positions (only start and stop times), so we would need to include this extra information. But then we run into a different issue:

Some of the parameters that we are storing data for can only be calculated if you know what the relationship is for a particular ophyd object. For instance with a detector a standard ADdetector has a different attribute (acquisition_period) if it is in fixed mode than if it is not (acquisition_time). Also some custom detectors have other naming conventions (exposure_time). A further complication occurs in the case of motors, which need to record telemetry on velocities based on there setpoints. 

In addition there will be some ophyd objects that have currently unknowable and unique formulas for determining the est_time that will require telemetry on currently unknowable attributes. These inverse of these formulas are required when recording telemetry and so updating 'set', 'trigger' etc.... is the only way that I can see to gain the required flexibility to change what and how the telemetry is saved for specific objects or classes. Do you have some thoughts on how to gain this flexibility using 'status' without having a unique 'status' function for each ophyd device?

danielballan:
> From what I read I think status objects do not track start and stop positions (only start and stop times)

The entries [`initial` and `target`](https://github.com/NSLS-II/ophyd/blob/master/ophyd/status.py#L461) refer to the starting position and the ending position. (That should be made clearer in the docstring.)

> Some of the parameters that we are storing data for can only be calculated if you know what the relationship is for a particular ophyd object. For instance with a detector a standard ADdetector has a different attribute (acquisition_period) if it is in fixed mode than if it is not (acquisition_time). Also some custom detectors have other naming conventions (exposure_time). 

This is a good point which I admit I had not fully thought through. Fortunately, there is already an easy path: area detectors already return a special status object when triggered (`ADTriggerStatus`) and some AD-specific telemetry brains can be added there quite naturally. The status objects already hold a reference to the device that generated them, so they can check on whatever settings they need to know (mode, the relevant acquire time signal, etc). If we need specific logic for a given model or use case, [there is already a hook](https://github.com/NSLS-II/ophyd/blob/master/ophyd/areadetector/trigger_mixins.py#L111) for customizing the type (class) of StatusObject used by a given detector.

The same can apply to motors: EpicsMotor can return `EpicsMotorStatus`, which can reach in and check the velocity, just as `StatusBase` currently reaches in to check the initial position before a move.

> Do you have some thoughts on how to gain this flexibility using 'status' without having a unique 'status' function for each ophyd device?

My proposal does put us down a path of having several more implementations of Status, but I see this as relatively low cost, in terms of work to add it and complexity to understand/maintain/debug. Adding more varieties of status objects (and including richer information in the ones we already have) is less work/complexity that inventing a parallel system from whole cloth. The concepts and interfaces we need are already present in the codebase. This is not a coincidence: it makes sense that telemetry will need many of the same things that progress bars need.

awalter-bnl:
OK so it appears that status objects are much broader than my initial understanding. It also appears that you would prefer to update the status objects instead of the set, trigger etc calls so I will make this change.

Thanks again for the feedback!-)

awalter-bnl:
I have added some more commits here, a description is found in [Bluesky PR 1048](https://github.com/NSLS-II/bluesky/pull/1048)

awalter-bnl:
Have completed the work on this to get the telemetry recording occuring on status objects (except for 'stage' and 'unstage' which require a refactor of bluesky to have them use status objects.

ophyd/ophydobj.py
```diff
@@ -54,8 +63,8 @@ class OphydObject:
 
     _default_sub = None
 
-    def __init__(self, *, name=None, parent=None, labels=None,
-                 kind=None):
+    def __init__(self, *, name = None, parent = None, labels = None,
```
danielballan: The original was correct, PEP8 wise. Assignment statements get spaces, but keyword arguments do not. Example:

```py
def f(x=3):
    y = x + 1
```

ophyd/ophydobj.py
```diff
@@ -72,6 +81,7 @@ def __init__(self, *, name=None, parent=None, labels=None,
             raise ValueError("name must be a string.")
         self._name = name
         self._parent = parent
+        self.est_time = est_time(self)
```
danielballan: Reminder to all of us to circle back to whether we can get away with `est_time(self.name)` or whether we really need to pass in the live object here. If `est_time` becomes responsible for generating the status objects, then we need the live object.

ophyd/sim.py
```diff
@@ -122,10 +127,17 @@ def sleep_and_finish():
...
                 threading.Thread(target=sleep_and_finish, daemon=True).start()
+
             return st
         else:
             self.put(self._func())
-            return NullStatus()
+            return ADTriggerStatus(self, done = True, success = True)
+
+    def stage(self):
+        pass
```
danielballan: I don't think we have settled the question of whether Signals should implement stage/unstage, which is currently limited to `Device`. If it _is_ added, it should be added to all Signals (on `Signal`), not here.

ophyd/sim.py
```diff
@@ -1002,6 +1073,14 @@ def hw():
     jittery_motor2 = SynAxis(name='jittery_motor2',
                              readback_func=lambda x: x + np.random.rand(),
                              labels={'motors'})
+    for axis in [motor, motor1, motor2, motor3, jittery_motor1, jittery_motor2]:
+        axis.est_time = EpicsMotorEstTime(axis)
```
danielballan: I'm not sure it should be allowed to set this on an initialized object. It would be nice to have a hard guarantee that this aren't getting hot-swapped out at random times. Can `EpicsMotorEstTime` be passed in at `__init__` time?

ophyd/status.py
```diff
@@ -536,6 +569,15 @@ def __str__(self):
...
     __repr__ = __str__
 
+    def _finished(self, success = True, **kwargs):
+        '''Inform the status object that it is done, and if it succeeded. If success is True
+        then also record telemetry about the move.
+        '''
+        
+        super()._finished(success = success, **kwargs)
+        if success:
+            self.pos.set.record(self)
```
danielballan: Is this missing an `est_time` accessor? Isn't `self.pos.set` the actual `set` method?

ophyd/signal.py
```diff
@@ -74,8 +74,17 @@ def trigger(self):
         '''Call that is used by bluesky prior to read()'''
         # NOTE: this is a no-op that exists here for bluesky purposes
         #       it may need to be moved in the future
+
```
danielballan: Now that all the reporting is done through the Status objects, do we need any of the changes to this file?

ophyd/sim.py
```diff
@@ -795,12 +866,12 @@ def __repr__(self):
 
 
 class SPseudo3x3(PseudoPositioner):
-    pseudo1 = C(PseudoSingle, limits=(-10, 10), egu='a', kind=Kind.hinted)
-    pseudo2 = C(PseudoSingle, limits=(-10, 10), egu='b', kind=Kind.hinted)
-    pseudo3 = C(PseudoSingle, limits=None, egu='c', kind=Kind.hinted)
-    real1 = C(SoftPositioner, init_pos=0)
-    real2 = C(SoftPositioner, init_pos=0)
-    real3 = C(SoftPositioner, init_pos=0)
+    pseudo1 = C(PseudoSingle, limits=(-10, 10), egu='a', kind=Kind.hinted)#, name = 'pseudo1')
```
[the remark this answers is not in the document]
awalter-bnl: yes. fixed

ophyd/sim.py
```diff
@@ -351,10 +367,62 @@ def position(self):
...
+            return st
+        else:
+            update_state()
+            return MoveStatus(self, value, start_pos = start_pos, done = True, success = True)
+
+
     readback = Component(ReadbackSignal, value=None, kind=Kind.hinted)
 
 
+
```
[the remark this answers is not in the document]
awalter-bnl: have removed them

ophyd/device.py
```diff
@@ -512,6 +512,8 @@ def read(self) -> OrderedDictType[str, Dict[str, Any]]:
             with the keys ``{'value', 'timestamp'}``
 
         """
+
```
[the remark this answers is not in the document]
awalter-bnl: removed them

ophyd/device.py
```diff
@@ -1174,7 +1188,7 @@ def put(self, dev_t, **kwargs):
 
         for attr in self.component_names:
             value = getattr(dev_t, attr)
-            signal = getattr(self, attr)
+            signal = getattr(self, att)
```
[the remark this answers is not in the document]
awalter-bnl: it is and has been undone

ophyd/ophydobj.py
awalter-bnl: went back to the original

ophyd/ophydobj.py
awalter-bnl: I have made this change, as I only ever used self.name inside the est_time classes

ophyd/sim.py
awalter-bnl: It is only done on simulated signals, and was required for testing purposes (the motor axes in the sim devices are based on these signals). Normal signals do not have these attrtibutes.

ophyd/sim.py
awalter-bnl: This is only done for the simulated devices because of the way the sim devices are defined. It is not used in normal ophyd objects.

ophyd/status.py
awalter-bnl: yes fixed

ophyd/signal.py
awalter-bnl: correct I have removed these changes

awalter-bnl: correct, they have been removed

awalter-bnl:
Thanks for the quick response @danielballan , I have made the changes or commented above on your points.

ophyd/status.py
```diff
@@ -446,6 +468,16 @@ def __init__(self, positioner, target, *, start_ts=None,
...
             self.pos.subscribe(self._notify_watchers,
                                event_type=self.pos.SUB_READBACK)
+        else:
+            self.finish_pos = self.target
+            self.finish_ts = time.time()
+            if self.success:
+                try:
+                    self.pos.est_time.set.record(self)
+                except AttributeError:
+                    print ('est_time.set.record attribute required on {}, but not found'.format(self.pos))
```
danielballan: The problem with printing is that there is no way for the caller -- who might be able to handle this error -- to suppress the print. Instead I suggest this pattern:

```py
except AttributeError:
    raise AttributeError(your_custom_message_here)
```

ophyd/positioner.py
```diff
@@ -81,9 +81,17 @@ def set(self, new_position: Any, *,
         status : StatusBase
             Status object to indicate when the motion / set is done.
         """
+        val_dict = {}   #a dictionary used in the time estimation process
```
danielballan: I guess this should go away too now?

ophyd/sim.py
awalter-bnl: I agree that having the sim devices significantly different does not make them effective 'test' systems or examples.

Unfortunately when I looked at this it appears to me that the sim devices are already 'very very' different and to make them consistent would require a complete rewrite of the sim devices. I think this is outside the scope of the current PR, but perhaps could be looked at in the future.

ophyd/signal.py
awalter-bnl: I thought I found them all, I will take another look.

ophyd/status.py
awalter-bnl: I tried this and kept getting an error, I will try again.

ophyd/positioner.py
awalter-bnl: yep. I will remove it.

awalter-bnl:
So I rebased and that has not resolved the issue with the release notes etc. I am sure I am doing something wrong but not sure what at present, will keep working on it

awalter-bnl:
The recent push leaves out the part where the mongo database is setup, I am canvassing for opinions on which client to use and where to do the setup of the database. (comments welcomed).

awalter-bnl:
Have added the components to telemetry.db that allow for a mongo db to be used to store the telemetry. NOTE: I am sure that this will need modification, I based it on how the databroker database is setup but there was a lot of stuff in there that I didn't understand and so it is significantly simpler and likely to not be what is required.

FYI this approach would require a new yml file in  playbooks/roles/telemetry_config/files/XXX.yml and for the line `TelemetryUI.configure_db(name)` to be added to `nslsii/configure_base`.

awalter-bnl:
It has some weird ophyd/sim related errors. These may be a result of the current code (although it is working so I can't be 100% sure yet). @danielballan I noticed on slack you mentioned 'work on getting ophyd tests passing' is this related to that? if not I will investigate further, if unsure let me know once that is done and I will come back to this.

awalter-bnl:
After a discussion with @danielballan this afternoon, and based on how much this work is required to bring this up to date with the current master, I am closing this PR. After a meeting tomorrow where we will decide how to run these PR's internally I suspect we will start again with a fresh PR.

PR #568, opened 2018-06-05
ROIplugin update to include 'set' attribute for ROI region

awalter-bnl:
This adds the set, read and describe attributes to the ROIplugin class. This allows the ROI regions to be 'set' and 'read' like normal positioners, and hence allows them to be included into 'log_pos' an 'baseline' readings as well as being set in 'mv(....)' plans.
This was added after seeing that a beamline had made some changes to allow this functionality in there startup file.

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -569,16 +569,7 @@ class OverlayPlugin(PluginBase):
 
 
 class ROIPlugin(PluginBase):
-    def __init__(self,*args, **kwargs):
-        self.min_xyz.min_x.name = '{name}_min_x.format(name=self.name)'
-        self.min_xyz.min_y.name = '{name}_min_y.format(name=self.name)'
-        self.min_xyz.min_z.name = '{name}_min_z.format(name=self.name)'
-        self.size.x.name = '{name}_size_x.format(name=self.name)'
-        self.size.y.name = '{name}_size_y.format(name=self.name)'
-        self.size.z.name = '{name}_size_z.format(name=self.name)'
-        super().__init__(*args, **kwargs)
-       
-      
+ 
     _default_suffix = 'ROI1:'
     _suffix_re = 'ROI\d:'
     _html_docs = ['NDPluginROI.html']
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -577,7 +577,9 @@ class ROIPlugin(PluginBase):
     _default_configuration_attrs = (PluginBase._default_configuration_attrs + (
         'roi_enable', 'name_', 'bin_', 'data_type_out', 'enable_scale')
     )
-    _default_read_attrs = ['enable', 'min_xyz.min_x', 'size.x', 'min_xyz.min_y', 'size.y', 'min_xyz.min_z', 'size.z']
+    _default_read_attrs = ('enable', 'min_xyz.min_x', 'size.x',
+                           'min_xyz.min_y', 'size.y', 'min_xyz.min_z',
+                           'size.z')
     array_size = DDC(ad_group(EpicsSignalRO,
                               (('x', 'ArraySizeX_RBV'),
                                ('y', 'ArraySizeY_RBV'),
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -639,20 +639,24 @@ class ROIPlugin(PluginBase):
     def set(self, region):
         ''' This functions allows for the ROI regions to be set.
 
-        This function takes in an ROI_number, and a dictionary of tuples and sets the ROI region.
+        This function takes in an ROI_number, and a dictionary of tuples and
+        sets the ROI region.
 
         PARAMETERS
         ----------
         region: dictionary.
-            A dictionary defining the region to be set, which has the structure:
-            { 'x':[min, size], 'y':[min, size], 'z':[min, size]}. 
-            Any of the keywords can be ommitted, and they will be ignored.
+            A dictionary defining the region to be set, which has the
+            structure:
+            ``{'x': [min, size], 'y': [min, size], 'z': [min, size]}``. Any of
+            the keywords can be omitted, and they will be ignored.
         '''
         if region is not None:
             status = []
             for direction, value in region.items():
-                status.append(getattr(self, 'min_xyz.min_{}'.format(direction)).set(value[0]) )
-                status.append(getattr(self, 'size.{}'.format(direction)).set(value[1]) )
+                status.append(getattr(
+                    self, 'min_xyz.min_{}'.format(direction)).set(value[0]))
+                status.append(
+                    getattr(self, 'size.{}'.format(direction)).set(value[1]))
                 
         return functools.reduce(operator.and_, status)
 
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -569,6 +569,16 @@ class OverlayPlugin(PluginBase):
 
 
 class ROIPlugin(PluginBase):
+    def __init__(self,*args, **kwargs):
+        self.min_xyz.min_x.name = '{name}_min_x.format(name=self.name)'
+        self.min_xyz.min_y.name = '{name}_min_y.format(name=self.name)'
+        self.min_xyz.min_z.name = '{name}_min_z.format(name=self.name)'
+        self.size.x.name = '{name}_size_x.format(name=self.name)'
+        self.size.y.name = '{name}_size_y.format(name=self.name)'
+        self.size.z.name = '{name}_size_z.format(name=self.name)'
+        super().__init__(*args, **kwargs)
+       
+      
     _default_suffix = 'ROI1:'
     _suffix_re = 'ROI\d:'
     _html_docs = ['NDPluginROI.html']
@@ -576,6 +586,7 @@ class ROIPlugin(PluginBase):
     _default_configuration_attrs = (PluginBase._default_configuration_attrs + (
         'roi_enable', 'name_', 'bin_', 'data_type_out', 'enable_scale')
     )
+    _default_read_attrs = ['enable', 'min_xyz.min_x', 'size.x', 'min_xyz.min_y', 'size.y', 'min_xyz.min_z', 'size.z']
     array_size = DDC(ad_group(EpicsSignalRO,
                               (('x', 'ArraySizeX_RBV'),
                                ('y', 'ArraySizeY_RBV'),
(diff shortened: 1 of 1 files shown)


ophyd/areadetector/plugins.py
```diff
@@ -632,6 +633,56 @@ class ROIPlugin(PluginBase):
...
+        PARAMETERS
+        ----------
+        region: dictionary.
+            A dictionary defining the region to be set, which has the structure:
+            { 'x':[min, size], 'y':[min, size], 'z':[min, size]}. 
+            Any of the keywords can be ommitted, and they will be ignored.
+        '''
+        if region is not None:
+            for direction, value in region.items():
+                getattr(self, f'min_xyz.min_{direction}').put(value[0])
```
tacaswell: we should use `set` and `&` all of the status object together instead of returning `NullStatus` here.

ophyd/areadetector/plugins.py
```diff
@@ -632,6 +633,56 @@ class ROIPlugin(PluginBase):
...
+                getattr(self, f'size.{direction}').put(value[1])
+
+        return NullStatus()
+        
+    def read(self):
```
tacaswell: Could this be done by setting the right `read_attrs` on the ROI and the right names on the children?

tacaswell:
Something went funny with git, re-targeted this PR against master.

ophyd/areadetector/plugins.py
awalter-bnl: I am not sure what you mean by 'use set and & all of the status objects together', can you elaborate?

ophyd/areadetector/plugins.py
awalter-bnl: I am not sure on this, possibly but the issue is that the children that are referenced here in read are a few levels down the tree and I was hoping to avoid including this level of detail in the read.

ophyd/areadetector/plugins.py
```diff
@@ -632,6 +633,56 @@ class ROIPlugin(PluginBase):
...
+        out_dict = OrderedDict({})
+        out_dict[f'{self.name}_status'] = {'timestamp': ttime.time(), 'value': self.enable.get() }
+        for direction in ['x', 'y', 'z']:
+            out_dict[f'{self.name}_{direction}'] = {'timestamp': ttime.time(), 
+                                        'value': (getattr(self, f'min_xyz.min_{direction}').get(),
+                                        getattr(self, f'size_{direction}').get())}
+
+        return out_dict
+
+    def dscribe(self):
```
klauer: `dscribe`

ophyd/areadetector/plugins.py
danielballan: `NullStatus()` is a cheat -- it says, "I'm done." Using it here would give the RunEngine false information: the settings above have not necessarily been processed.

To give the RunEngine accurate information, we normally replace:

```
signal.put(val)
return NullStatus()
```

with

```
status = signal.set(val)
return status
```

but here because you are setting multiple things, and we want to know when they are _all_ done, we can use the fact that status objects can be logically AND-ed like `status1 & status2 & status3`. To do this for an arbitrary number of status objects you'll need `functools.reduce` and `operator.and_`.

ophyd/areadetector/plugins.py
awalter-bnl: Thanks @klauer I got that one in the recent commit where I changed the f string syntax

ophyd/areadetector/plugins.py
danielballan: Would setting

```
_default_read_attrs = ['x', 'y', 'z', 'status']
```

and using the standard `read()` method achieve the same effect as this custom `read()` and `describe()` methods?

awalter-bnl: I am not sure it would because 'x' is actually the attribute  min_xyz.min_x so I would need to use
_default_read_attrs = ['min_xyz_min_x', .........
which is what I was trying to avoid with the custom read as then the read is a little to verbose for my liking?

also 'status' in this case would need to be named 'enable' (which I think is OK).

danielballan: > the read is a little to verbose for my liking

Meaning the keys in the dictionary returned by read? The hook that ophyd provides for fixing that is in changing the attribute's `.name`.

> Also I would have to also include the 'size' attributes seperately where as in my read method I combine them into a tuple.

This kind of rearrangement should be done by making special Signals that give you the readings you want. The generality of `read()` across all devices is a really powerful assumption to be able to rely on, one we shouldn't break unless we have no other option, because the cost of having to account for specialize `read()` behavior in downstream code is high.

ophyd/areadetector/plugins.py
```diff
@@ -646,15 +646,12 @@ def set(self, region):
...
-                if status == None:
-                    status = getattr(self, 'min_xyz.min_{}'.format(direction)).put(value[0])
-                else:
-                    status = status & getattr(self, 'min_xyz.min_{}'.format(direction)).put(value[0])
-                status = status & getattr(self, 'size.{}'.format(direction)).put(value[1])
+                status.append( getattr(self, 'min_xyz.min_{}'.format(direction)).put(value[0]) )
+                status.append( getattr(self, 'size.{}'.format(direction)).put(value[1]) )
                 
-        return status
+        return functtools.reduce(operator.and_, status)
```
danielballan: There's an extra `t` in `functools`.

ophyd/areadetector/plugins.py
```diff
@@ -646,15 +646,12 @@ def set(self, region):
...
         if region is not None:
-            status = None
+            status = []
             for direction, value in region.items():
-                if status == None:
-                    status = getattr(self, 'min_xyz.min_{}'.format(direction)).put(value[0])
-                else:
-                    status = status & getattr(self, 'min_xyz.min_{}'.format(direction)).put(value[0])
-                status = status & getattr(self, 'size.{}'.format(direction)).put(value[1])
+                status.append( getattr(self, 'min_xyz.min_{}'.format(direction)).put(value[0]) )
```
danielballan: The extra space after `append(` breaks PEP8. I imagine you put it there in a reasonable effort to make this long line easier to parse. Another way to do that, and reduce repetition, would be:

```py
status0 = getattr(___).set(value[0])
status1 = getattr(___).set(value[1])
statuses.extend([status0, status1])
```

Also remember the earlier comment that `put` needs to be changed to `set`. "Put" is the low-level "Just send the EPICS command," which is ultimately at the bottom of "set" as well, but "set" adds additional machinery to track done-ness.

ophyd/areadetector/plugins.py
```diff
@@ -576,6 +576,7 @@ class ROIPlugin(PluginBase):
     _default_configuration_attrs = (PluginBase._default_configuration_attrs + (
         'roi_enable', 'name_', 'bin_', 'data_type_out', 'enable_scale')
     )
+    _default_read_attrs = ['enable', 'min_x', 'size_x', 'min_y', 'size_y', 'min_z', 'size_z']
```
danielballan: These have to match the attribute names of components, so `max_x` should be `max_xy.x`. Separately (in `__init__`), you can tell `max_xy.x` to call itself `max_x` like so:

```py
    self.max_xy.x.name = '{name}_max_x'.format(name=name)
    ...
```

The change will effect what the Component thinks its name is -- what shows up in `read()` and goes into the documents and thus all the downstream display/database etc. -- but it does not change its _attribute name_ so the reference `_default_read_attrs` and `read_attrs` etc. should stay the same.

ophyd/areadetector/plugins.py
```diff
@@ -569,13 +569,24 @@ class OverlayPlugin(PluginBase):
...
 
 class ROIPlugin(PluginBase):
+    def __init__(self,*args, **kwargs):
+        self.min_xyz.min_x.name = '{name}_min_x.format(name = self.name)'
+        self.min_xyz.min_y.name = '{name}_min_y.format(name = self.name)'
+        self.min_xyz.min_z.name = '{name}_min_z.format(name = self.name)'
+        self.size.x.name = '{name}_size_x.format(name = self.name)'
+        self.size.y.name = '{name}_size_y.format(name = self.name)'
+        self.size.z.name = '{name}_size_z.format(name = self.name)'
+        super().__initi__(*args, **kwargs)
```
danielballan: `__initi__` -> `__init__`

ophyd/areadetector/plugins.py
```diff
@@ -632,6 +643,26 @@ class ROIPlugin(PluginBase):
...
+        PARAMETERS
+        ----------
+        region: dictionary.
+            A dictionary defining the region to be set, which has the structure:
+            { 'x':[min, size], 'y':[min, size], 'z':[min, size]}. 
+            Any of the keywords can be ommitted, and they will be ignored.
+        '''
+        if region is not None:
+            status = []
+            for direction, value in region.items():
```
danielballan: Re-uppping a previous comment that may have been swallowed by GitHub's UI:

The extra space after `append(` breaks PEP8. I imagine you put it there in a reasonable effort to make this long line easier to parse. Another way to do that, and reduce repetition, would be:

```py
status0 = getattr(___).set(value[0])
status1 = getattr(___).set(value[1])
statuses.extend([status0, status1])
```

Also remember the earlier comment that `put` needs to be changed to `set`. "Put" is the low-level "Just send the EPICS command," which is ultimately at the bottom of "set" as well, but "set" adds additional machinery to track done-ness.

ophyd/areadetector/plugins.py
```diff
@@ -569,13 +569,24 @@ class OverlayPlugin(PluginBase):
 
 
 class ROIPlugin(PluginBase):
+    def __init__(self,*args, **kwargs):
+        self.min_xyz.min_x.name = '{name}_min_x.format(name = self.name)'
```
danielballan: `(name = self.name)` -> `(name=self.name)`

ophyd/areadetector/plugins.py
danielballan: ^ The only really important comment here is `put` -> `set`.

danielballan:
@awalter-bnl I rearranged the hit history a bit and force-pushed to this branch.

* I squashed most of your commits into just one commit, f0bd4d0. All of those old commit messages are still included in it. We do this sometimes when a PR takes a indirect/exploratory path toward the final result -- it's not always worth keeping that whole path in the commit history of the project.
* I kept your "remove `__init__` ..." commit separate so that you can go back that the `__init__` is _not_ scrubbed from history and you can look it up to keep trying to make it work in future.
* I added two of my own commits on top -- one fixing a detail I didn't catch yesterday and the other just addressing PEP8.

danielballan:
Stepping back, this is a very nice addition to have! We probably should have done this in the first place! :- )

PR #570, opened 2018-06-06
BLD: include more files in distribution

mrakitin:
See https://github.com/NSLS-II/bluesky/pull/1042 for description of the problem.


mrakitin:
Thanks for the review. I updated the sorting, which originally was case-sensitive.

danielballan:
Ah, makes sense. Thanks.

mrakitin:
Thanks for merging, @danielballan!

PR #572, opened 2018-06-12
Pre defined positions not device

awalter-bnl:
This is an in-progress work. It creates a PreDefinedPositionsClass which is not a child of Device. Instead it takes in a list of 'components' and attribute references to them.

The aims of this new class are:

1. To provide a common interface for a collection of components that are experimentally connected but are not the same physical hardware.
2. To allow for a series of 'pre defined positions' to be specified so that the device can be used as a 'multi-state' positioner. This is useful when there are a series of 'locations' which are regularly accessed, and hence a quick interface to move to these locations is required.
3. To provide a method for moving between states through semi-complicated 'paths' .

### DIFF
--- a/ophyd/PreDefinedPositions.py
+++ b/ophyd/PreDefinedPositions.py
@@ -14,12 +14,15 @@ class PreDefinedPositions():
     between these locations via calculated 'paths'. These 'paths' are
     calculated based on the optional 'neighbours' dictionary, which for each
     'location' defines a list of other locations that it is 'safe' to move
-    directly to. The 'components' in the instance can be any 'component' in
-    ophyd: such as motor axes, detectors, gate valves, their configuration
-    attributes and or even another PreDefinedPositions instance. The
-    'locations' do not need to have a value specifed for every component in the
-    collection, and the collection can be in more than one 'location' at any
-    given time. A further optional dictionary allows the user to define a
+    directly to.
+
+    The 'components' in the instance can be any 'component' in ophyd: such as
+    motor axes, detectors, gate valves, their configuration attributes and or
+    even another PreDefinedPositions instance. The 'locations' do not need to
+    have a value specifed for every component in the collection, and the
+    collection can be in more than one 'location' at any given time.
+
+    A further optional dictionary ('regions') allows the user to define a
     'volume' for all or some 'locations', within which the device is considered
     'in the location'. Moves to/from a location always end at the 'location'
     defined in the 'locations' dictionary. If a 'volume' is not defined for any
@@ -43,22 +46,24 @@ class PreDefinedPositions():
                    location2:[axis1,value1,axis2,value2,...],.....}.
             .. note::
                 #. Not all axes need to have a specifed value for each device
-                   location, only those with a specifed value are moved/checked
-                   for a given location.
+                   location.
+                #. Only axes with a specifed value are moved/checked for a
+                   given location.
                 #. All axes specifed in this dictionary must be specifed as
                    components in 'components'.
     neighbours : Dictionary, optional.
         A keyword:value dictionary where each keyword is a location defined in
         'locations' and each value is a list of 'neighbours' for that location.
-        Motion can only occur between neighbours, for non-neighbours a path
-        through various locations will be used, if it is found using
-        self.find_path. Optionally if the list contains 'All' this indicates
-        that evey location is accesible from this one. If no neighbours are
-        defined for this location then it is assumed that no direct motion is
-        allowed from this location.
+        a 'neighbour' is defined as a second location for which it is safe to
+        move firectly too from the keyword 'location'. Motion can only occur
+        between neighbours, for non-neighbours a path through various locations
+        will be used, if it is found using self.find_path. Optionally if the
+        list contains 'All' this indicates that evey location is accesible from
+        this one. If no neighbours are defined for this location then it is
+        assumed that no direct motion is allowed from this location.
     regions : Dictionary, optional.
         The optional keyword:value dictionary that has location keywords and
-        'region' values 'region' is a dictionary that has axis_name keywords
+        'region' values. 'region' is a dictionary that has axis_name keywords
         and [min_val, max_val] values denoting the range of values for this
         location and axis. This dictionary has the form:
             {location1:{'axis1_name':[axis1_min_val,axis1_max_val],
@@ -67,11 +72,11 @@ class PreDefinedPositions():
                         'axis2_name':[axis2_min_val,axis2_max_val],...},
                                           ....}.
             .. note::
-                Not all locations in the 'locations' require an entry in this
-                dictionary and not all axes defined with a value in 'locations'
-                , for a given location, must have a range in this dictionary
-                for the given location. For any axis that a 'range' is not
-                provided a default range of +/- 1% is used.
+                #. Not all locations in the 'locations' require an entry in
+                   this dictionary and not all axes defined with a value in
+                   'locations', for a given location, must have a range in this
+                   dictionary for the given location. For any axis that a
+                   'range' is not provided a default range of +/- 1% is used.
 
     .. note::
         NOTES ON PREDEFINED MOTION WITH NEIGHBOURS:
@@ -109,12 +114,12 @@ class PreDefinedPositions():
         def mv_plan(to_location):
             '''Returns a plan to move to 'to_location'.
 
-            A function that returns a plan that can move the collection  to the
+            A function that returns a plan that can move the collection to the
             location defined by 'to_location'
 
             Parameters
             ----------
-            to_location: string
+            to_location : string
                 The name of the location that it is required to move too.
             '''
 
@@ -144,17 +149,17 @@ class PreDefinedPositions():
 
         Parameters
         ----------
-        value, string:
+        value : string
             The pre defined location that the object is to move to.
 
         Returns
         -------
-        status, MoveStatus object:
+        status : MoveStatus object
             Returns a MoveStatus object.
 
         Raises:
         -------
-        ValueError:
+        ValueError :
             On invalid locations.
 
         '''
@@ -183,7 +188,7 @@ class PreDefinedPositions():
 
         Parameters
         ----------
-        read_dict: ordered dictionary, ouput
+        read_dict : ordered dictionary, ouput
             The output dictionary that matches the standard output for a
             Device.
         '''
@@ -201,12 +206,12 @@ class PreDefinedPositions():
         '''
         An attribute that returns the current 'location' description of the
         output data as an ordered dictionary. This is used identically to the
-        describe attribute function for a standard device and therefore can be
-        used in the baseline.
+        describe attribute function for a standard device and therefore allows
+        pre-defined locatioans to be used in the baseline.
 
         Parameters
         ----------
-        describe_dict: ordered dictionary, ouput
+        describe_dict : ordered dictionary, ouput
             The output dictionary that matches the standard output for a
             Device.
         '''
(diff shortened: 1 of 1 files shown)


awalter-bnl:
The process to get the first 2 objectives is reasonably obvious, the first is done through the input of a key/value dictionary of 'attribute name'/'component' pairs. The second is done via the input of a 'location' dictionary with the key being the 'location name' and the value being a list of succesive 'axis,value' pairs, similar to the input for the mv sub-plan.

The third is realized by specifying a 'location' for each point on the path, and a 'neighbours' dictionary that indicates for each location which other locations can be directly accessed. This is used to calculate the shortest path between a start and end location. It is further expanded by the addition of a 'regions' dictionary that for each location defines a volume for which it can be considered 'in' this location.

awalter-bnl:
The only concern I have with this approach is that we are starting to pollute the name space. As we will now have 'Devices' and 'PreDefinedPositions' which can both be used to access the same components. My workaround for this is that perhaps the 'Devices' can have an `_` prefix in order to hide them from the tab-to-complete list to avoid user confusion.

teddyrendahl:
There is `SLAC` code for storing positions of devices. We refer to it as `presets`. We write it to a `YAML` file so that these locations survive sessions, e.t.c. 

https://github.com/pcdshub/pcdsdevices/blob/40ee15a5e1cfada63f3c9b0a77da165713d7d94c/pcdsdevices/mv_interface.py#L169

awalter-bnl:
Thanks @teddyrendahl. This is good to know, but this class goes a bit further by allowing complicated ‘path’ motion between locations and grouping of non-hardware related, but experimentally related, components. Our solution at present is that the positions info are defined in the beamline startup files, and hence carry over between sessions.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+               'source': 'None',
+               'units': None,
+               'upper_ctrl_limit': None}
+              
+        describe_dict = super().describe()
+        describe_dict.update(out_dict)
+        
+        return describe_dict
+
+
```
mrakitin: should be one blank line.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+    def find_path(self,from_location=None,to_location=None):
```
mrakitin: blank spaces after `,`'s

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        '''
+        Find the shortest path from 'from_location' to 'to_location' passing only thorugh the
```
mrakitin: through

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        describe_dict = super().describe()
+        describe_dict.update(out_dict)
+        
+        return describe_dict
+
+
+    def find_path(self,from_location=None,to_location=None):
+        '''
+        Find the shortest path from 'from_location' to 'to_location' passing only thorugh the 
+        neighbours for each location defiend by the dictionary 'neighbours'. Returns an empty list
```
mrakitin: defined

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+    def find_path(self,from_location=None,to_location=None):
+        '''
+        Find the shortest path from 'from_location' to 'to_location' passing only thorugh the 
+        neighbours for each location defiend by the dictionary 'neighbours'. Returns an empty list 
+        if no path found otherwise returns a list of 'locations' that define the path. If to_location 
+        is None then it returns a dictionary showing the shortest path to all possible locations. If
+        from_location is None it returns a dictionary showing the shortest path from all locations to
+        the current location. If both are None it returns a dictioanry of dictionaries. If
```
mrakitin: dictionary

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        Find the shortest path from 'from_location' to 'to_location' passing only thorugh the 
+        neighbours for each location defiend by the dictionary 'neighbours'. Returns an empty list 
+        if no path found otherwise returns a list of 'locations' that define the path. If to_location 
+        is None then it returns a dictionary showing the shortest path to all possible locations. If
+        from_location is None it returns a dictionary showing the shortest path from all locations to
+        the current location. If both are None it returns a dictioanry of dictionaries. If 
+        from_location is 'current_location' the starting point is changed to the current location.
+        Parameters
+        ---------
+        from_location: string
```
mrakitin: string or None

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+            The name of the starting location required for the path.
+        to_location: string
```
mrakitin: see above

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+            The name of the ending location required for the path.
+        path_list: list, output
+            A list locations indicating the path to take to reach the required position.
+        '''
+
+
```
mrakitin: one blank line should be here

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        if from_location is not 'current_location':
+            path_list = nx.shortest_path(self.nxGraph,source=from_location,target=to_location)
```
mrakitin: blank spaces after `,`'s

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        elif isinstance(self.status_list,str) and self.neighbours is not None:
+            path_list=(self.status_list)
```
mrakitin: do you need parentheses here? Are you converting a string value to a list? Then `[]` should be used.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+            else:
+                path_list=[]
+                for location in self.status_list:
+                    prev_path_list=path_list
+                    path_list=nx.shortest_path(self.nxGraph,source=location,target=to_location)
+
+                    if len(prev_path_list)>1 and len(prev_path_len)<len(path_list):
+                        path_list=prev_path_list
+
+        return path_list
```
mrakitin: Please include all missing blank spaces in the code block above (` = `, etc.)

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+
+
```
mrakitin: one blank line is enough.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+    def visualize_paths(self, fig_size = [10,10], axis_labels = ['arb. axis', 'arb. axis'] ,
+                        options = {}):
```
mrakitin: I'd format is as follows:
```py
    def visualize_paths(self, fig_size=(10, 10), axis_labels=('arb. axis', 'arb. axis'),
                        options={}):
```

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+                    'arrow_size':12}
+            default_options.update(options)
+
+            plt.figure('visualize {} paths'.format(self.name),figsize=(fig_size[0],fig_size[1]))
+            plt.xlabel(axis_labels[0])
+            plt.ylabel(axis_labels[1])
+            nx.draw_networkx(self.nxGraph,arrows=True,**default_options)
+
+
+
```
mrakitin: too many blank lines, should be one.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        PARAMETERS
+        ----------
+        fig_size: list, optional
+            A list containing the horizontal and vertical size to make the figure.
+        axis_labels: list, optional 
+            A list of horizontal and vertical axis names for the figure.
+        options: dict, optional
+            An optional dictionary that contains kwargs for the draw_networkx function from the
+            networkx module.
+
```
mrakitin: Please remove the blank line.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        '''
+
+        if self.locations is None:
+            print ('No locations to visualize')
+        else:
+
```
mrakitin: Please remove the blank line.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+            
+            default_options={'pos':nx.circular_layout(self.nxGraph),'node_color':'darkturquoise',
+                    'edge_color':'grey','node_size':6000, 'width':3,'arrowstyle':'-|>',
+                    'arrow_size':12}
+            default_options.update(options)
+
+            plt.figure('visualize {} paths'.format(self.name),figsize=(fig_size[0],fig_size[1]))
+            plt.xlabel(axis_labels[0])
+            plt.ylabel(axis_labels[1])
+            nx.draw_networkx(self.nxGraph,arrows=True,**default_options)
```
mrakitin: Please format the code above better, so that the `default_options` dict would look more readable and the blank spaces after `,`'s are added where necessary.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+        Returns
+        -------
+        location_list : list
+        '''
+   
+        if self.locations is not None:
+            location_list='unknown location'
+            for location in self.locations:
+                in_position=True
+                for i in range(0,len(self.locations[location]),2):
```
mrakitin: blank spaces in two lines above.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+                    axis = self.locations[location][i]
+                    value = self.locations[location][i+1]
+
+                    if isinstance(value,float):
+                       if axis in self.regions[location]:
+                           if get_attr(self,axis).position < self.regions[location][axis][0] \
```
mrakitin: `getattr`, not `get_attr`

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+                                or get_attr(self,axis).position > self.regions[location][axis][1]:
+                               in_location = false
```
mrakitin: `False`

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+                       else:
+                           if get_attr(self,axis).position < .99*self.locations[location][axis] \
+                               or get_attr(self,axis).position > 1.01*self.regions[location][axis]:
+                               in_location = false
```
mrakitin: `getattr` and `False`. Also, add missing blank spaces

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+
+                    elif isinstance(value,str):
```
mrakitin: what is `value`? Did you mean `location`?

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+                    if location_list == 'unknown location':
+                        location_list = [location]
+                    else:
+                        location_list.append(location)
+        else:
+            location_list=['no locations']
+
+        return location_list
+   
+
```
mrakitin: remove those two blank lines.

ophyd/PreDefinedPositions.py
```diff
@@ -0,0 +1,321 @@
...
+                        if getattr(self,axis).status is not value:
+                            in_position=False
+
+                if in_position: 
+                    if location_list == 'unknown location':
+                        location_list = [location]
+                    else:
+                        location_list.append(location)
+        else:
+            location_list=['no locations']
```
mrakitin: Take care of the blank spaces around `=` above.

awalter-bnl:
Thanks @mrakitin, this is a long way form being ready to commit, and has not yet been tested. It was placed as a PR because @danielballan wanted to be able to look at the differences. In fact it is not yet sure that this is the approach we will adopt, and was written more to see how this implementation would look. I suspect that this PR and PR 575 will be closed this week and a new version opened based on todays meeting.

klauer:
I like the idea overall but got lost in the verbose details of the docstrings. Think you could provide a simple code example using these? (That applies to #575 as well, I think...)

awalter-bnl:
@klauer we are still in deep discussion about how to proceed with this. Once we agree on the best path forward then I can provide the simple examples.

klauer:
Would supplying an example of how you expect the code to be used not contribute toward finding the preferred path forward? I often find it to be essential...

awalter-bnl:
Hi @klauer , 
I actually have 2 fully working versions (an old one and a new one) which are being used at beamlines at NSLS-II already. The discussion is not about what we want to achieve but how we structure it. That being said the exact API is yet to be finalized, so the stuff below may end up being wrong.

For the existing versions the examples are:
- say we have a PreDefinedPosition instance called `pdp`. Then:
`RE(pdp.some_location)` will move all of the child 'positioners' to the values for 'some_location' defined in the input 'locations' dictionary. The output of pdp.some_location is a 'plan' defined using 'move' to go form one location to the other.

As it may be that we don't want to go from our current location directly to some_location, then the system also determines a path from one to the other only passing through the 'neighbours' defined in neighbours. From a code point of view think of it like the locations being 'nodes' on a networkx graph and the neighbours being the 'edges'.

How do I suspect this will be used:
from the command line allowing a quick way to move between pre-defined 'locations'
`>RE(pdp.some_location)`

In user defined plans I think it may look like

`def some_plan():`
`      yield from scan(...........)`
`      yield from pdp.some_location`
`      yield from scan (..........)`

Hopefully this clarifies

klauer:
I think I see now. Thanks for helping me understand.

awalter-bnl:
I have moved the import of matplotlib to avoid an ophyd dependency based on a comment from @danielballan in PR #596

PR #573, opened 2018-06-12
FIX: read_attrs/config_attrs backcompatibility __contains__

ZLLentz:
This method was raising an `AttributeError` instead of returning `False` when something was not in the list e.g. `'not_a_real_signal' in device.read_attrs` would error out. I opted to check the full list rather than taking any shortcuts.

I ran into this when testing some of our older modules with `ophyd=1.2.0` that were checking if certain strings were in the `read_attrs` list. Is there a suggested way to check if something will be read if `read_attrs` is ever removed?

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1274,7 +1274,7 @@ class Device(BlueskyInterface, OphydObject, metaclass=ComponentMeta):
             getattr(self._parent, value).kind &= ~self._remove_kind
 
         def __contains__(self, value):
-            return getattr(self._parent, value).kind & self._kind
+            return self.__internal_list().__contains__(value)
 
         def __iter__(self):
             yield from self.__internal_list()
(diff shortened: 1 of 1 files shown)


ophyd/device.py
```diff
@@ -1274,7 +1274,7 @@ def remove(self, value):
             getattr(self._parent, value).kind &= ~self._remove_kind
 
         def __contains__(self, value):
-            return getattr(self._parent, value).kind & self._kind
+            return self.__internal_list().__contains__(value)
```
tacaswell: Might as well write this as `value in self.__internal_list()`

ZLLentz: Probably correct for readability, I'll swap it

tacaswell:
I was going to say `getattr(obj, name).kind & Kind.normal` but that does not check that all of it's parents are in the read tree (and actually the current implementation does not either)!

Despite what we were originally thinking, I have come around on read_attrs/configuration_attrs being super useful interfaces for managing the state of the device and am against removing them (but think distributing the state like this is a better option).

Somewhat related, we may want to do things like cache the `__interanl_list()` and nuke it when ever the state of any of it's children change, but that is a second order thing..

danielballan:
Does this fix the problem that @teddyrendahl reported on Slack, or is that a separate issue?

teddyrendahl:
Nope. I'll make a separate issue.

PR #575, opened 2018-06-14
Added a new child class of PseudoMotor that works with discrete states

awalter-bnl:
This is an alternative approach to the PreDefinedPositions Class shown in PR #572 . The two options have been done at the request of @danielballan to compare the two approaches.

This has not been tested, and still needs some work but does show how this would be implemented. The two issues that have not been resolved in this approach are:

- I have not yet got a good solution for allowing the device to be in more than one 'state' at once, which from an implementation point of view it is possible to be in more than one state at a time.
- I still don;t have a good solution for adding components to more than one of these 'devices'.

### DIFF
--- a/ophyd/pseudopos.py
+++ b/ophyd/pseudopos.py
@@ -1,5 +1,3 @@
-# vi: ts=4 sw=4
-
 
 import logging
 import time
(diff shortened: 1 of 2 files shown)


ophyd/pseudopos.py
```diff
@@ -914,3 +912,131 @@ def set(self, position, **kwargs):
...
+    NOTES ON PREDEFINED MOTION WITH NEIGHBOURS:
+    1. To ensure the motion to a predefined location always occurs when using neighbours to define
+        motion 'paths' it is best to ensure that the device is always in a 'location' by making sure
+        that motion can not move the device outside of all 'locations'.
+    2. To ensure that motion occurs always via a path then each 'point' in the path should be a 
+        location, and it should only have the neighbours that are before or after it in the required
+        path.
+
+    '''
+    def __init__(self, *,locations={}, neighbours={}, regions={}):
```
danielballan: Should this just take in a `DiGraph`? That seems maximally flexible.

awalter-bnl: I am very reluctant to have it take in a DiGraph, because to non-coding background BL staff that is a confusing way to specify things. I tried at SIX and both BL staff there where completely confused with the structure of DiGraph and how to implement one. This set of dictionaries, they claim, was much easier for them to understand and modify when necessary. When discussing it with Eli and Klaus at ISS they also much preferred this system, as they found it easy to understand (although they would like an extra abstraction so that it takes in an excel file).

Of course I could write an 'external' function that takes in these dictionaries and returns a DiGraph, which is what happens internally at present. A second function could be written to read in an excel file and return the DiGraph to give options for how to specify the interactions.

awalter-bnl:
I have pushed some commits to show how a combined version of the two approaches (previous commits to this PR and to PR [#572](https://github.com/NSLS-II/ophyd/pull/572) might work.
**_This is not tested and shared to show the structure only_**
Once we have decided on a general structure then I will move onto the details of filling in the blanks and testing it to get it working.

My thoughts are that this approach does meet the initial requirements that led to the PR, although I personally don't yet see the advantage of splitting this into 2 sections, 1 for the 'state' work and one as an 'orchestrator' of everything and providing the 'path' work. There is significant overlap between the two, with the only difference really being the addition of the path section and the ability to add other 'PreDefinedPosition' elements to the 'orchestrator'.

The only issue I see with this approach is that I do not see how to make the motion of the 'Pseudo' component (with all the real device motion) and the 'orchestrator' components parallel. at present they occur in series.

What are everyones thoughts on this?

awalter-bnl:
I have moved the import of matplotlib to avoid an ophyd dependency based on a comment from @danielballan in PR #596

PR #582, opened 2018-06-28
Immutable component names

### LINKED ISSUE #581: component_names should be immutable
should things like `pilatus2M.component_names.remove("trans1")`  be possible etc?

@danielballan

tacaswell:
Closes #581.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -417,7 +417,7 @@ class ComponentMeta(type):
             cpt.attr = cpt_attr
 
         # List Signal attribute names.
-        clsobj.component_names = list(clsobj._sig_attrs.keys())
+        clsobj.component_names = tuple(clsobj._sig_attrs.keys())
 
         # The namedtuple associated with the device
         clsobj._device_tuple = namedtuple(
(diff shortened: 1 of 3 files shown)


mrakitin:
Should the description be changes in the [docs](https://github.com/NSLS-II/ophyd/blob/6877e1c70a582049eca9fef7518bf853a994c03c/doc/source/device-overview.rst#classdevice) too? In particular:
```
:attr:`component_names` | List of the names components on this device. Direct children only
```

danielballan:
I take that usage of "List" to be the English word not the Python data structure. I think it's probably more readable/accessible that `tuple`, and the distinction is not important in that context.

tacaswell:
@danielballan then failures look like caproto bugs....

danielballan:
@tacaswell Good catch. Opened issue: https://github.com/NSLS-II/caproto/issues/328

mrakitin:
OK, that's clear, thanks @danielballan. Yes, I also noticed the consistently failing tests. Great the issue is uncovered!

PR #583, opened 2018-06-28
BLD: update for new python version

tacaswell:


setup.py
```diff
@@ -5,15 +5,16 @@
...
       license='BSD',
       install_requires=requirements,
-      packages=find_packages())
+      packages=find_packages(),
+      classifiers=[
+        "Development Status :: 5 - Production/Stable",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+      ])
```
mrakitin: @tacaswell, since you already touched `setup.py`, do you mind to also update it to add proper description to the PyPI page? Please see https://github.com/NSLS-II/databroker/pull/369 for more details. The changes would be:
```py
# Get the long description from the README file
with open(path.join(here, 'README.md'), encoding='utf-8') as f:
long_description = f.read()

setup(
    ...
    long_description=long_description,
    long_description_content_type='text/markdown',
    ...
)
```

klauer:
@mrakitin once tests pass, mind giving this another look? Should resolve your comment above.

(... look at all those glorious force-pushes as I upgraded from trusty -> xenial!)

setup.py
mrakitin: Looks good to me now! Thanks @klauer for the reminder :)

klauer:
Thanks @mrakitin !

.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: python
-sudo: required
-python:
-  - 3.6
-dist: trusty
+dist: xenial
```
mrakitin: 👍

PR #584, opened 2018-06-28
ENH: add config and ommitted components to SynAxis classes

tacaswell:
These don't relate to the delay function of the axis at all but
are useful to exercise all the kinds.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -269,8 +269,14 @@ class SynAxisNoHints(Device):
         used for ``subscribe`` updates; uses ``asyncio.get_event_loop()`` if
         unspecified
     """
-    readback = Component(ReadbackSignal, value=None)
-    setpoint = Component(SetpointSignal, value=None)
+    readback = Component(ReadbackSignal, value=None, kind='hinted')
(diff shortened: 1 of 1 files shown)


awalter-bnl:
Can we also add a 'settle_time' component? This is required for the time estimation?

tacaswell:
real motors do not have settle_time as a component either, it looks like it is a property on `PositionerBase`.  We should probably clean that inconsistency up first.

awalter-bnl:
OK, so if we don't think it is necessary to ever add a 'settle_time' to a motor movement I can remove that from the time estimate. It is true that the same outcome can be done by using 'fixed_mode' on the detector and then setting acquisition_period = acquisition_time + settle_time.

tacaswell:
Sorry, what I meant is that we need to sort out a way to promote 'settle_time' to be a component on `PositionerBase` (rather than a property) to support the time estimation.  We should get this fixed on the 'mainline' part of the codebase before we add it to the simulators.

teddyrendahl:
This makes nicer Typhon screens!

tacaswell:
@teddyrendahl that may or may not be what motivated me to add these....

PR #589, opened 2018-07-10
ENH: Add enum spoofing to FakeEpicsSignal, fix some bugs

ZLLentz:
After submitting the `FakeEpicsSignal` and `make_fake_device` code a couple months ago, I used it to clean up the tests for the SLAC devices library, but noticed some things were missing.

This adds:
- `sim_set_enum_strs` so that you can get/set using the enum string as a real PV would behave
- some checks `EpicsSignal` was doing but this class was not
- `EpicsSignalWithRBV` to the base cases so that fake area detector classes are built correctly
- relevant tests for the above

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -14,6 +14,7 @@ import copy
 import logging
 
 from .signal import Signal, EpicsSignal, EpicsSignalRO
+from .areadetector.base import EpicsSignalWithRBV
 from .status import DeviceStatus, StatusBase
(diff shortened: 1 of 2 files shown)


danielballan:
Thanks for the nudge. This was held up by our broken CI. Now 3/4 builds pass the the failures look unrelated to this change.

Sorry for the hold-up @ZLLentz!

ZLLentz:
No worries @danielballan, I had actually forgotten about this PR

PR #590, opened 2018-07-12
MAINT: Signal should describe itself

### LINKED ISSUE #574: make Signal respect its actual data type
Trying to use soft-signals for stashing some data (that was strings) which I thought was a bug in bluesky (https://github.com/NSLS-II/bluesky/issues/1053).

The issue is that we hard-code the type to be 'number'

https://github.com/NSLS-II/ophyd/blob/224285eacc1d4c7eb913142ff4b39ebdeb092861/ophyd/signal.py#L214-L218

teddyrendahl:
## Description
Currently `ophyd` assumes you have a single floating point number as the value. This is not necessarily the case as `Signal` will be whatever you tell it.

This code simply mirrors the `data_type` and `data_shape` methods used in `EpicsSignal`.

### The Catch
`Signal` initializes itself as `None`. `None` is not a valid `data_type` therefore:

```python
signal = ophyd.Signal('signal')
signal.describe()
```
Raises an error. This same situation is true if the user puts any unacceptable data type into the signal.

## Motivation
Downstream clients like Typhon expect that `Signal` has an accurate description of itself. Currently if a user places a string in a `Signal` it tells us it is a number which then breaks when we convert it for display.

Closes #574

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -213,9 +213,10 @@ class Signal(OphydObject):
 
     def describe(self):
         """Return the description as a dictionary"""
+        val = self.value
         return {self.name: {'source': 'SIM:{}'.format(self.name),
-                            'dtype': 'number',
-                            'shape': []}}
+                            'dtype': data_type(val),
+                            'shape': data_shape(val)}}
 
     def read_configuration(self):
         "Subclasses may customize this."
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -52,7 +52,7 @@ class Signal(OphydObject):
     SUB_VALUE = 'value'
     _default_sub = SUB_VALUE
 
-    def __init__(self, *, name, value=None, timestamp=None, parent=None,
+    def __init__(self, *, name, value=0., timestamp=None, parent=None,
                  labels=None,
                  kind=Kind.hinted, tolerance=None, rtolerance=None, cl=None):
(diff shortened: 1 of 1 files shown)


danielballan:
I am surprised that this has not come up yet. Is this just a mistake? Do @dchabot, @klauer, @tacaswell remember any reason we would have been inconsistent about this on purpose?

klauer:
According to `git log -p ophyd/signal.py`, that `describe` implementation has been there for 3 years - so it doesn't appear to be a regression.

My only guess is that we had intended `Signal` to be more of a base class which people would subclass to make their own kind of signal, or use `EpicsSignal`, `DerivedSignal`, etc. That said, we definitely used them as basic "soft" signals over the years, so this was definitely unintentional/a bug.

tacaswell:
I _did_ write that code, but it has not changed from 2015.  It is `EpicsSignal` that is currently using helpers.

teddyrendahl:
Been using this branch for a few days, it is more annoying than I thought that `Signal(name='a').describe()` raises an error. There are two options:

1. `Signal` instantiates itself with a valid data type i.e `0`
2. We catch the exception and default to the old version of describe if `self.value == None`

Thoughts? Leaning towards option 1 myself.

tacaswell:
Lets go with option 1

danielballan:
All test failures are caproto-related. Need to fix the caproto bug and re-release caproto. I'm on it....

PR #592, opened 2018-07-26
DOC: fix DDC docstring + naming convention of python descriptor

tacaswell:


tacaswell:
Issues I noticed talking through this with @mrakitin .

ophyd/device.py
```diff
@@ -353,7 +353,7 @@ def __get__(self, instance, owner):
 
         return instance._signals[self.attr]
 
-    def __set__(self, instance, owner):
+    def __set__(self, instance, value):
```
danielballan: FYI, confirmed here: https://docs.python.org/3/howto/descriptor.html#descriptor-protocol

PR #596, opened 2018-08-09
Adding a method to ADBase that generates a plot of the asyn diGraph

awalter-bnl:
This PR adds a method to ADBase that generates a plot of the asyn port diGraph using `networkx.draw_networkx` The new method takes in all `*args` and `**kwargs` that are available in  `draw_networkx` and  pases them through. It finds the diGraph using self.get_asyn_diGraph().

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -270,7 +270,7 @@ class ADBase(Device):
             plt.tick_params(axis='y', which='both', left=False, right=False,
                             labelbottom=False)
 
-        nx.draw_networkx(G, ax, *args, **kwargs)
+        nx.draw_networkx(G, ax=ax, *args, **kwargs)
 
     def validate_asyn_ports(self):
         '''Validate that all components of pipeline are known
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -4,6 +4,7 @@ import re
 import sys
 from collections import OrderedDict
 import networkx as nx
+import matplotlib.pyplot as plt
 
 from ..signal import EpicsSignal
 from . import docs
@@ -239,6 +240,30 @@ class ADBase(Device):
                 G.add_edge(in_port, out_port)
 
         return G, port_map
+
+    def visualize_asyn_digraph(self, *args, **kwargs):
+        '''This generates a figure showing the current asyn port layout.
+
+        This method generates a plot showing all of the currently enabled
+        Areadetector plugin asyn ports and their relationships. The current
+        ports and relationships are found using self.get_asyn_digraph.
+
+        Parameters
+        ----------
+        *args, **kwargs : networkx.draw_networkx args and kwargs.
+            For the allowed args and kwargs see the `networkx.draw_networkx documentation <https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html>`
+        '''
+        # Generate the port_map Digraph.
+        G, port_map = self.get_asyn_digraph(self)
+        # Create and label the figure.
+        plt.figure('AD port map for {}'.format(self.name))
+        # Add the plot to the figure.
+        nx.draw_networkx(G, *args, **kwargs)
+        # Turn off the axis ticks and axis labels.
+        plt.tick_params(axis='x', which='both', bottom=False, top=False,
+                        labelbottom=False)
+        plt.tick_params(axis='y', which='both', left=False, right=False,
+                        labelbottom=False)
 
     def validate_asyn_ports(self):
         '''Validate that all components of pipeline are known
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -250,9 +250,7 @@ class ADBase(Device):
         Parameters
         ----------
         *args, **kwargs : networkx.draw_networkx args and kwargs.
-            For the allowed args and kwargs see the `networkx.draw_networkx
-documentation
-<https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html>`_
+            For the allowed args and kwargs see the `networkx.draw_networkx documentation <https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html>`_
         '''
         # Importing matplotlib.pyplot here as it is not a dependency except for
         # this method.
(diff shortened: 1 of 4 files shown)


awalter-bnl:
The Travis build errors appear to be unrelated to this change.

ophyd/areadetector/base.py
```diff
@@ -4,6 +4,7 @@
 import sys
 from collections import OrderedDict
 import networkx as nx
+import matplotlib.pyplot as plt
```
danielballan: Ophyd [does not require matplotlib](https://github.com/NSLS-II/ophyd/blob/master/requirements.txt) so this import would introduce a new dependency. Since matplotlib is quite a heavy dependency and only needed for this one method, I think we should take a special approach: import it inside the method where it is used (i.e. Line 256). This is not usually good practice because it goes against the principle of failing early and clearly and it makes it harder for a developer to immediately see a module's dependencies. But for cases like this, it is a standard approach.

Would you add `matplotlib` to `test-requirements.txt` and add a quick test to [this file](https://github.com/NSLS-II/ophyd/blob/master/ophyd/tests/test_areadetector.py)? It can be a "smoke test" --- a test that just checks that a function runs without errors and doesn't bother to `assert` anything about its output.

awalter-bnl:
Thanks @danielballan for the info, I was struggling to work out how to do a test on this.


In regard to matplotlib, in the PreDefinedPosition PR's there is a very similar method that also uses matplotlib, but even 2 cases is probably not enough to make it a dependency so I have made the change requested. FYI matplotlib was already a test requirement.

danielballan:
> In regard to matplotlib, in the PreDefinedPosition PR's there is a very similar method that also uses matplotlib, but even 2 cases is probably not enough to make it a dependency so I have made the change requested.

Yeah, I think matplotlib is pretty far from ophyd's core function (device abstraction layer) so I would argue that if we start accumulating a lot plotting tools in here that's a sign they should be moved into some special utils module or even separate repo/package.

> FYI matplotlib was already a test requirement.

I don't think in [text-requirements.txt on the master branch](https://github.com/NSLS-II/ophyd/blob/master/test-requirements.txt) or in this PR. Where are you looking?

ophyd/areadetector/base.py
```diff
@@ -240,6 +240,34 @@ def get_asyn_digraph(self):
...
+        '''This generates a figure showing the current asyn port layout.
+
+        This method generates a plot showing all of the currently enabled
+        Areadetector plugin asyn ports and their relationships. The current
+        ports and relationships are found using self.get_asyn_digraph.
+
+        Parameters
+        ----------
+        *args, **kwargs : networkx.draw_networkx args and kwargs.
+            For the allowed args and kwargs see the `networkx.draw_networkx documentation <https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html>`
```
danielballan: This won't format as a link unless you add a `_` after the closing backtick. See [this section of this Sphinx docs](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links). I sympathize: the external link syntax took me forever to commit to memory.

ophyd/areadetector/base.py
```diff
@@ -250,7 +250,9 @@ def visualize_asyn_digraph(self, *args, **kwargs):
         Parameters
         ----------
         *args, **kwargs : networkx.draw_networkx args and kwargs.
-            For the allowed args and kwargs see the `networkx.draw_networkx documentation <https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html>`
+            For the allowed args and kwargs see the `networkx.draw_networkx
```
danielballan: Looks like your editor "helpfully" broke this one line into three lines for you.

awalter-bnl:
Thanks for the comments @danielballan I have made the changes

danielballan:
Are test failures just caproto-related? If I was being super-finicky I would mention that the requirements file is alphabetically sorted, but I wouldn't block merging over that. :-P

awalter-bnl:
I think I have resolved the issue with one of the tests (the others where related to caproto). I have also moved matplotlib up the requirements list as I had to make a change anyway.

tacaswell:
The added test is failing:

```
ophyd/tests/test_areadetector.py::test_visualize_asyn_digraph_smoke[pyepics] FAILED
```

ophyd/areadetector/base.py
```diff
@@ -240,6 +240,34 @@ def get_asyn_digraph(self):
...
+            For the allowed args and kwargs see the `networkx.draw_networkx documentation <https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html>`_
+        '''
+        # Importing matplotlib.pyplot here as it is not a dependency except for
+        # this method.
+        import matplotlib.pyplot as plt
+
+        # Generate the port_map Digraph.
+        G, port_map = self.get_asyn_digraph()
+        # Create and label the figure.
+        plt.figure('AD port map for {}'.format(self.name))
```
tacaswell:
```python
fig, ax = plt.subpplots(num='...')
```
and then passing the `ax` into the `draw_networkx` call is a bit better. 

Having `ax=None` in the signature and then only doing this logic if the user did not pass an axes would also be a good idea.

tacaswell: It does not rely on the global state in pyplot to make sure the graph goes where you want it to.

tacaswell:
ah, it is trying to use the qt backend on travis where there is no X server.

Add `export MPLBACKEND=agg` to the travis file.

awalter-bnl:
Made the changes requested by @tacaswell and hopefully resolved the Travis build error

ophyd/areadetector/base.py
```diff
@@ -240,6 +240,38 @@ def get_asyn_digraph(self):
...
+        # Create and label the figure if no ax is provided.
+        if not ax:
+            fig, ax = plt.subplots()
+            ax.set_title('AD port map for {}'.format(self.name))
+            plt.tick_params(axis='x', which='both', bottom=False, top=False,
+                            labelbottom=False)
+            plt.tick_params(axis='y', which='both', left=False, right=False,
+                            labelbottom=False)
+
+        nx.draw_networkx(G, ax, *args, **kwargs)
```
tacaswell: Needs to be `ax=ax`

awalter-bnl:
So now I think we only have the caproto related errors on 3 of the 4 Travis builds. It passes one of them.

Issue #598, opened 2018-08-17
Check that AD file plugins are "warmed up" during stage()


prjemian:
This is a consistent pattern:

* start bluesky (ipython, notebook, whatever)
* connect with area detector AND its HDF5 file writer plugin
* reboot area detector IOC
* try areadetector.trigger()

This will fail with `TimeoutError` on the `$(P):HDF1:Capture_RBV`

The fix for this is to always, *first* after rebooting the IOC, acquire a frame and save it from the HDF5 file writer plugin.

the order of ophyd connecting with the detector and the IOC reboot can be swapped.  Start with the IOC reboot, then start the bluesky session and connect with the detector.  Then trigger.  Same problem.  

Happens with SimDetector, PointGrey, Apogee Alta, and other detectors.

prjemian:
Here's the exception trace from today:

```
[user2bmb@lyra,42,startup]$ use_bluesky.sh 2bmb
BlueSky executable path: /home/beams/USER2BMB/Apps/BlueSky/bin
Python 3.6.5 :: Anaconda, Inc.
conda 4.5.8
pip 10.0.1 from /home/beams/USER2BMB/Apps/BlueSky/lib/python3.6/site-packages/pip (python 3.6)
starting ipython with profile: 2bmb
Python 3.6.5 |Anaconda, Inc.| (default, Apr 29 2018, 16:14:56) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: 2bmb
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/00-0-checks.py
BlueSky version: 1.3.3
Ophyd version: 1.2.0
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/00-startup.py
Loading metadata history from /home/beams/USER2BMB/.config/bluesky/bluesky_history.db
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/01-databroker.py
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/02-pyepics.py
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/10-imports.py
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/11-motors.py
/home/beams8/USER2BMB/.ipython/profile_2bmb/startup/15-custom-devices.py
[...]

PR #599, opened 2018-08-22
ENH: Try to resolve errors due to HDF5 warmup automatically.

danielballan:
A quick attempt at resolving #598, dashed off during a conference.

@prjemian, do you have time to test this?

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -11,6 +11,7 @@ import re
 import time as ttime
 import logging
 from collections import OrderedDict
+import warnings
 import numpy as np
 
 from ophyd import Component as Cpt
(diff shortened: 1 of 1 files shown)


tacaswell:
While practical, I am uncomfortable with this.  I don't think we understand the underlying problem well enough to say that in all cases if staging times out we should try exactly once to run the warmup and then try again.  Are we sure that if `stage` has failed `unstage` will not also fail?  If stage is what is crashing the IOC this just delays the final issue.  Are we sure that hdf5 is the only plugin with this problem?

I think it would be better to add a call to warmup to the `__init__` method (not sure if putting it in the hdf5 plugin or in the top-level AD devices is better) and if we get a timeout during stage re-raise with a clearer exception method (which is how I read @prjemian's comment in #598).

tacaswell:
> Hmm. Not sure I like the idea that init would cause the detector to acquire though.

That was my compromise on not liking `stage` to acquire.  On board with warm-up being a fully explicit activity and this error message just telling the user they need to do it.

This may be a problem for long un-attended operation, but if this is becoming an issue then that means your AD IOC is restarting randomly which is probably a sign of a deeper problem.

Maybe raise a custom Exception sub-class so plans that _do_ want to auto-warmup can?

prjemian:
Random ioc restarts?  Hardly the case.  We have a lot of IOCs.  People
reboot them by their choice.  We see this in fresh IOCs (after they
restart).  Also, sometimes they crash, needing a reboot.  People forget or
overlook this step, though.

We must be prepared to identify this.

danielballan:
> We must be prepared to identify this.

Yep, this aligns with my feeling that biggest yet-unfulfilled potential for ophyd is more informative errors. But we should generally resist magically fixing things and thus masking deeper problems (as this PR of my does...).

hhslepicka:
@teddyrendahl In any case you get the connection callback so you would be covered even in the edge case that you mentioned.

```python
In [1]: from epics import PV

In [2]: def conn_cb(conn, *args, **kwargs):
   ...:     print("Connection Changed to: ", conn)
   ...:

In [3]: def val_cb(value, *args, **kwargs):
   ...:     print("PV changed to: ", value)
   ...:

In [4]: pv = PV('MTEST:Float', connection_callback=conn_cb, callback=val_cb)

In [5]: Connection Changed to:  True
PV changed to:  0.0
```

PR #600, opened 2018-08-29
ENH: add classes for handling AD 33 IOCS

tacaswell:
- add v33_cam_mixin to add `wait_for_plugins` component and configure
   the stage sigs for parallel use
 - add SingleTrigger33 which assumes that `wait_for_plugins` is
   configured correctly
 - mixin classes to bring a a few other 33 pvs

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -5,10 +5,17 @@ import sys
 from collections import OrderedDict
 import networkx as nx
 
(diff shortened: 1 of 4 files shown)


tacaswell:
See https://github.com/NSLS-II-FXI/profile_collection/commit/fce334b4159d5ad844906be65ef996df86160b42 for these used live.

awalter-bnl:
This looks good to me

PR #602, opened 2018-08-31
FIX: Follow up #600

danielballan:
https://github.com/NSLS-II/ophyd/pull/600 went in before I got a chance to look
at it. I think I found some bugs, along with some requisite name-standardizing.

One thing I didn't go in and change but might be worth quickly considering:

We have classes named ``SingleTrigger33`` and ``V33CamMixin``. Any thoughts on
the best consistent convention for naming version-specific mixin classes? Also
--- as of df20cad in this PR --- we also have some ``V26`` mixins.
Should ``V33`` classes inherit from those? The upcoming ophyd release sets a
precedent in the ophyd API for how AD versions will be handled and deserves
some attention.

### DIFF
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -48,6 +48,8 @@ class V33CamMixin(V33Mixin):
         self.stage_sigs['wait_for_plugins'] = 'Yes'
         for c in self.parent.component_names:
             cpt = getattr(self, c)
+            if cpt is self:
+                continue
             if hasattr(cpt, 'ensure_nonblocking'):
                 cpt.ensure_nonblocking()
 
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -36,7 +36,7 @@ __all__ = ['CamBase',
            ]
 
 
-class V33CamMixin(V33Mixin):
+class CamV33Mixin(V33Mixin):
     wait_for_plugins = C(EpicsSignal, 'WaitForPlugins',
                          string=True, kind='config')
 
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -46,7 +46,7 @@ class V33CamMixin(V33Mixin):
 
     def ensure_nonblocking(self):
         self.stage_sigs['wait_for_plugins'] = 'Yes'
-        for c in self.component_names:
+        for c in self.parent.component_names:
             cpt = getattr(self, c)
             if hasattr(cpt, 'ensure_nonblocking'):
                 cpt.ensure_nonblocking()
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -3,7 +3,7 @@ import logging
 from ..utils import enum
 from .base import (ADBase, ADComponent as C, ad_group,
                    EpicsSignalWithRBV as SignalWithRBV,
-                   v33_mixin)
+                   V33Mixin)
 from ..signal import (EpicsSignalRO, EpicsSignal)
 from ..device import DynamicDeviceComponent as DDC
 
@@ -36,7 +36,7 @@ __all__ = ['CamBase',
            ]
 
 
-class v33_cam_mixin(v33_mixin):
+class V33CamMixin(V33Mixin):
     wait_for_plugins = C(EpicsSignal, 'WaitForPlugins',
                          string=True, kind='config')
 
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -11,11 +11,18 @@ from ..device import (Device, Component as Cpt)
 from ..signal import (ArrayAttributeSignal)
 
 
-class V26Mixin(Device):
+class V22Mixin(Device):
+    ...
+
+
+class V26Mixin(V22Mixin):
     adcore_version = Cpt(EpicsSignalRO, 'ADCoreVersion_RBV',
                          string=True, kind='config')
     driver_version = Cpt(EpicsSignalRO, 'DriverVersion_RBV',
                          string=True, kind='config')
+
+class V33Mixin(V26Mixin):
+    ...
 
 
 class EpicsSignalWithRBV(EpicsSignal):
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -11,7 +11,7 @@ from ..device import (Device, Component as Cpt)
 from ..signal import (ArrayAttributeSignal)
 
 
-class V33Mixin(Device):
+class V26Mixin(Device):
     adcore_version = Cpt(EpicsSignalRO, 'ADCoreVersion_RBV',
                          string=True, kind='config')
     driver_version = Cpt(EpicsSignalRO, 'DriverVersion_RBV',
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -11,7 +11,7 @@ from ..device import (Device, Component as Cpt)
 from ..signal import (ArrayAttributeSignal)
 
 
-class v33_mixin(Device):
+class V33Mixin(Device):
     adcore_version = Cpt(EpicsSignalRO, 'ADCoreVersion_RBV',
                          string=True, kind='config')
     driver_version = Cpt(EpicsSignalRO, 'DriverVersion_RBV',
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -34,7 +34,7 @@ from ..device import (GenerateDatumInterface, BlueskyInterface, Staged,
                       Component as Cpt)
 from ..signal import EpicsSignal
 from ..utils import set_and_wait
-from .base import v33_mixin
+from .base import V33Mixin
 
 logger = logging.getLogger(__name__)
 
@@ -103,11 +103,6 @@ def resource_factory(spec, root, resource_path, resource_kwargs,
         return datum
 
     return resource_doc, datum_factory
-
-
-class v33_file_mixin(v33_mixin):
-    create_directories = Cpt(EpicsSignal,
-                             'CreateDirectory', kind='config')
 
 
 class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -34,7 +34,6 @@ from ..device import (GenerateDatumInterface, BlueskyInterface, Staged,
                       Component as Cpt)
 from ..signal import EpicsSignal
 from ..utils import set_and_wait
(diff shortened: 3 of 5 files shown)


awalter-bnl:
sorry @danielballan, I will try to be a bit less trigger happy with the merge button !-).

ophyd/areadetector/cam.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, *args, **kwargs):
 
     def ensure_nonblocking(self):
         self.stage_sigs['wait_for_plugins'] = 'Yes'
-        for c in self.component_names:
+        for c in self.parent.component_names:
```
tacaswell: This is going to infinitely recurse.

danielballan: Fix pushed.

danielballan:
@tacaswell Thanks for the code review. Re-upping my questions about how organize and name version mixins. If we don't have time to get the names sorted I would rather revert #600 than merge not being sure of our version-handling convention. This is only going to get more complex over time, and AD is a ball of class spaghetti as it is....

ophyd/areadetector/cam.py
```diff
@@ -46,7 +46,9 @@ def __init__(self, *args, **kwargs):
...
+            if c is self:
```
tacaswell: `c` here is a string so this will never be true.

tacaswell:
I do not care on the names.

We are going to have a number of beamlines moving to AD33 this cycle, if we revert #600 we will up with this code vendoreded a bunch of places which has the same problems but worse.

ophyd/areadetector/cam.py
danielballan: force-pushed

danielballan:
Anybody up for engaging on the names? Seems important, as eventually _many_ objects that currently have simple names like `SingleTrigger` will have names like:

* `SingleTrigger33`
* `SingleTriggerV33`
* `V33SingleTrigger`  <-- has disadvantage that it doesn't alphabetically sort near `SingleTrigger`

which seems like a step down in comprehensibility and even more so if we are inconsistent about it.

And with `CamV33Mixin`, why not make it `CamV33` inheriting from `Cam`? Is that better or worse, and why? There are couple ways to do it and the implications are worth thinking through.

> We are going to have a number of beamlines moving to AD33 this cycle, if we revert #600 we will up with this code vendoreded a bunch of places which has the same problems but worse.

Disagree it would be worse, because it would isolated to local profiles, not imposed as an API in a stable library. We could also stuff this into `nslsii` as a stop-gap. That might be the best way forward if we are deadlocked on this.

awalter-bnl:
I tend to agree with Dan that we should consider how we structure these version specific classes. My gut reaction is that each should be a child of the previous version. As for naming we should have a standard, but I am not yet convinced what the standard should be.

danielballan:
Had a quick call with @tacaswell who would very much like to tag ophyd today with #600 and this in it. I'm going to take a pass at making the names at least internally consistent. For the record I would prefer this go into `nslsii` and be hashed out here more slowly, but I won't block merging.

tacaswell:
I am in favor of merging these with whatever names they have and then cleaning it up later.

The whole AD fleet of classes probably needs an overhaul and re-think (and in general we need to think about how to best organize ophyd objects that need to support multiple versions of the underlying IOC that may have changed PVs / behavior).  Long term we need input from @teddyrendahl @ZLLentz and @prjemian.

Moving these to nslsii just feels like it is adding epicycles and complexity (when we sort this out profiles are going to have to be changed anyway unless we keep the AD tree in two places....).

awalter-bnl:
After thinking about it more I am thinking placing it in nsls-ii is probably the best path forward for now. Then we can take the time to decide on the appropriate structure when things aren’t so hectic

prjemian:
not ready to test AD3.3 features

klauer:
PV naming backward-compatibility for most AD versions < 3 made this a non-issue, largely. 

Is it worth considering that `Device` should change to better accommodate these types of version differences? While I always thought that `Device`'s inflexibility (component immutability?) was one of its strong points, there have been changes with the addition of hints, suggestions to remove the `DeviceTuple`s/get/put, and so on that seem to indicate a general desire to go in a different direction. 

What would it look like to have one device capable of supporting different versions (kwarg, dict mapping, lists, ...)? If not that, it seems it would be necessary to allow subclassed devices to remove superclass components (to support removed PVs after a version bump). Would it be possible to have a 'hint' flag indicating a potentially non-existent component?

Just some random thoughts, in any case.

danielballan:
> If not that, it seems it would be necessary to allow subclassed devices to remove superclass components (to support removed PVs after a version bump).

We currently do this by overriding the relevant attribute with `None`. I think it would be slightly less opaque to use some special sentinel instance named something like `REMOVED_COMPONENT`. In the rare times when AD has _removed_ a PV, it's probably not bad for ophyd to carry that name anyway and make it clear that it is no longer exant.

tacaswell:
Replaced by https://github.com/NSLS-II/nslsii/pull/34 and #603

tacaswell:
1) Correct, the `wait_for_plugins` is way better (as it lets you use all your cores and not block framing on file writing etc), but there are other changes to AD that need to be addressed (ex https://github.com/NSLS-II/nslsii/pull/36)

2) things like the 'create directories' PV came in in 2-something but we don't support uniformly yet.

I think I am also in favor of moving the area detector related classes out of main ophyd (and maybe move it to the AD organization?).   I don't think dropping support for older versions of the IOCs is an option.

More generically, I think we need to re-organize ohpyd / nslsii / ophyd-AD / ... to be better coupled / version to the underlying IOCs (does EpcisMotor ever change? Do we need versioned classes for all of the synapps IOCs?).

danielballan:
OK, How about this for a roadmap.

1. Rip areadetector and its tests out of ophyd and into a separate package. Replace it with a shim import (at least for now) for backward-compatibility.
2. Move that separate package into the AreaDetector GH organization if Mark thinks that is sensible.
3. Improve our test stand to run against a range of AD versions, including at least one 1.9.x one (which will likely be at some NSLS-II beamlines for some time yet) at the latest release of ADCore.
4. Make a public API like this:
   ```py
   from areadetector.v33 import SingleTrigger, AreaDetector  # targets a specific release
   from areadetector import SingleTrigger, AreaDetector  # an alias for the latest release
   ```
   That will involve a lot of shimming -- many classes will be unchanged between versions and therefore just aliased -- but I like that it makes version-bumping a configuration file into a one-line change to `vXX`.

teddyrendahl:
@danielballan I agree on the roadmap. From a SLAC perspective, I would say this is however a low priority. (We are a while away from `v3.3`). However, maybe there are helpful things we can had to `2.x` classes that will be useful as mentioned by @tacaswell 

> More generically, I think we need to re-organize ohpyd / nslsii / ophyd-AD / ... to be better 
coupled / version to the underlying IOCs (does EpcisMotor ever change? Do we need versioned classes for all of the synapps IOCs?).

I would like AD to be the exception rather than the rule. I don't think this kind of thing is necessary for things like `EpicsMotor`. Let's say that the newest version of `EpicsMotor` has the latest and greatest record that you absolutely can't live without. It is extremely trivial to subclass and add a signal. Underneath you still probably will wait for `DMOV`in the same way. `AD` is different because of the complications in maintaining the state and dependency chain of all your plugins that you would never want another user to re-write.

prjemian:
I'm not happy moving this ophyd support into the EPICS-based AreaDetector organization.  That makes it the exception in that org.  Fragmented code.  Different style of repo that the others.

danielballan:
> What are the chances that areaDetector will make a significant break in functionality again? It will be so much work and maintenance to separate the functionality for what will likely be split into simply: < 3.3 and >= 3.3 for many years (if not for the life of areaDetector itself).

There are other (less important) differences between the versions that we have been glossing over or patching clumsily at the IPython profile level. Whether or not we make a separate package, I think something like (4) above is necessary.

> I would like AD to be the exception rather than the rule. 

Agreed. I think there is value in making ophyd a "one stop shop" where you can go to look if there is a bluesky interface to X. By default, anything generally useful should be put there.

However, to parry @klauer a bit, good-naturedly:

* As Teddy pointed out, the release cycles are de-coupled. We really ought to have a new `areadetector` for each ADCore, even if's a trivial update to add a couple new PVs.
* I think separating the test beds will actually make our lives easier. We don't need to run the full ophyd test suite against every AD version, for example. And, thinking of our recent experience with caproto, isolating AD-specific tests to separate builds will reduce the expectation value of random failures per a given job.
* Consider the stage of development. It would have been annoying to co-develop `areadetector` and `ophyd` as separate packages while the interfaces between them were a moving target, but less so now.

@prjemian That's a fair counterargument. But I think there could be symbolic, cultural benefits to moving it over, possible improving coordination between the core ophyd devs and the core AD devs, and making maintenance of ophyd--AD integration something that we share with the AD community. But the org it lives in is less important than making it a separate package. As long we can separate it out, whether it lives in the NSLS-II or AD org, we might persuade some ADCore contributors to be co-maintainers, whereas they might be less willing to sign up for the firehose of notifications from ophyd generally.

tacaswell:
> 'm not happy moving this ophyd support into the EPICS-based AreaDetector organization. That makes it the exception in that org. Fragmented code. Different style of repo than the others.

On the other hand AD already contains all of the layouts for the gamut of client screens so it is already not a pure c++ repo.  It fits thematically that the python client code also lives there.   I would be pretty happy if we had `areadetector` be a namespace package, the python code for the `Devices` lived scattered in the AD code base, and it produces an sdist/wheel as part of the epics build chain.

Much of the tricky things around the ophyd.AD code are related to making sure all of the plugins are configured correctly and being sure things are done etc which ends up being very tightly coupled to the details of the implementation which are best known to AD developers.

> I would like AD to be the exception rather than the rule.

If we have 1 that needs it, then we will likely have 2 and we should do them the same.  While practically we may never need `EpicsMotorV2`, we should keep an eye on things other than AD using the scheme.

> It will be so much work and maintenance to separate the functionality for what will likely be split into simply: `< 3.3` and `>= 3.3` for many years (if not for the life of areaDetector itself).

the 'create folders' check box ;)  The new functionality is back compatible, the breaking change is https://github.com/NSLS-II/nslsii/pull/36 which I suspect we will see a hand full of every version.

> More packages, more problems.

To some degree we have already gone to the logical extreme of this were every beamline profile is it's own package (and that is unmanageable) so splitting AD out into it's own package is actually a nett reduction in packages ;)

> Agreed. I think there is value in making ophyd a "one stop shop" where you can go to look if there is a bluesky interface to X. By default, anything generally useful should be put there.

This is dragging up some old arguments with @klauer and @danielballan (about pulling `Device` up into a package called 'gadget' that ophyd depends on), but I am in mildly in favor of pushing all `Device` subclasses out of ohpyd and into stand-alone packages.  This is following the pattern of epics-base it's self ;)

-----

I have now disagreed with everyone else on this thread at least once ;)

PR #609, opened 2018-09-18
ENH: fixes to Signal from working at TES

tacaswell:
- the check in set_and_wait used to fail with
   value error for arrays and the exception was going
   to the debug log
 - at TES we used list(...) to get around this
 - Teddy updated Signal to correctly report it's type,
   but the type sniffer did not correctly handle lists

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -190,8 +190,8 @@ class Signal(OphydObject):
                                value)
                 success = False
             except Exception as ex:
-                self.log.debug('set_and_wait(%r, %s) failed', self.name, value,
-                               exc_info=ex)
+                self.log.exception('set_and_wait(%r, %s) failed',
+                                   self.name, value)
                 success = False
             else:
                 self.log.debug('set_and_wait(%r, %s) succeeded => %s',
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -290,7 +290,7 @@ def _compare_maybe_enum(a, b, enums, atol, rtol):
 
 
 _type_map = {'number': (float, np.floating),
-             'array': (np.ndarray, ),
+             'array': (np.ndarray, list, tuple),
              'string': (str, ),
              'integer': (int, np.integer),
              }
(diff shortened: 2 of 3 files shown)


mrakitin:
Any ideas on having a test for this case?

tacaswell:
Yes, I just wanted to get the code out of my head before I left TES.

mrakitin:
That's where we saw the problem: https://github.com/NSLS-II-TES/profile_collection/commit/cf96189fe8d6dacfae507ed0d617a8b093f8d14a#diff-c675cd8248c0fa1ec9653e62d8ec1ccfL189

And that's where it was temporary fixed: https://github.com/NSLS-II-TES/profile_collection/commit/cf96189fe8d6dacfae507ed0d617a8b093f8d14a#diff-c675cd8248c0fa1ec9653e62d8ec1ccfR143

klauer:
Looks like this is an important one to get in?

ophyd/signal.py
```diff
@@ -164,11 +164,12 @@ def set_thread():
                                value)
                 success = False
             except Exception as ex:
-                self.log.debug('set_and_wait(%r, %s) failed', self.name, value,
-                               exc_info=ex)
+                self.log.exception('set_and_wait(%r, %s) failed',
```
klauer: This turns the `debug` message into an `error` message. Was that intentional or was the desire to just remove `exc_info=ex`?

klauer:
Wrapping the important `data_type` bits into #653.

danielballan:
Thank you, @klauer.

ophyd/signal.py
tacaswell: Mostly driven by number 2, but I think number 1 is also a correct change as `set_and_wait` failing should be a big deal.

ophyd/utils/epics_pvs.py
```diff
@@ -282,11 +282,15 @@ def _compare_maybe_enum(a, b, enums, atol, rtol):
...
+    ret = a == b
+    try:
+        return bool(ret)
+    except ValueError:
+        return np.all(ret)
 
 
 _type_map = {'number': (float, np.floating),
-             'array': (np.ndarray, ),
+             'array': (np.ndarray, list, tuple),
```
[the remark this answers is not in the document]
tacaswell: we also require it to have a concrete length.  Agree a refactor is a good idea, but I think out of scope for current effort.

PR #610, opened 2018-09-19
MAINT: AttributeSignal should run subscriptions

teddyrendahl:
## Description
Noticed this while making a `typhon` widget for an `AttributeSignal`. When you call `.put` on your `AttributeSignal`it never runs it's subscriptions. I realize we won't get updated when the underlying attribute gets swapped out but it is nice to get updated when somebody writes using the `AttributeSignal` interface.

## Question
Was also curious how people felt about the ability to add a little metadata to these signals. (I'd do this in a separate PR, but since you have read this far). The idea is that we might want to add a little extra data to `.describe` but not subclass.

```python
class MyDevice(Device):
        calculated_width = Cpt(AttributeSignal, attr='width_calc', desc={'units': 'mm'})
```

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1007,7 +1007,10 @@ class AttributeSignal(Signal):
         return getattr(self.base, self.attr)
 
     def put(self, value, **kwargs):
-        return setattr(self.base, self.attr, value)
+        old_value = self.get()
+        setattr(self.base, self.attr, value)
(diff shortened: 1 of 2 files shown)


teddyrendahl:
Looks like `caproto` / `areadetector` related test failure, don't think related to this PR

danielballan:
Sorry about that caproto/AD failure... I'm honing in on a fix.

PR #616, opened 2018-10-03
adding a note about the hdf5 fileplugin requirement to the docstring of MultiTrigger

awalter-bnl:
This PR updates the docstring of MultiTrigger to add a note about the requirement to use it with the hdf5 fileplugin. 

## Description
This adds a note to the MultiTrigger doc string to warn that it must be used with the hdf5 plugin, failure to do this means that only the last image in each `device.trigger()` call will be found when runnign the `device.read()` afterwards.

## Motivation and Context
At the NSLS-II LIX beamline they attempted to use this without a fileplugin, and where only able to extract the last image from each trigger sequence from databroker. Investigation revealed that this is because without a fileplugin each succesive image is overwrites the previous one in memory and hence without reading the data between triggers it is not accessible. Further investigation revealed that as the `device.read()` method does not know anything about the `device.trigger()` method it will only return a reference to the final image of the `device.trigger()` call. As the hdf5 fileplugin works in streaming mode (where all files for a single run are saved to the same hdf5 file) then this by default ensures that all data is referenced in databroker.

## How Has This Been Tested?
This is a docstring change only, so no testing has been completed.

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -153,7 +153,7 @@ class MultiTrigger(TriggerBase):
     positions.
 
     NOTE: At present this trigger mixin only works in conjunction with the
-    ``ophyd/areadetector/filestore_mixin/FileStoreHDF5 mixin. Use of other
+    ``ophyd/areadetector/filestore_mixin/FileStoreHDF5`` mixin. Use of other
     ``ophyd/areadetector/filestore_mixins`` or reading the array data using
     `device.read()` after `device.trigger()` will result in only the last image
     being recognised by databroker.
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -152,6 +152,12 @@ class MultiTrigger(TriggerBase):
     acquisitions with, for example, different gain settings or shutter
     positions.
 
+    NOTE: At present this trigger mixin only works in conjunction with the
+    ``ophyd/areadetector/filestore_mixin/FileStoreHDF5 mixin. Use of other
+    ``ophyd/areadetector/filestore_mixins`` or reading the array data using
+    `device.read()` after `device.trigger()` will result in only the last image
+    being recognised by databroker.
+
     The are two levels of nesting here:
 
      - cycling through different actions on successive calls to `trigger`


awalter-bnl:
The errors in the Travis build are unrelated to the current PR as this only changes some doc-strings whereas the errors relate to a caget issue.

ophyd/areadetector/trigger_mixins.py
```diff
@@ -152,6 +152,12 @@ class MultiTrigger(TriggerBase):
     acquisitions with, for example, different gain settings or shutter
     positions.
 
+    NOTE: At present this trigger mixin only works in conjunction with the
+    ``ophyd/areadetector/filestore_mixin/FileStoreHDF5`` mixin. Use of other
```
tacaswell: what hdf5 specific assumptions does this make?

awalter-bnl: It works with hdf5 in steaming mode, as all images are written to a single file which is referenced by data broker.

danielballan: I don't follow.

awalter-bnl: So based on my work over the last few days I think now that I am convinced that it will work for all file_plugins. It just doens't work if no file_plugin is used, as we `'read'` from memory only after the `'trigger'` has completed and only the last acquisition is stored in memory. I will update this accordingly if everyone agrees with that statement.

danielballan: I don't think things are being overwritten. Here's how MultiTrigger works:
1. The RunEngine calls `trigger()` (once!)
2. The MultiTrigger mixin, which implements `trigger()`, flips the acquire bit on the detector _multiple_ times, but it stashes each reading in a separate data_key. This is how we avoid a collision/overwrite.
3. The RunEngine calls `read()`. It gets a dictionary with multiple items, with names like `image_gain_100`, `image_gain_200`, etc., each mapped to a different `datum_id`, corresponding to the different images acquires during that one call to `trigger()`.

What do you mean by `seq_number`? Do you mean `event['seq_num']`? That counter is independent of `point_number`. The happen to correlate in simple cases, but handlers do not rely on this, or even have access to it. They rely on `point_number` only.

awalter-bnl: To avoid getting caught in a detailed discussion about what the observed _issue_ was I will restate the motivation, without any attempt to explain _why_ we see the issue. At least then we can avoid going down a rabbit hole that may or may not be necessary.

> At the NSLS-II LIX beamline they attempted to use this without a file-plugin, and where only able to extract the last image from each trigger sequence from databroker.

I will add to this that I was able to recreate this behaviour using a standard prosilica camera, and requests to extract the data using any but the 'last' data_key resulted in an empty array. In my testing with I found that using the hdf5 fileplugin resolved the issue.

Also for the record, I mistakenly assumed that using other file formats (like tiff) would result in losing data, but subsequent testing showed I was wrong in this assumption. So based on this the warning should be update to state that this does not work _without a file-plugin_  not that it only works with the _hdf5 file-plugin_.

I also agree with this description of how the trigger mixin works:
> 1. The RunEngine calls trigger() (once!)
> 2. The MultiTrigger mixin, which implements trigger(), flips the acquire bit on the detector multiple times, but it stashes each reading in a separate data_key. This is how we avoid a collision/overwrite.
> 3. The RunEngine calls read(). It gets a dictionary with multiple items, with names like image_gain_100, image_gain_200, etc., each mapped to a different datum_id, corresponding to the different images acquires during that one call to trigger().

OK so hopefully we both agree with everything up to this point (at least that is what I think).

My explanation of why I think this doesn't work without a file-plugin is I think where we disagree. So for clarity let me restate it here as clearly as I can.

- In step 2 above, and without a file-plugin, every time `trigger()` is called it is writing the image to RAM (not to a disk, as is done when using a file-plugin). 
-I suspect then that each subsequent `trigger()` overwrites the previous image in RAM. 
- When we finally call `read()` ophyd only gets back a _single image_ from the detector, which is associated with the final data_key in the dictionary that it returns to the Run-Engine, with the rest of the data_keys being associated with empty arrays.

I have not done any testing on my hypothesis on why it doesn't work without a file-plugin, and so I could very well be wrong about it. If we agree that this does not work _without a file-plugin_ then I propose that I update the warning to suit and we move forward with this. If we don't agree that multi-trigger does not work _without a file-plugin_ then I think the best option is for me to formalize my tests and present them for discussion.

klauer:
What should we do with this PR - any resolution to the comments above?

PR #618, opened 2018-10-06
ENH/API: checks on AD plugin 'primed-ness' and access to dotted name

tacaswell:


teddyrendahl:
I really like the `attr_name` and `dotted_name` part of this PR!

danielballan:
Well, @tacaswell, your new error works. Lots of

```
ophyd.areadetector.plugins.UnprimedPlugin: The plugin tiff1 on the area detector with name det has not been primed.
```

in the Travis logs now.

tacaswell:
One API question here for the dotted name, do we want

```
o.a.x.dotted_name == 'a.x'
```

or 

```
o.a.x.dotted_name == '.a.x'
```

I went back and forth (and ended up with code that went one way and tests that went the other!)

The argument for the first (no leading '.') is that you can do `getattr(obj, some_great_grand_child.dotted_name)`.  The argument for the leading '.' is that it is a hint to the user reading the error message that they need to find a parrent to paste it on to (and that is the name that we can not extract from python).

prjemian:
:)

teddyrendahl:
Can we separate this into two PRs? I'd very much like the `attr_name` `dotted_name` will be very useful in `typhon`and #630

klauer:
I'm splitting this into two PRs:
1. AD plugin primed-checks
2. `Device` cleanup PR: dotted attr_names will be absorbed into this

PR #620, opened 2018-10-11
CI: Use epics-on-travis instead of conda or Debian.

danielballan:


danielballan:
Travis did not pick this up....

mrakitin:
What's the motivation?

tacaswell:
Bit more reliable / controlled and faster to install.

klauer:
Travis does _not_ like these changes or this PR for whatever reason.

klauer:
Let's just move to azure eventually...

PR #623, opened 2018-10-12
add a filestore mixin to be used with AD hdf5 plugin in Single mode

awalter-bnl:
## Description:
This PR adds a new filestore mixin that should be used to run an Area Detector HDF5_plugin in `Single` mode instead of `Streaming` mode. The result of this mode is that a new hdf5 file is created for every `trigger`.It also adds a second filestore plugin that mixes into this the `FileStoreIterativeWrite` plugin.

## Motivation:
This was motivated by a custom plugin which has an array with one dimension determined only after running the plugin The result is that each `trigger` then creates a different sized array to the previous one, and so the `Streaming` mode which sets the array size prior to any `trigger`  did not correctly save the data.

## Testing:
This has been tested in the field as submitted here (NSLS-II, SIX beamline, rixscam detector, XIP plugin) and is working there.

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -463,6 +463,35 @@ class FileStoreHDF5(FileStorePluginBase):
         self._generate_resource(res_kwargs)
 
 
+class FileStoreHDF5Single(FileStorePluginBase):
+    '''This FileStore mixin is used when running the AreaDetector hdf5 plugin
+    in `Single` mode (ie. one hdf5 file per trigger).
+
+    '''
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.filestore_spec = 'AD_HDF5_SINGLE'  # spec name stored in res. doc
+
+        self.stage_sigs.update([('file_template', '%s%s_%6.6d.h5'),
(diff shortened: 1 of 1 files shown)


awalter-bnl:
The failure here is in the nightly test and is related to a caproto timeout error. I don't think that the current change touches that.

klauer:
I don't see any major issues with this, except:

> This was motivated by a custom plugin 

Based on that, it seems like this may be better off in the `nslsii` package (or in the SIX beamline config?).
What plugin is it?

awalter-bnl:
Thanks for the comment @klauer. In response:
1. While this was motivated by a custom plugin, that plugin just does some on the fly processing. This custom plugin uses the standard AD-HDF5 file plugin.
2. This filestore mixin works with the standard AD-HDF5 file plugin, but running in 'single' mode (one file per trigger) as opposed to 'streaming mode' (1 file per run) which the current hdf5 filestore mixin works with.
As this works with the standard AD-HDF5 plugin, but in a different mode to the current one, I think it has general appeal. If it is decided that it doesn't we currently already have this in nslsii and deployed.

klauer:
@awalter-bnl OK, I see no other reason to not merge this from my perspective.

Objections? Will merge by Pacific COB if none.

PR #625, opened 2018-10-24
ENH: add instantiate_fake_device, clear_fake_device

klauer:
Adds a couple convenience functions for more easily working with fake devices.

`instantiate_fake_device`: creates a fake device class and instantiates it for you, making its best guess at keyword arguments

`clear_fake_device`: resets all components of a device (and its sub-devices) to specific values. This works around the default of `None` for all signals that causes problems with `.describe()`.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -916,7 +916,7 @@ def make_fake_device(cls):
 
 
 def clear_fake_device(dev, *, default_value=0, default_string_value='',
-                      ignore_exceptions=True):
+                      ignore_exceptions=False):
     '''Clear a fake device by setting all signals to a specific value
 
     Parameters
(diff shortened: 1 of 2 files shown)


ophyd/sim.py
```diff
@@ -913,6 +915,95 @@ def make_fake_device(cls):
...
+
+    Parameters
+    ----------
+    dev : Device
+        The fake device
+    default_value : any, optional
+        The value to put to non-string components
+    default_string_value : any, optional
+        The value to put to components determined to be strings
+    ignore_exceptions : bool, optional
```
teddyrendahl: Why did we need this?

klauer: Are we guaranteed `sim_put` will exist on all faked signals and that it will never raise?

My thought was: broadcast "clear" values where possible, ignoring anything that might fail.
If it's not a concern, I can easily nuke it.

ZLLentz: `sim_put` will exist on all faked signals, but we are not guaranteed it will never raise

klauer: I flipped the default `ignore_exeptions` to `False`, and given that we're guaranteed `sim_put` to exist I'm not including a `hasattr` check.

PR #627, opened 2018-10-24
FIX/TST: Device.put for non-DeviceTuple values

klauer:
Fix the (much-maligned) `.put` functionality of `ophyd.Device` in cases where the user did not use the Device-specific `DeviceTuple`.

I'm a bit surprised this never popped up before, but it was broken for all inputs where there was more than one signal. Incidentally, that also happens to be where it is useful - on a device composed of multiple components.


danielballan:
Yeah. This might turn out to be useful for V4 but has not yet come up in our use of `Device`.

Issue #626, opened 2018-10-24
Motor Subscribes in Inititalization


teddyrendahl:
## Description
`ophyd.EpicsMotor` subscribes to both the readback and `DMOV` flag on initialization.
https://github.com/NSLS-II/ophyd/blob/011c3e4a7d60a0eb98ae8cc669208d9984fca875/ophyd/epics_motor.py#L81

## Why does this matter?
1. In general we try to avoid connecting to any signals on `init`. This means any network hiccup or IOC down prevents the device from being loaded. Also prevents offline loading as discussed (#578)
2. The callback for the `DMOV` flag has some high level logging messages. This means that even if you sit with a blank screen and someone is moving one of those motors anywhere else and error occurs it shows up in your screen. Maybe this is desirable, but it is confusing when you are running a scan and an error message from another device shows up in the middle.

https://github.com/NSLS-II/ophyd/blob/011c3e4a7d60a0eb98ae8cc669208d9984fca875/ophyd/epics_motor.py#L258

danielballan:
I suspect this is because we haven’t revisited this since we added “stage” to the ophyd lexicon. 👍 for moving from init to stage.

klauer:
I think I'd still prefer to keep the state of `EpicsMotor` up-to-date regardless of whether it's being used in a scan.

If we had connection/disconnection hooks as in #607, subscription setup could be moved there. (Similarly, areaDetector plugin checking could be moved to `on_connection` and make everyone happier.)

`stage`, indicating the motor is used in a scan, could fix up (2). When used in a scan, those log messages probably should be very prominent.

danielballan:
That seems reasonable to me.

klauer:
Closed by #630

Issue #607, opened 2018-09-07
Add connection and disconnection callbacks to control layer.


danielballan:
On a group call, we have agreed that control layers should be required to provide a connection callback and a disconnection callback. "Can I use this yet?" is a universal concept. Simulated signals would of course just call the connection callback immediately and never call the disconnection one.

Issue #578, opened 2018-06-23
Allow disconnected devices to be instantiated


mrakitin:
We met a request from a BL staff to allow disconnected devices to be instantiated without raising. The idea is, for example, to avoid editing IPython collection profiles to exclude a detector from a pool of available detectors. After discussion with DAMA in NSLS-II Slack, we came up with an idea that such disconnected devices still can be silently instantiated, but when they are attempted to be staged/triggered, the attempt will fail with a raised error (this tripwire will help to avoid situations of collecting overnight data without a camera).

It is open for discussion.

tacaswell:
Not allowing this was a conscious design choice.  We figured it was better to know if you device was off line (or you typed the PV wrong) as soon as possible.  It is probably worth re-addressing this choice in light of the above discussion and the impending GUI future.

The minimal change would be to make Component lazy by default, however that seems to produce race conditions when trying to use some complex devices (not clear why or where, but things go sideways badly).

mrakitin:
Thanks for your feedback, @tacaswell! I thought it may be handy to have something like that in beamline profiles to work-around the situation with frequently changed equipment:
```py
try:
    det = DetectorClass(...)
except:
    det = None  # or NotImplemented
```

This way the profile will load fine, but it won't be possible to trigger the `det` detector.

danielballan:
But then _every_ place that you use `det` you have to `if det is None` checks. Very much against that.

ZLLentz:
We do something like this at SLAC with the try/except cascade. It creates the problems that Dan is talking about but our scientists much prefer the session get started with "something" than raising an exception and quitting. We wrap it with a context manager so that we only need one line and so that there is a clear ERROR log message with the failed device (bold red in our terminal):

https://github.com/pcdshub/hutch-python/blob/5e3aa362af93bde853eabe6d3417b1058dec4946/hutch_python/utils.py#L24-L53

mrakitin:
@danielballan, fair point.

tacaswell:
You could also do

```python
try:
    det = DectorKlass(...)
    sd.baseline.append(det)
except ConnectionError:
    pass
```

so it just is not there if it can not connect.

A down side of these approaches is if the device becomes available again later you have to restart the profile to get it.

I suspect the best path forward is sorting out how to make everything lazy by default and then explicitly checking connection on critical devices (ex refuse to start if we can not read the ring current?).

klauer:
[...]

Issue #433, opened 2017-08-22
Make Component lazy by default


danielballan:
This will improve startup times and make it safer to move config from _ad hoc_ IPython profiles into more formally managed modules.

Anecdotally, when I flipped `lazy=True` to `lazy=False` at CSX, a lot broke. We need to carefully go through `__init__` methods and make sure we aren't making unsafe assumptions there.

PR #630, opened 2018-10-26
WIP/ENH: add connection status, access rights subscriptions to all Signals

klauer:
Changes
---------
* `ophyd.Signal` gains `SUB_META`, which for now includes:
    1. Connection status information
    2. Read/write access information
* `ophyd.Component` supports subscriptions through simple decorators: `@component.sub_value` `@component.sub_meta` and the general `@component.subscriptions`
* AreaDetectors and EpicsMotor will no longer require a connection on instantiation - thereby speeding up initialization and allowing inspection of these objects when not fully connected
* `ophyd.Device` gains better support for iterating through the full tree of components and signals:
    1. `walk_components` (classmethod - give me all `Components` and `DynamicDeviceComponents`)
    2. `walk_signals` (instancemethod - give me all `Signal`s)
    3. `walk_subdevice_classes` (classmethod - give me all sub-[sub-]`Device` classes)
    4. `walk_subdevices` (instancemethod - give me all sub-[sub-]`Device` instances)
* `ophyd.Device` and `ophyd.Signal` get the `destroy` method - removing subscriptions and disconnecting signals
* The control layer now has `release_pvs`, and `name`

Test suite
----------
* Better handling of `Device` and `Signal` clean-up through a fixture
* Support for the `epics-on-travis` ADSimDetector `13SIM1:`
* caproto testing is now entirely `xfail`, but we plan to fix support prior to the next ophyd release

Related issues
---------------
Addresses #626 , #607, 
Partially addresses #578, #433


klauer:
cc @tacaswell @danielballan @ZLLentz @teddyrendahl

ophyd/_pyepics_shim.py
```diff
@@ -1,31 +1,74 @@
...
+from types import SimpleNamespace
+
+import epics
+from epics import get_pv as _get_pv, caget, caput, ca, dbr  # noqa
 
 try:
-    epics.ca.find_libca()
-except epics.ca.ChannelAccessException:
+    ca.find_libca()
+except ca.ChannelAccessException:
```
tacaswell: This logic is to handle tho no-pyepics case so ophyd will import.  You can probably safety rip all of this out (and replace it by raising `ImportError` so the the fallback works).

klauer: Thanks! Resolved in 55f1b8d.

ophyd/areadetector/plugins.py
```diff
@@ -232,6 +244,29 @@ def _asyn_pipeline_configuration_names(self):
...
+
+        plugin_type = self.plugin_type.get()
+        self._misconfigured = not plugin_type.startswith(self._plugin_type)
+        if self._misconfigured:
+            logger.warning(
+                'Plugin prefix %r: trying to use %r class (plugin type=%r) '
+                ' but the plugin reports it is of type %r',
+                self.prefix, self.__class__.__name__, self._plugin_type,
+                plugin_type
+            )
```
ZLLentz: Any worry about the change in behavior here? The previous setup would raise an exception on init, while this logs a warning instead. I think this is better but are we concerned about changing it?

ophyd/device.py
```diff
@@ -184,12 +189,48 @@ def __get__(self, instance, owner):
...
-        raise RuntimeError('Use .put()')
+        raise RuntimeError('Do not use setattr with components; use '
+                           'cpt.put(value)')
+
+    def subscriptions(self, event_type):
+        '''(Decorator) Specify subscriptions callbacks in the Device definition
+        '''
+        if isinstance(self, DynamicDeviceComponent):
+            raise NotImplementedError('DynamicDeviceComponent does not yet '
+                                      'support decorator subscriptions')
```
ZLLentz: perhaps this should be moved to the DynamicDeviceComponent class rather than checking here

ophyd/areadetector/plugins.py
klauer: Good question - my opinion is that the old behavior was annoying enough to warrant this change. 

This will `warn` and won't allow you to `stage` the detector, which is probably the most important portion. What does everyone else think?

ophyd/device.py
klauer: 👍 Good call

ophyd/epics_motor.py
```diff
@@ -194,7 +194,7 @@ def home(self, direction, wait=True, **kwargs):
         direction = HomeEnum(direction)
 
         self._started_moving = False
-        position = (self.low_limit+self.high_limit)/2
+        position = (self.low_limit + self.high_limit) / 2
```
ZLLentz: I agree with this change but the fact this is the default behavior is odd

klauer: Open an issue, please

ZLLentz:
I'm a big fan of this PR, especially the decorator syntax for creating these subscriptions.

ophyd/device.py
```diff
@@ -184,12 +189,48 @@ def __get__(self, instance, owner):
...
         if self.attr not in instance._signals:
-            instance._signals[self.attr] = self.create_component(instance)
+            cpt_inst = self.create_component(instance)
+            instance._signals[self.attr] = cpt_inst
+
+            for event_type, functions in self._subscriptions.items():
+                for func in functions:
+                    cpt_inst.subscribe(types.MethodType(func, instance),
+                                       event_type=event_type,
+                                       run=cpt_inst.connected)
```
teddyrendahl: So we will run the callback if the `cp_inst` is `lazy`?

ophyd/epics_motor.py
```diff
@@ -215,10 +215,12 @@ def check_value(self, pos):
         '''Check that the position is within the soft limits'''
         self.user_setpoint.check_value(pos)
 
+    @user_readback.sub_value
```
teddyrendahl: 👍

ophyd/device.py
```diff
@@ -184,12 +189,48 @@ def __get__(self, instance, owner):
...
+
+    def sub_connect(self, func):
+        'Connection subscription decorator'
+        return self.subscriptions('connect')(func)
+
+    def sub_value(self, func):
+        'Value subscription decorator'
+        return self.subscriptions('value')(func)
+
+    def sub_access(self, func):
```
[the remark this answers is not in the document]
klauer: I'm of the opinion that all devices and controls layers that should have concepts of connectivity, values, and read/write-ability. This seems to be the bare minimum of representing a device state, regardless of if EPICS is talking to it or not.

Onto your Tango note - if we have to rewrite `Device` (effectively all that ophyd is at this point) for every other control layer, that would indicate to me we are making a huge step in the wrong direction. If that's truly the case, then before implementing anything further we should all together do a serious survey of other possible control systems that could be integrated.

klauer:
I quite like @teddyrendahl 's idea and plan to rework that. I also have some thoughts on redoing the pyepics shim at a different level, such that it can be leveraged for pyepics/caproto, but that will take additional work.

We (NSLS-II, SLAC) should come to a decision regarding what @danielballan is noting above before proceeding, in any case.

teddyrendahl:
Agreed, what @danielballan brought up is a long conversation. Also if we are taking our one example of using a different controls system (Labview) we ended up just making a `caproto` IOC anyways 😄 

I was thinking about my `SUB_MD` in more general terms. Like what if `caproto` gives us a larger set of metadata /fields we can add callbacks to? This would still be `EPICS` underneath but just a richer interface because of a better client. It would be nice if this information got both into `describe` and my proposed `SUB_MD` without needs to make additional subscriptions based on control layer.

ophyd/areadetector/plugins.py
```diff
@@ -101,6 +100,19 @@ def __init__(self, *args, **kwargs):
     def stage(self):
         super().stage()
 
+        if self._misconfigured is None:
```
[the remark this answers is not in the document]
klauer: I don't think we do. I'd say 99%+ chance that this detects an ophyd misconfiguration rather than an IOC misconfiguration, meaning that you should be adjusting your startup/detector class and then restarting the session. 

Any strong feelings on this @tacaswell @danielballan ?

ophyd/device.py
klauer: How this works (admittedly somewhat confusingly) is as follows:
Callback data is cached on the ophydobject level. If you subscribe for the first time, you have the option of using that cached information to run the callback immediately. If you don't do that, then the next time you'll see an update is when the subscription fires again for all other subscribers.

This specific line works as follows: if it's already connected, run the subscription, otherwise wait until connected as we're likely to get the first valid subscription callback from that.

danielballan:
:+1: for `SUB_MD` and for punting the larger question a raised to a discussion on a call

ophyd/_caproto_shim.py
```diff
@@ -1,11 +1,120 @@
...
+        pyepics_compat.get_pv = pyepics_compat._get_pv
+
+        logger.debug('Performing ophyd cleanup')
+        if _dispatcher.is_alive():
+            logger.debug('Joining the dispatcher thread')
+            _dispatcher.stop()
+
+        _dispatcher = None
+
+    logger.setLevel('DEBUG')
```
danielballan: I don't think we should set logger levels. Feel free to punt this to some future comprehensive review of loggers in ophyd, if you want. Now that I finally understood them for caproto, I want to clean up ophyd and bluesky logging....

ophyd/areadetector/plugins.py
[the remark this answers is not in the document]
danielballan: Usually this requires adjusting the plugins, not rebooting the IOC or restarting the IPython session, but I don't like this idea of the scan sitting here and periodically polling whether the plugins have been fixed.I think it makes sense to keep this as an error.

ophyd/ophydobj.py
```diff
@@ -192,7 +192,7 @@ def _run_subs(self, *args, sub_type, **kwargs):
         for cb in list(self._callbacks[sub_type].values()):
             cb(*args, **kwargs)
 
-    def subscribe(self, cb, event_type=None, run=True):
+    def subscribe(self, callback, event_type=None, run=True):
```
danielballan: :heart:

ophyd/signal.py
```diff
@@ -70,6 +71,14 @@ def __init__(self, *, name, value=0., timestamp=None, parent=None,
         # self.tolerance is a property
         self.rtolerance = rtolerance
 
+        # Signal defaults to being connected, with full read/write access.
```
danielballan: Hm. I could argue this in both directions. What's your reasoning?

ophyd/_caproto_shim.py
klauer: 💯% agreed and good catch - this was for my debugging purposes and needs to be removed.

ophyd/signal.py
klauer: As-is, `Signal` can be used as a soft signal. Because of that, you generally want the ability to read, write, and say that it's "connected" straightaway.

We are cheating in a way here by not having a `SignalBase` that defaults to all of these being `False` and raising `NotImplementedError` for all default implementations. I'm 50/50 on whether this is actually a deficiency or a fair removal of one probably-unnecessary level of the class hierarchy...

danielballan: Makes sense. I forget who said, "In Python, inheritance is just for code reuse," but in that spirit I'm happy with the current situation. We don't have much code to reuse.

ophyd/signal.py
```diff
@@ -1096,6 +1118,14 @@ def _write_pv_access(self, read_access, write_access, pv):
                        timestamp=self._metadata.get('timestamp'),
                        **self._metadata)
 
+    def destroy(self):
+        '''Destroy the EpicsSignal from the underlying PV instance'''
+        super().destroy()
+        if self._write_pv is not None:
+            self._write_pv.clear_auto_monitor()
```
danielballan: Maybe the control layer should provide a `release_pvs` method, symmetric to `get_pvs`. For the pyepics CL, it would call this. For the caproto CL, it would do the same + `pv._caproto_pv.go_idle()` which will disconnect the channel.

ophyd/_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+                    self.logger.exception(
+                        'Exception occurred during callback %r', callback
+                    )
+
+        self.logger.debug('Callback thread %s exiting', self.name)
+        self.detach_context()
+
+    def attach_context(self, context):
+        self.logger.debug('Callback thread %s attaching to context %s',
+                          self.name, self.context)
```
danielballan: Is this missing

```py
self.context = context
```

danielballan: It surprised me that `__init__` takes a `dispatcher` and picks up a lot of its attributes, but `attach_context` takes a `context` directly but hangs onto the old dispatcher's other attributes.

ophyd/device.py
```diff
@@ -780,9 +852,125 @@ def __init__(self, prefix='', *, name, kind=None,
...
+            Where ancestors is all ancestors of the signal, including the
+            top-level device `walk_components` was called on.
+        '''
+        for attr, cpt in cls._sig_attrs.items():
+            yield ComponentWalk(ancestors=(cls, ),
+                                dotted_name=attr,
+                                item=cpt
+                                )
+            if isinstance(cpt, DynamicDeviceComponent):
+                ...
```
danielballan: Is this not implemented or something clever?

ophyd/signal.py
```diff
@@ -466,22 +597,38 @@ def subscribe(self, callback, event_type=None, run=True):
...
+            if not pv.wait_for_connection(timeout=timeout):
+                raise TimeoutError('Failed to connect to %s' % pv.pvname)
+
+        # Ensure callbacks are run prior to returning, as
+        # @raise_if_disconnected can cause issues otherwise.
+        try:
+            self._pvs_ready_event.wait(timeout)
+        except TimeoutError:
+            raise TimeoutError('Control layer {} failed to send connection and '
+                               'access rights information within a reasonable '
```
danielballan: Can this include the `timeout` instead of "a reasonable time period"?

ophyd/_dispatch.py
klauer: I could make up excuses here, but it was just sloppy. All things picked up from `dispatcher` previously are now kwargs.

danielballan:
Implementation details. I can't see anything about the API that feels off, but we might shuffle code around a little bit internally once we understand the latent issues better.

klauer:
... it... it... passed!

PR #632, opened 2018-11-05
Kind Tests

teddyrendahl:
A more explicit explanation of a few gripes that I have with how `Kind` currently works. I think the `Signal` test was an explicit feature that I disapprove of, the other I think may be unintentional.

The `read_attrs` one is probably not a huge deal if you plan on deprecating it. It is however slightly annoying if you want to introspect without calling `read` itself.

### DIFF
--- a/ophyd/tests/test_hints.py
+++ b/ophyd/tests/test_hints.py
@@ -13,7 +13,8 @@ def test_device_hints():
 
     assert {'fields': ['dev_a']} == Dongle(name='dev').hints
     assert {'fields': ['dev_a']} == Dongle(name='dev').hints
-
+    # A signal should record its own hints
+    assert len(Dongle(name='dev').b.hints.get('fields')) > 1
 
 def test_motor_bundle_hints():
     class Bundle(MotorBundle):
--- a/ophyd/tests/test_hints.py
+++ b/ophyd/tests/test_hints.py
@@ -20,8 +20,10 @@ def test_motor_bundle_hints():
     class Bundle(MotorBundle):
         a = Component(SynAxis, kind=Kind.hinted)
         b = Component(SynAxis, kind=Kind.hinted)
+        c = Component(SynAxis, kind=Kind.normal)
 
     assert {'fields': ['dev_a', 'dev_b']} == Bundle(name='dev').hints
+    assert {'fields': ['dev_c']} == Bundle(name='dev').c.hints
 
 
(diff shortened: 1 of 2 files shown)


danielballan:
> I think the Signal test was an explicit feature that I disapprove of

Yes, but I think we were wrong and this should be changed in the way you describe.

I think this is how we got here: we thought of `obj.hints`, `obj.read_attrs`, `obj.configuration_attrs` as meaning, "Tell me about your kids!" And of course `Signal` has none. But we _should_ think about `obj.hints` (and friends) as meaning, "Tell me how to categories your data keys." All of a Device's data keys propagate up from its children, but a Signal owns its own data keys and thus should refer to them in `Signal.hints`.

Same page?

> The `read_attrs` one is probably not a huge deal if you plan on deprecating it

I believe we started down the `Kind` path with the idea of deprecating `read_attrs`, but it became clear that `read_attrs` is a really useful way of setting Kind in batch. Now _storing_ the state locally is unequivocally better, but we other _access_ modes have their place.

I guess the current behavior is a straight-ahead bug then, right? Am I missing some downside to changing the behavior so as to fix it?

teddyrendahl:
> Same page?

👍 

> I guess the current behavior is a straight-ahead bug then, right? Am I missing some downside to changing the behavior so as to fix it?

I'd lean towards bug. No harm in fixing. Obviously not mission critical as`read` gives you the correct data. Where I run into this in `typhon` where I'm using `ophyd.Device` not in a scan, but more as a way to organize the importance of various signals and sub-devices as you mentioned in the `hints` discussion. This doesn't work so well as calling `read` will give me a dictionary but I'm not guaranteed to be able to retrace the steps from event key to attribute name.

I also think it would be of interest to address the test case in c187964. Even if I have a `normal` Component, the hints of that Component end up in the top level dictionary. (It is the existing test that is failing). I added the third line because I'm hoping the fix also preserves the `hints` of the Component for the same reasoning as the above `Signal` discussion.

danielballan:
> I also think it would be of interest to address the test case in c187964. Even if I have a normal Component, the hints of that Component end up in the top level dictionary. (It is the existing test that is failing). I added the third line because I'm hoping the fix also preserves the hints of the Component for the same reasoning as the above Signal discussion.

I _think_ I follow and agree. I think we are on the same page about desired behavior and should move on to implementing it.

PR #633, opened 2018-11-06
FIX: missing imports

klauer:
`ROIPlugin.set` (not shown in the diff) had some missing imports.

Also made a pass to get flake8 to stop complaining, hence the unrelated whitespace changes.

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -7,13 +7,16 @@
 '''
 
 
+import functools
+import logging
+import numpy as np
+import operator
 import re
 import time as ttime
-import logging
-from collections import OrderedDict
-import numpy as np
 
-from ophyd import Component as Cpt
+from collections import OrderedDict
+
+from .. import Component as Cpt
 from .base import (ADBase, ADComponent as C, ad_group,
                    EpicsSignalWithRBV as SignalWithRBV)
 from ..signal import (EpicsSignalRO, EpicsSignal, ArrayAttributeSignal)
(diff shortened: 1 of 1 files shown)


danielballan:
> Also made a pass to get flake8 to stop complaining, hence the unrelated whitespace changes.

:+1: We agreed a couple months ago to "bomb" the codebases with a flake8-fixing PR at some point since they are (1) not that old and (2) pretty close to compliance anyway, so the diff won't touch _that_ much of substance. I actually have branches for this but they are so stale now I should just start over. Will wait until the next release when we have fewer PRs in motion.

teddyrendahl:
@klauer Can you also add the missing `WidthX` and `WidthY` signals to Overlay in this PR? 

https://github.com/NSLS-II/ophyd/commit/0e1e809366690ed831d5d8aa980bd2ffe6be9d1e

teddyrendahl:
@klauer has raised a good point that `widthX` is not available in `AreaDetector < 1.9`. We do not have this deployed at LCLS anymore, but as I recall NSLS-II might? Is this true?

klauer:
As it appears that ophyd has to support 1-9-1 (or at least that NSLS-II is unable to confirm the versions they are running, which likely still includes R1-9-1), we can't add `Width[XY]` just yet. 

This will sadly have to be done locally at SLAC for now.

mrakitin:
@klauer, indeed, we still use R1-9-1 at a few beamlines according to our internal records for the software on the floor.

tacaswell:
We have started to deal with AD versions via code like https://github.com/NSLS-II/nslsii/blob/master/nslsii/ad33.py

There was a, ah, [vigorous discussion](https://github.com/NSLS-II/ophyd/pull/602) around this.  Given that SLAC needs to support a different set of AD, this maybe a good time to return to that discussion.

PR #636, opened 2018-11-07
FIX/TST: fix some test issues in areadetector/sim

danielballan:
Cherry-picked from 6d287ff276e38a21b4bfce62b506f7ea389b11bb to fix the tests


danielballan:
Merging as a hot-fix to get CI passing again, per discussion in Slack.

PR #641, opened 2018-11-10
TST: move fully to pytest

klauer:
Remove asserttools (= unittest shim) in the test suite, moving it fully to pytest

There should be no functional differences here.


klauer:
Refactored tests passing (cc @tacaswell)

danielballan:
:+1: for pytest's nice, introspect-y error messages!

PR #642, opened 2018-11-12
REF: remove Python 3.5 support

klauer:
Background
------------
Python 3.6 offers features that make it much easier to implement all features of `ophyd.Device` without using metaclasses.

ComponentMeta, the metaclass implementation detail used by `Device` under the hood, has been removed in favor of `__init_subclass__` and `__set_name__`, offered by Python 3.6. Each new `Device` subclass will automatically have `__init_subclass__`  called on it, at which point we do everything previously done in `ComponentMeta`.

Changes
---------
* Removes backported `Enum`
* Removes `ComponentMeta`
* Moves RESERVED_ATTRS/INSTANCE_ATTRS to an accessible place
* Mucks with CI as caproto is making development and testing on travis difficult


klauer:
Modulo CI failing due to caproto, I think this is ready.

Is everyone ready for a 3.6+-only world?

danielballan:
I recall @hhslepicka voicing support for extended 3.5 support last time we brought this up. Would also like sign off from @prjemian.

The long-term strategy, I believe, it to support the "last two" stable releases from Python. This is what most projects did in the Python 2.x days.

teddyrendahl:
I'm fine with this, PCDS is only Python 3.6.

tacaswell:
> The long-term strategy, I believe, it to support the "last two" stable releases from Python. This is what most projects did in the Python 2.x days.

Matplotlib's policy is "minor versions of python initially release in the last 36 months of our projected release date" (which more-or-less gets you to "last 2", but with less weird edge cases).

ophyd/device.py
```diff
@@ -857,6 +795,62 @@ def __init__(self, prefix='', *, name, kind=None, read_attrs=None,
             [getattr(self, attr) for attr, cpt in self._sig_attrs.items()
              if not cpt.lazy or cpt._subscriptions]
 
+    @classmethod
+    def _initialize_device(cls, **kwargs):
```
tacaswell: If we do not use `kwargs` we should not take them.

tacaswell:
I agree @prjemian should sign off on this.

prjemian:
It's OK

klauer:
Tests are passing. I think we have agreement from all required. 

Anyone want to click this nice button?
image

portrain:
No problems from my (and Australian Synchrotron) side to remove Python 3.5 support

klauer:
Thanks @portrain - I'll know from now on to reach out to you for future PRs similar to this.

PR #645, opened 2018-11-13
CI: epics-on-travis, take 3

klauer:
It's not as clean as it could be, but I think it's still better than the old docker solution.

TODO:
- Fake motor tests repeatably failing - I'd like to replace this with a simple caproto IOC that will be much easier to debug

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,105 +1,87 @@
 language: python
-
 sudo: required
+python:
+  - 3.6
 dist: trusty
+
 env:
   global:
     # Doctr deploy key for NSLS-II/NSLS-II.github.io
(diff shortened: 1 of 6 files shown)


klauer:
I plan to sit on this PR until we can remove 3.5, as I'd like to assume we can use caproto (which of course requires 3.6+).

.travis.yml
```diff
@@ -1,105 +1,86 @@
...
-       ./miniconda.sh -b -p /home/travis/mc
-       export PATH=/home/travis/mc/bin:$PATH
-       conda config --set show_channel_urls True
-       conda config --set always_yes True
-    fi
+  # eventually, we should use the specific version tag of epics-on-travis:
+  # - git clone --single-branch --branch "${EPICS_ON_TRAVIS_VERSION}" https://github.com/klauer/epics-on-travis "$CI_TOP"
+  # but for now, get the latest scripts:
+  - git clone --single-branch --branch master https://github.com/klauer/epics-on-travis "$CI_TOP"
+  - sudo ifup docker0
```
tacaswell: why bring up the docker network?

klauer: The docker network addresses seem to be consistent between runs, and I was having trouble with the broadcast traffic of `eth0` for whatever reason.

PR #646, opened 2018-11-15
TST/CLN: Remove all traces of FakeEpicsPV

klauer:
~(Builds on #645 - that should be merged first)~

Pytest fixture-based IOCs replace old simulated `pyepics.PV` instances that were a source of a bunch of testing instability/frustration.

Not to be confused with `ophyd.sim.FakeEpicsSignal` and `ophyd.sim.make_fake_device` which will remain (and are very useful).

Changes:
* test_signal - simple signals replaced with caproto IOC
* test_mca - dxp/mca replaced with caproto IOC
* test_scaler - scaler replaced with caproto IOC
* test_quadem - replaced with `ophyd.sim.make_fake_device(QuadEM)`
* test_motor - (already in epics-on-travis PR) replaces fake motor IOC with simple caproto IOC


klauer:
OK, this is ready for review, too. Directory structure fix-up for areadetector file saving + inclusion of databroker/h5py in travis now allow for more tests to run than with the previous docker solution.

klauer:
Thanks, @ZLLentz! Any opinion, @tacaswell?

klauer:
Force-pushed to rebase on master and get travis/coverage going.

.travis.yml
```diff
@@ -66,13 +66,15 @@ before_install:
   - bash "${CI_SCRIPTS}/run-epics-iocs-on-procserv.sh"
 
 install:
+  - |
+    if [[ $TRAVIS_PYTHON_VERSION == 'nightly' ]]; then
+      # Nightly has some issues with h5py
+      sed -i -e '/h5py/d' test-requirements.txt
+    fi
```
ZLLentz: This doesn't fill me with fuzzy feelings

klauer: @tacaswell will installing `h5py-dev` help here or is there something else?

ophyd/signal.py
```diff
@@ -18,7 +18,7 @@
 
 
 class Signal(OphydObject):
-    '''A signal, which can have a read-write or read-only value.
+    r'''A signal, which can have a read-write or read-only value.
```
ZLLentz: Is there a reason to make docstrings raw strings? Or is this just for the contents of this particular docstring?

ophyd/tests/mca_ioc.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+from caproto.server import (pvproperty, PVGroup, ioc_arg_parser, run, SubGroup,
+                            get_pv_pair_wrapper)
+from caproto import ChannelType
+
+pvproperty_with_rbv = get_pv_pair_wrapper(setpoint_suffix='',
+                                          readback_suffix='_RBV')
+unknown = int
```
ZLLentz: This looks odd. Is it just a quick way to set up defaults?

ophyd/tests/test_areadetector.py
```diff
@@ -417,17 +439,26 @@ class MyDetector(SingleTrigger, SimDetector):
             assert Path(fn).exists()
 
 
-@pytest.mark.skipif(not os.path.exists('/data'), reason='No /data')
+@pytest.fixture
+def h5py():
+    try:
+        import h5py
+    except ImportError as ex:
+        raise pytest.skip('h5py unavailable') from ex
```
ZLLentz: This is a neat trick, didn't realize you could do this. Maybe I should read the pytest docs more closely.

ophyd/signal.py
klauer: Looks like there are some invalid escape sequences in this specific docstring, according to @tacaswell (cherry picked his commit)

ophyd/tests/test_quadem.py
```diff
@@ -1,67 +1,36 @@
...
-        current2 = Cpt(FakeStats, 'Current2:')
-        current3 = Cpt(FakeStats, 'Current3:')
-        current4 = Cpt(FakeStats, 'Current4:')
-        sum_all = Cpt(FakeStats, 'SumAll:')
-
+    FakeQuadEM = make_fake_device(QuadEM)
     em = FakeQuadEM('quadem:', name='quadem')
+    clear_fake_device(em)
 
-    ''' Beware: Ugly Hack below
```
ZLLentz: It makes me happy to see lines removed on a comment that says "beware: ugly hack"

ophyd/tests/mca_ioc.py
klauer: Good catch! Yes, it's what `caproto.server.conversion.ophyd_device_to_caproto_ioc` gives for `dtype` when you only have a `Device` class and not an instance which can talk to a real IOC.

The changes from `unknown` to some dtype were the minimum changes necessary to mock the IOC for ophyd's usage. If these PVs become important later, we will have to revisit the caproto IOC.

ophyd/signal.py
ZLLentz: :+1:

PR #647, opened 2018-11-16
ENH: add check in AD.HDF5Plugin that plugins are 'primed'

klauer:
(Separated from #618  - thanks @tacaswell )

Raise in stage the plugin things it has a 0 size array (which typically means it has never had and array pushed through it).


klauer:
Things to look into prior to merging:
* What happens with disabled plugins?
* Have we resolved how defining all plugins on a detector works and only staging the ones we care about? (If not, could this information be set via `kind`?)

klauer:
I think this one is ready. The comment I made above is not directly related - I'm not sure why.

PR #655, opened 2018-11-27
MNT: Ensure gaussian parameters are available after class construction

teddyrendahl:
We are creating a simulation with `SynGauss` as a base. However, it is quite annoying needing to construct a new instance each time we want to modify the shape of our gaussian. This makes the parameters available post class construction so that they can be modified / inspected later. 

I went ahead and also made sure `motor` and `motor_field` were available as well largely for completeness.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -404,20 +404,20 @@ class SynGauss(SynSignal):
             raise ValueError("noise must be one of 'poisson', 'uniform', None")
         self._motor = motor
         self._motor_field = motor_field
-        self._center = center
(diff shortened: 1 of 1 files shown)


prjemian:
+1 -- wanted this for a while

teddyrendahl:
Only question ... Should these be public variables instead?

klauer:
It's solely for simulation - feel free to make them public @teddyrendahl.

teddyrendahl:
Made the gaussian parameters public, left the motor ones private as most of the times these should be constant usually and `_motor` already exists.

PR #656, opened 2018-12-03
ENH: added Dexela detector

mrakitin:
During a visit at NSLS-II SRX beamline, we with @ke-zhang-rd figured out we miss the [Dexela detector](http://cars9.uchicago.edu/software/epics/DexelaDoc.html) (from [this list](http://cars9.uchicago.edu/software/epics/areaDetector.html)). This PR intends to fix this omission.


klauer:
Missing the cam? Original was https://github.com/NSLS-II-HXN/hxntools/blob/master/hxntools/detectors/dexela.py#L63 - is SRX's modified compared to HXN's?

ophyd/areadetector/detectors.py
```diff
@@ -109,6 +110,11 @@ class BrukerDetector(DetectorBase):
     cam = C(cam.Andor3DetectorCam, 'cam1:')
 
 
+class DexelaDetector(DetectorBase):
+    _html_docs = ['DexelaDoc.html']
+    cam = C(cam.DexelaDetectorCam, 'cam1:')
```
awalter-bnl: Do we also need to add `DexelaDetectorCam` into `cam` in order to get this to work? I just want to make sure that if it is added here it will work as expected.

danielballan:
Thanks, @mrakitin. I believe this is the first model we have added since @klauer's initial work years ago. Once it's ready, let's all await final approval from @klauer required before merging.

klauer:
Sounds good. 

Somewhat related: I also have a 66.667% ready-to-submit PR that includes support for all AreaDetector plugin versions from 1-9-1 to 3-3. I haven't done anything with the cams just yet, though. That work still shouldn't conflict with this specific PR, but just so you're aware it's coming down the line.

mrakitin:
Sorry for the delay, this PR was lost after I returned back from vacation in December. Going to revive it this week. @klauer, do you think your PR mentioned above will affect this one somehow?

mrakitin:
Ready for your review; the cam was copied from https://github.com/NSLS-II-HXN/hxntools/blob/baa8a9819b965db6ca2fe19b40d3981823d9a41c/hxntools/detectors/dexela.py#L34-L60.

mrakitin:
Thank you, @klauer!

PR #659, opened 2018-12-06
WIP - ENH: Add support for pvAccess normative types

klauer:
**Work in progress - do not merge**

Builds on #653, where the vast majority of these commits come from. Not to be merged prior to that PR.

Goal
-----
* Adds preliminary support for pvAccess normative types. 
* Test with AreaDetector PVA plugin

Limitations
------------
* This only supports normative types for now

Though that may seem like a major restriction, all of the examples of pvAccess in the wild I can find are solely based on normative types. Arbitrary PV groupings as available in pva2pva are not yet in wide usage.

Bugs
-----
There are a significant number of not useful debug messages dumped out to standard output/error from somewhere in pvAccess. It's not something we have control over.

TODO
------
* Work on merging the PR this is based on
* Further testing, with pva2pva and p4p mailbox examples
* Eventually write a `PvaSignal` for non-normative types; allow structure to be mapped onto Device components

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 networkx>=2.0
 numpy
 pyepics>=3.3.2
+p4p
--- a/ophyd/__init__.py
+++ b/ophyd/__init__.py
@@ -33,6 +33,8 @@ def set_cl(control_layer=None, *, pv_telemetry=False):
         from . import _pyepics_shim as shim
     elif control_layer == 'caproto':
         from . import _caproto_shim as shim
+    elif control_layer == 'p4p':
+        from . import _p4p_shim as shim
     elif control_layer == 'dummy':
         from . import _dummy_shim as shim
     else:
--- a/ophyd/_p4p_shim.py
+++ b/ophyd/_p4p_shim.py
@@ -1,16 +1,6 @@
 '''
 P4P TODO list
 
-* Discuss with upstream
-    1. There does not appear to be a way to monitor connection status outside
-       of a value monitor
-    2. No such thing as read/write access?
-    3. Wrapping basic Python data types...
-    4. pvRequest filters for monitoring data? pvRequest seems limited to a
-       comma-delimited list of dotted fields
-    5. Issue: '[13SIM1:Pva1:Image] warning : not a channel request'
-       What does this mean/why is it not an exception we can catch?
-
 * In this shim
     1. Support non-normative types (and attributes/etc with AD images)
     2. Process by default?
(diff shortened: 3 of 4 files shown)


klauer:
I suppose if there's no interest/input from the community, we can just merge this after a couple minor fixes. Then we'll use it here at SLAC, and iterate on the solution.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 networkx>=2.0
 numpy
 pyepics>=3.3.2
+p4p
```
ZLLentz: Does this belong here so soon? We don't even include caproto in this list yet.

ZLLentz:
This seems sufficient for near-term use at SLAC. I don't know enough about p4p to comment on the implementation, and it's hard to tell which test failed when the caproto test prints overflow the 10,000 line travis build log limit...

requirements.txt
klauer: Good catch - a mistake on my part. It should be a dev-requirement.

tacaswell:
`p4p` definitely should be optional!

In light of https://github.com/NSLS-II/ophyd/issues/678 is it worth the effort to shoe-horn the v4 stuff into `EpicsSignal` (which is the most v3 coupled part of ophyd) or would it be better to jump straight to `PVASignal`?

I think `control_layer` was a good theory that did not survive contact with reality and should really be called `v3_control_layer`.

if SLAC needs this, then :+1: but I also do not know enough about p4p to review the details of the implementation.

klauer:
I think we can bring this into SLAC's repository and continue development there. ophyd doesn't need it, and @tacaswell will develop the apparently superior `PVASignal`.

OK to close this?

tacaswell:
APS is also interested in V4 so there is value to having this work be done in ophyd. Long term, I think `PVASignalDevice` (however #678 lands and whatever we call it) does belong on ophyd so this does to.

My question was given the amount of work that went into this to degrade a V4 PV to look like a V3 PV would it be simpler to write a new `Signal` subclass that owns a `p4p` PV?  This would insulate the V4 work from any changes we want to make to control_layer (which @teddyrendahl has been advocating, but I am not sure what those are) and I personally strongly suspect that there will someone will have to write the `PVASignal` eventually.  This was meant as a technical implementation question (and I could be discounting the amount of fiddly detail in `EpicsSignal` ) not a philosophical or need-based question.

As I said above I am :+1: on this going in, but think @prjemian should be the one to merge.

klauer:
You make good points, @tacaswell, and upon reflecting on all of the above: this intermediate step is simply not worth it. It gives us 50% of what will eventually be needed at the cost of twice as much to maintain.

As the author of it, let me make the difficult decision to launch it into orbit. On to `PvaSignal`...

PR #660, opened 2018-12-11
MNT: Add collect_asset_docs to Mockflyer

ke-zhang-rd:


ophyd/sim.py
```diff
@@ -611,6 +617,39 @@ def _scan(self):
...
+
+        # This is temporarily more complicated than it will be in the future.
+        # It needs to support old configurations that have a registry.
+        resource = {'spec': self._spec,
+                    'root': self.save_path,
+                    'resource_path': self._file_stem,
+                    'resource_kwargs': {},
+                    'path_semantics': os.name}
+        # If a Registry is set, we need to allow it to generate the uid for us.
+        if self.reg is not None:
```
tacaswell: I am not sure we want to support this code path here?  I like @danielballan 's suggestion from the bluesky issue about supporting this in a sub-class.

PR #662, opened 2018-12-12
ENH: support versioning of ophyd objects

klauer:
This is an important feature for the upcoming AreaDetector plugin PR, which I'll open after this one goes in. Let's hash over the details of this and get it to a good state before that.

This adds two keyword-only arguments to subclasses of ophyd objects, primarily intended for use with Devices:
* `version`: the version itself - no type restrictions
* `version_of`: the class in the hierarchy that `version` is in reference to

The information on the object then allows for programmatic access of different versions of the class.

The test case below shows the intent well enough, I think:

```python
    class MyDevice(Device, version=1):
        ...
     info = MyDevice._class_info_
    assert info['version'] == 1
    assert info['versions'] == {1: MyDevice}
     # Define a new version of that Device:
    class MyDevice_V2(MyDevice, version=2, version_of=MyDevice):
        ...
     info = MyDevice_V2._class_info_
    assert info['version'] == 2
    assert info['versions'] == {1: MyDevice, 2: MyDevice_V2}
     # Ensure that the original Device has also been updated:
    assert MyDevice._class_info_['versions'] == {1: MyDevice, 2: MyDevice_V2}
     # Define a user device that inherits - but does not define a new version
    class UserDevice(MyDevice_V2):
        ...
     assert UserDevice._class_info_ == {'versions': {1: MyDevice, 2: MyDevice_V2},
                                       'version': 2
                                       }
```

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -97,6 +97,50 @@ class OphydObject:
         # Instantiate logger
         self.log = logging.getLogger(base_log + '.' + name)
 
+    def __init_subclass__(cls, version=None, version_of=None, **kwargs):
+        'This is called automatically in Python for all subclasses of OphydObject'
+        super().__init_subclass__(**kwargs)
+
+        if version is None:
+            # Only in the simple case of single inheritance, copy in the parent
+            # version information
+            try:
+                parent_class, = cls.__bases__
+            except ValueError:
+                parent_info = None
+            else:
+                parent_info = getattr(parent_class, '_class_info_', None)
+
+            if parent_info is not None:
+                # This class does not *define* a version, but it may be of
+                # a specific version.
+                # Propagate the version information from the parent
+                cls._class_info_ = parent_info.copy()
+            else:
+                cls._class_info_ = dict(versions={})
+
+            return
+
+        if version_of is None:
+            # This class is the first of its version
+            ...
+            version_dict = {}
+        elif not issubclass(cls, version_of):
+            raise RuntimeError(
+                f'Versions are only valid for classes in the same '
+                f'hierarchy. {cls.__name__} is not a subclass of '
+                f'{version_of.__name__}.'
+            )
+        else:
+            version_dict = version_of._class_info_['versions']
+
+        version_dict[version] = cls
+
+        cls._class_info_ = dict(
+            versions=version_dict,
+            version=version,
+        )
+
     def _validate_kind(self, val):
         if isinstance(val, str):
             return Kind[val.lower()]
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -820,7 +820,7 @@ class Device(BlueskyInterface, OphydObject):
     def __init_subclass__(cls, **kwargs):
         'This is called automatically in Python for all subclasses of Device'
         super().__init_subclass__(**kwargs)
-        cls._initialize_device(**kwargs)
+        cls._initialize_device()
 
     @classmethod
     def walk_components(cls):
--- a/ophyd/tests/test_versioning.py
+++ b/ophyd/tests/test_versioning.py
@@ -1,0 +1,36 @@
+import logging
+from unittest.mock import Mock
+
+from ophyd import (Device, Component, Signal)
+
+
+logger = logging.getLogger(__name__)
+
+
+def test_specify_version():
+    # Define a versioned Device:
+    class MyDevice(Device, version=1):
+        ...
+
+    info = MyDevice._class_info_
+    assert info['version'] == 1
+    assert info['versions'] == {1: MyDevice}
+
+    # Define a new version of that Device:
+    class MyDevice_V2(MyDevice, version=2, version_of=MyDevice):
+        ...
+
+    info = MyDevice_V2._class_info_
+    assert info['version'] == 2
+    assert info['versions'] == {1: MyDevice, 2: MyDevice_V2}
+
+    # Ensure that the original Device has also been updated:
+    assert MyDevice._class_info_['versions'] == {1: MyDevice, 2: MyDevice_V2}
+
+    # Define a user device that inherits - but does not define a new version
+    class UserDevice(MyDevice_V2):
+        ...
+
+    assert UserDevice._class_info_ == {'versions': {1: MyDevice, 2: MyDevice_V2},
+                                       'version': 2
+                                       }


danielballan:
Slick! I have never seen custom kwargs passed in to class info, but this is a compelling place to use that language feature.  Will review in detail later this week but my initial feeling is 👍

tacaswell:
Overall :+1: 

To @teddyrendahl 's suggestions 

1. yes, that seems good
2. rename 'version' -> 'ioc_version' ?

klauer:
As far as naming goes, I don't have much in the way of strong feelings. A few related points:
* `ioc_version` works from the perspective of `Device`, but not necessarily other ophyd objects
* `version` is generic enough that it wouldn't have to apply to an IOC version if you didn't want it to - a double-edged sword, I suppose
* `_class_info_` itself is named poorly, what should it be called? 
* What name should the classmethods have that expose this information to the end user?

teddyrendahl:
Spitballing:

```python

class EpicsMotor(dependency='EpicsMotor', version='2.1.2',  version_of=EpicsMotorBase):

@classmethod
def dependencies(cls):
     return (cls._class_info['dependency], cls._class_info['version'])
```

tacaswell:
Versions should be tuples not strings (it will make sorting / comparison easier later).

`ioc_version` may be EPICS specific, was going for something that says "the version of thing I am sitting in front of" not "my version".  

`server_version`, `device_version`, `controller_version`, `other_side_version`?

ophyd/ophydobj.py
```diff
@@ -97,6 +97,50 @@ def __init__(self, *, name=None, attr_name='', parent=None, labels=None,
...
+            except ValueError:
+                parent_info = None
+            else:
+                parent_info = getattr(parent_class, '_class_info_', None)
+
+            if parent_info is not None:
+                # This class does not *define* a version, but it may be of
+                # a specific version.
+                # Propagate the version information from the parent
+                cls._class_info_ = parent_info.copy()
```
klauer: I'll have to remove this portion - but it will further simplify `__init_subclass__`. 

We can still figure out which version an object is by inspecting `cls.mro()._class_info_` and then checking `issubclass(new_obj, cls_version) for ver, cls_version in _class_info_['versions'].items()` (semi-pseudo-code here, of course)

klauer:
Agreed about versions being tuples, even if it is slightly more unwieldy. Not sure if there could (or should) be magic to automatically convert `'R2.1.2'` -> `(2, 1, 2)`.

When you get to this number of arguments:
```python
class MyMotor(EpicsMotor, version_reference='EpicsMotor', version=1, version_of=EpicsMotor)
```

It almost seems like there should be then another type that encompasses all of this.
```python
vinfo = VersionInfo(version=2, of=EpicsMotor, reference='EpicsMotor')
class MyMotor(EpicsMotor, version=vinfo)
```
Despite the suggestion, I'm not a fan of this at all...

klauer:
It's necessary to specify the base class that a version specification is for. That is what `version_of` indicates, as in the code in this PR.

The "reference" portion came up as @tacaswell and @teddyrendahl indicate that the word "version" itself is too generic, and should have some indicator of what that version means. Is it the IOC version, is it the motor record version, etc.

My personal opinion is that what the version refers to isn't so important, just that the user remains consistent in their versioning of the classes. It _should_ be apparent that if you are versioning a `Device` tied to a specific IOC, that the version uses the same scheme as the IOC you are basing it on. That being said, I'm not against adding more keywords/information to give meaning to that version. It might be something that's only specified in the base class, even.

teddyrendahl:
* Agreed, adding the extra key for `dependency`/`reference` is unwieldy. I think most cases you can infer from the class what IOC / module you are versioning against.

* I'm fine with `tuple` being passed directly into the class, I think less magic might be better in some of these cases. 

* What about `external_version` instead of `version`? Kind of verbose, but it is clear that it has nothing to do with the versioning of the `Python` module without loss of generalization for what you are versioning against.

teddyrendahl:
I like this suggestion, now just thinking about names:
```python
{'version': (2, 0, 0),  'version_base': MyIOC,  'version_of': 'ioc_version'}
```

tacaswell:
What if we want to track the version of more than one thing on the class?

klauer:
@tacaswell what would be an example of that?

tacaswell:
maybe something about how the Resource/datum are structured or something about how it expects to be triggered?  Are there any cases where we may want to track both the IOC version but also something about the version of the hardware underneath (could we extend this to also picking the right version of `cam` automatically?)?  I'm not sure that there is a good case for any of those, but it seems better to sort that out now than discover in 6mo we really did have a compelling case.

Is there anyway we can avoid having to list the base class twice? Maybe put a class attribute on one of the base classes saying it _may_ be sub-classed in a versioned way and then walk the base-classes until you find that flag?   I think that may allow us to be much terser in the init and allow a path to versioning many things.

Something like

```python
VERSION_ME_MAYBE = object()

class Base:

    def __init_subclass__(cls, ioc_version=None, **kwargs):
        super().__init_subclass__(**kwargs)
        if 'ioc_version' in cls.__dict__:
            cls.extant_versions = {}
        
        if ioc_version is not None:
            for bc in cls.mro():
                if bc.__dict__.get('ioc_version', None) is VERSION_ME_MAYBE:
                    base_class = bc
                    break
            else:
                raise TypeError("Nothing in MRO has a version")
            cls.ioc_version = ioc_version
            cls.ioc_class = bc

            bc.extant_versions[ioc_version] = cls
            

class A(Base):
    ioc_version = VERSION_ME_MAYBE



class B(A, ioc_version=1):
    ...


class C(A, ioc_version=2):
    ...


class D(A, ioc_version=3):
    ...

print(A.extant_versions)
```

Now that I have typed that out, you could probably invert the logic a bit and look for any class member that are `VERSION_ME_MAYBE` and then automatically pop out the right kwargs and build the right secondary data etc...

klauer:
For versioning multiple things - how about keeping `version` simple and requiring the user to pass in a tuple (or namedtuple) with whatever is to be versioned?

For not specifying the base class multiple times, take the complex example of AreaDetector:

```python
class PluginBase(version=...):
    ...
class FilePlugin(PluginBase, version=...):
    ...
class HDF5Plugin(FilePlugin, version=...)
    ...
```

All of these classes are versioned. However, `HDF5Plugin` is not a version of `FilePlugin`. `FilePlugin` is not a version of `PluginBase`.

I haven't found a good way to automatically determine which class is being versioned, even in the case of simple inheritance like this, without explicitly stating the base class that defines the first version. Multiple inheritance further muddies the waters.

ZLLentz:
I'm not sure there's a compelling use case for multiple versions to be tracked, and I don't think this is just because my head is spinning thinking about the complexity it adds to the syntax. In the EPICS/ophyd context, the IOC and the hardware are essentially the same thing. Either the IOC takes care of the hardware differences for you, or there is a different IOC that needs a differently-named class. The main use case here is to reflect that different beamlines and different facilities can have slightly different versions of the same software for controlling the same devices. We can already do this without this per-class version tagging, but this makes it very explicit which version the class is designed to connect to.

For multi-versioned classes, you can always use two subdevices that are properly single-versioned.

klauer:
I think @ZLLentz 's suggestions are quite reasonable. Any objections to moving in that direction?

danielballan:
> For multi-versioned classes, you can always use two subdevices that are properly single-versioned.

I had the same thought. I think this approach would be cleaner and clearer than anything we could come up with my for "multi-version syntax" anyway.

> Any objections to moving in that direction?

None. I think we should move ahead now.

tacaswell:
```python
VERSION_ME_MAYBE = object()

class Base:

    def __init_subclass__(cls, version=None, **kwargs):
        super().__init_subclass__(**kwargs)
        if 'version' in cls.__dict__:
            cls.extant_versions = set()
        
        if version is not None:
            for bc in cls.mro():
                if bc.__dict__.get('version', None) is VERSION_ME_MAYBE:
                    base_class = bc
                    break
            else:
                raise TypeError("Nothing in MRO has a version")
            cls.version = version
            cls.ioc_class = bc

            bc.extant_versions.add((version, cls))

class ADIOCVersionBase(Base):
    version = VERSION_ME_MAYBE
            

class PluginBase(ADIOCVersionBase, version=(2, 3)):
    ... 

class FilePlugin(PluginBase, version=(2, 4)):
    ...

class HDF5PluginBase(FilePlugin):
    ...

class TIFFPluginBase(FilePlugin):
    ...    
    
class HDF5Plugin(HDF5PluginBase, version=(2, 5)):
    ...

class HDF5Plugin2(HDF5PluginBase, version=(2, 6)):
    ...

class HDF5Plugin3(HDF5PluginBase, version=(2, 7)):
    ...

class TIFFPlugin(TIFFPluginBase, version=(2, 5)):
    ...

class TIFFPlugin2(TIFFPluginBase, version=(2, 6)):
    ...

class TIFFPlugin3(TIFFPluginBase, version=(2, 7)):
    ...



def get_compatible(target_class, ver):
    for v, klass in target_class.extant_versions:
        if ver == v and issubclass(klass, target_class):
            return klass
    else:
        raise ValueError("No acceptable class")

    
def get_all_compatible(target_class, ver):
    out = []
    for v, klass in target_class.extant_versions:
        if issubclass(klass, target_class) and ver == v:
            out.append(klass)
    return out

```

which gives you things like

```python
n [54]: get_all_compatible(FilePlugin, (2, 7))
Out[54]: [__main__.TIFFPlugin3, __main__.HDF5Plugin3]

In [55]: get_all_compatible(FilePlugin, (2, 6))
Out[55]: [__main__.HDF5Plugin2, __main__.TIFFPlugin2]


In [61]: get_compatible(HDF5PluginBase, (2, 10))
ValueError                                Traceback (most recent call last)
<ipython-input-61-75a61862e7c8> in <module>()
----> 1 get_compatible(HDF5PluginBase, (2, 10))

~/src/p/notebook/timetracking/.org/sw.org[*Org Src sw.org[ python ]*] in get_compatible(target_class, ver)

ValueError: No acceptable class

In [62]: get_compatible(HDF5PluginBase, (2, 7))
Out[62]: __main__.HDF5Plugin3

In [63]: get_compatible(HDF5PluginBase, (2, 6))
Out[63]: __main__.HDF5Plugin2

In [64]: 
```

klauer:
I like the idea in general, @tacaswell. These still stick out to me as negatives:

1. Mixing class kwargs and class attributes feels a bit off. I think it should be either all class kwargs or all class attributes.
2. While the user can request compatible versions of a specific class, the class itself has no idea which set of versions applies to it. That means it would require the caller to know that the corresponding "version base" of `HDF5Plugin3` is `HDF5PluginBase` and not `FilePlugin` or `PluginBase`. That's where the explicitness of `version_of` came in in the current PR code.

PR #664, opened 2018-12-18
ENH: add log.py and update __init__.py

awalter-bnl:
This duplicates the `bluesky.log.py` file to `ophyd.log.py` It adds a default logging handler to ophyd at import time and provides a convenience funciton `ophyd.set_handler`to configure that handler to adress common use cases. This brings it into line with blueksy and caproto.

### DIFF
--- a/ophyd/__init__.py
+++ b/ophyd/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import types
 import os
+from ophyd.log import set_handler
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -1,0 +1,193 @@
+# The LogFormatter is adapted light from tornado, which is licensed under
+# Apache 2.0. See other_licenses/ in the repository directory.
+
+import logging
+import sys
+try:
+    import colorama
+    colorama.init()
+except ImportError:
+    colorama = None
+try:
+    import curses
+except ImportError:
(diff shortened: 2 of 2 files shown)


ophyd/__init__.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import types
 import os
+from ophyd log import set_handler
```
danielballan: This is a `SyntaxError`.

awalter-bnl: Sorry, I forgot to make the final commit before opening the PR.

awalter-bnl:
There is a build error on the python 3.6 build, but all the others passed. I will restart it and see if it passes ok or not

awalter-bnl:
Only the codecov test are failing the Travis ones now all pass.

klauer:
@awalter-bnl can you squash into one commit, and edit your commit messages such that they follow the standards of this repo? This should be an `ENH` tag.

awalter-bnl:
Thanks for the comments @klauer , I have squashed to one commit and it appears that @mrakitin updated the commit message for me (thanks @mrakitin ).

I was not aware of any standard for commit messages in any of our repos, is this standard documented somewhere for reference?

danielballan:
The commit prefixes are a convention that many projects adopt from scipy: https://docs.scipy.org/doc/numpy-1.13.0/dev/gitwash/development_workflow.html

We should add a CONTRIBUTING.rst to this and all repos.

PR #674, opened 2019-01-18
ENH: support all areaDetector plugins versions 1-9-1 to R3-4

klauer:
I may have forgotten some details here, as this branch was done around the time of the proposal of #662. Apologies if anything is unclear.

Notes
------
* Largely auto-generated
* Allows specification of `version_type` only, for use in base plugin classes
* Adds`CommonPlugin` class hierarchy, allowing for easy access to common plugins specified by `commonPlugins.cmd` (or `EXAMPLE_commonPlugins.cmd` in some versions)
* See PV coverage below
~* Adds `PluginVersion` class hierarchy, allowing for easy access to specific versions of plugins compatible with an ADCore release~

`ophyd.areadetector.plugins` cleanup
---------------
* Removes the mistake that was the `C` abbreviation of `Component` (now `Cpt`)
* Adds the required versioning
* Uses `kind` instead of attr lists
* Refactors `ad_group` into much more readable `DDC_SignalWithRBV`-style partials

TODO
------
* Detectors/cams to follow in another PR

What a messy class hierarchy, ew!
-----------------------------------
To those who have that comment, this solution:
* Allows for code re-use such that methods can be implemented at the correct points in the versioning scheme for either (1) a specific plugin, (2) a specific file plugin, or (3) all plugins of a version.
* The style lends toward seeing an easy "diff" of what changed from one version to the next in a fairly succinct manner
* The versioning scheme makes selection of these classes significantly easier from a user or coding standpoint

If the above isn't convincing enough, and you have a better idea and time to implement it, by all means go for it.

Coverage
----------
Coverage of PVs versus those in a standard AreaDetector IOC using the example common plugins (as is fairly standard practice) are as follows:
```
R2-0 total pvs in ophyd 2102 of 2102 : PVs from IOC 100.0 %
R2-1 total pvs in ophyd 2193 of 2193 : PVs from IOC 100.0 %
R2-2 total pvs in ophyd 2868 of 2868 : PVs from IOC 100.0 %
R2-3 total pvs in ophyd 2913 of 2913 : PVs from IOC 100.0 %
R2-4 total pvs in ophyd 2913 of 2913 : PVs from IOC 100.0 %
R2-5 total pvs in ophyd 3039 of 3039 : PVs from IOC 100.0 %
R2-6 total pvs in ophyd 4099 of 4099 : PVs from IOC 100.0 %
R3-1 total pvs in ophyd 4746 of 4746 : PVs from IOC 100.0 %
R3-2 total pvs in ophyd 4757 of 4757 : PVs from IOC 100.0 %
R3-3 total pvs in ophyd 5555 of 5555 : PVs from IOC 100.0 %
R3-4 total pvs in ophyd 5982 of 5982 : PVs from IOC 100.0 %
```


danielballan:
Wow. https://www.youtube.com/watch?v=KlLMlJ2tDkg

ZLLentz:
Initial thoughts:
- Wow! So many versions!
- A bunch of this code is auto-generated. Should the code that generates it be part of this repo?
- Not sure on organization yet

klauer:
> A bunch of this code is auto-generated. Should the code that generates it be part of this repo?

I'd rather it not be, since cleaning it up would require much more effort than I'd like. That said, incremental additions shouldn't be terribly difficult to do manually as new versions are released. Doing 10 versions at a time is where automation became a necessity.

klauer:
No opinions on organization, @tacaswell?

ZLLentz:
Misc thoughts on current organization:

`common_plugins`: ophyd devices that contain all common plugins for each ophyd version

- I think this is good and makes sense as is

`new_plugins`: all versions of the individual plugins above 1-9-1
`plugins`: all 1-9-1 versions of the plugins

- `new_plugins` feels like a placeholder name
- Having these separated confuses me
- Perhaps it should just all be in a massive plugins file? Or is that unmanageable?

`plugin_versions`: collections of all plugin versions, grouped by adbase version

- I think this is good and makes sense as is

klauer:
Thanks, @ZLLentz! `new_plugins` was indeed a placeholder. It can be merged with `plugins`, if people think that's best. The only other obvious solution to me was separate modules for each plugin type. I'm not sold on either as being superior.

ophyd/device.py
```diff
@@ -909,6 +909,10 @@ def walk_subdevice_classes(cls):
         '''
         for attr in cls._sub_devices:
             cpt = getattr(cls, attr)
+            if cpt is None:
+                # Subclasses can override this, making this None...
```
danielballan: This is a useful escape hatch. Good call.

ophyd/device.py
```diff
@@ -812,7 +812,7 @@ def _initialize_device(cls):
 
         # The namedtuple associated with the device
         cls._device_tuple = namedtuple(f'{cls.__name__}Tuple',
-                                       [comp for comp in cls.component_names
+                                       [comp for comp in cls.component_names[:254]
```
danielballan: I have no guess about why this would be sliced. Can you add a comment?

danielballan:
One possibility is a plugins _package_ with a module per plugin. I think that would work out fairly cleanly. (We would need the requisite back-compat imports into `plugins/__init__.py` of course.

But given the auto-generated nature of this code, "one giant module" is also fine with me.

ophyd/device.py
klauer: Will do. `namedtuple` is limited to 254 items, believe it or not. That means that prior to this commit, any device with more than that could not even be created in ophyd.

klauer:
Thanks, @danielballan - I am definitely looking for opinions on organization. `new_plugins` is merely a placeholder.

So far, it seems like the votes are:

One giant plugins file
- @ZLLentz 
- maybe @danielballan 

Separate file per plugin
- maybe @danielballan 

Undecided (or no comment)
- @klauer 
- @tacaswell

danielballan:
Correct summary. I think either option is fine — and we can change our mind without breaking API if we really want to!

ophyd/device.py
tacaswell: This is 'fixed' in py3.7+ (https://stackoverflow.com/a/48051450/380231)

ophyd/areadetector/plugin_versions.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+class PluginVersions_V32(PluginVersions, version=(3, 2), version_of=PluginVersions):
+    CircularBuffPlugin = new_plugins.CircularBuffPlugin_V31
+    ColorConvPlugin = new_plugins.ColorConvPlugin_V31
+    CommonAttributePlugin = _make_common_attribute('CommonAttributePlugin_V32', (3, 2))
+    CommonGatherPlugin = _make_common_gather('CommonGatherPlugin_V32', (3, 2))
+    CommonOverlayPlugin = _make_common_overlay('CommonOverlayPlugin_V32', (3, 2))
+    CommonROIStatPlugin = _make_common_roistat('CommonROIStatPlugin_V32', (3, 2))
+    FFTPlugin = new_plugins.FFTPlugin_V31
```
tacaswell: Is this expected to be mixed versions?

tacaswell:
To check my understanding:

 - common_plugins -> `Devices` per version that you could mix into a user-space `Device` to get a functioning AD
 - new_plugins.py -> the actual implementation of the plugin classes
 - plugin_versions.py -> namespaces of all the plugins grouped by version.
 - plugins.py -> the existing source of plugins (which were implicitly 1.9.1) 

I think `new_plugins.py` should be made private, `plugin_versions` should be renamed to `versioned_plugins.py`, `plugins.py` should be deprecated (that is going to be fun!), and we need a better name for `common_plugins` (as the same suggests it contains the common plugins, not `Devices` that have all the common plugins as children)

tacaswell:
This isn't tested, but I think something like the following will also work:

```python
from ophyd import Component

class VersionedComponent(Component):
    def __set_name__(self, owner, attr_name):
        super().__set_name__(owner, attr_name)
        self.cls = _select_version(self.cls, owner._class_info_['version'])
```

That way on the floor when the IOC gets updated it is (ideally) a 1 line change to update the version of all of the plugins.

How do we want to fold in the changes to the trigger method and the changed default value of 'blocking_callbacks' to this (a-la https://github.com/NSLS-II/nslsii/blob/master/nslsii/ad33.py )?

klauer:
Do you really think the dynamic version you've provided is clearer, @tacaswell? I personally like being able to look at the source code and see what it's doing, which your version completely obfuscates.

`VersionedComponent` was going to be something that came out of the versioning PR (it was part of my original proposal), but that discussion took so long that it never got that far.

tacaswell:
I don't find many screens of auto generated almost the same but different in subtle ways code particularly clear either.  At least with for loops you can see what the patterns are instead of trying to reverse engineer them.

I'm glad we all agree `VersionComponent` is a good idea.

Is

```python
from ophyd import Component

class VersionedComponent(Component):
    def __set_name__(self, owner, attr_name):
        super().__set_name__(owner, attr_name)
        try:
            self.cls = _select_version(self.cls, owner._class_info_['version'])
        ValueError:
            setattr(owner, attr_name, None)
```

to evil / magical?

klauer:
Taking most of @tacaswell's suggestions; do not merge as there are massive PV coverage-related issues.

klauer:
OK, I believe I have fixed @tacaswell's dynamic generation of common plugins/plugin versions. Since I was going over all of this again, I've also increased the common plugin PV coverage to 100% across the board (it required adding a PluginBase for R2-5, as that's when QueueSize was introduced - not R2-6 as erroneously noted before).

ophyd/ophydobj.py
```diff
@@ -104,10 +107,25 @@ def __init_subclass__(cls, version=None, version_of=None,
...
+                version_of=version_of
+            )
             return
 
         if version_of is None:
             versions = {}
+            version_of = cls
         else:
             versions = version_of._class_info_['versions']
             version_type = version_of._class_info_['version_type']
```
tacaswell: We should probably check that `version_type` was not also passed in (or if it was it matches it's parent)?

ophyd/device.py
tacaswell: Do we still need this with https://github.com/NSLS-II/ophyd/pull/688 merged?

ophyd/ophydobj.py
klauer: 👍

ophyd/device.py
klauer: It can be removed

ophyd/ophydobj.py
```diff
@@ -6,6 +6,26 @@
 from enum import IntFlag
 
 
+module_logger = logging.getLogger(__name__)
+
+
+def select_version(cls, version):
+    """Select highest version of a class less that the requested version
```
klauer:
```suggestion
    """Select the closest version of a class compatible with the requested version
```

ophyd/ophydobj.py
```diff
@@ -6,6 +6,26 @@
...
+
+    Parameters
+    ----------
+    cls : type
+        The base class to find a version child  of
```
klauer:
```suggestion
        The base class to find a version of
```

tacaswell:
I am :+1: on this going in on the technical side of things, but I think there are a few API questions that should get more eyes than @klauer and I

### how to get access to a 'versioned namespace' of the plugins

06d0055 implements a dict-of-SimpleNamespace and 5892bc1 implements a versioned ophyd object that is the namespace.

In either case, the imagining use would be

```python

plugin_ns = select_version(Plugins, (3, 2)
plugin_ns = versioned_plugins[(3, 2)]

class MyAD(ADBase):
    tiff = Cpt(plugin_ns.TIFFPlugin)
    ....
```

### names of `CommonPlugins` and `Plugins`

The first is a proper `Device` with a coherent set of plugins an components that you can mix into your own device, the second is a namespace of coherently versioned plugins. The names collide too much in (at least my) head space.  `CommenPlugins` matches some of the naming schemes 



```python

class MyDet(SomeDetWithCam, CommonPlugins):
   ....
```

### should `CommonPlugins` include the file writing ones?

The classes for the file plugins do not include the Asset Document minting logic

### how to expose access to the `*_VXXX` classes

Currently systematically named `Plugins` and `CommonPlugins` are exposed in `__all__` all the way up through to the top-level ophyd namespace, where as for the individual plugins, only the non-qualified base class (which is implicitly v1.9.1) is.

Do we want to:
 - leave it as is
 - only expose the un-qualified base class of `CommonPlugins` and `Plugins` (and _force_ the use of `select_version` on them)
 - expose all of the VXX classes out.

danielballan:
Addressed @tacaswell's comments out of order because it will be easier to follow:

> names of `CommonPlugins` and `Plugins`

I have renamed `Plugins` to `PluginNamespace`. I hope that is satisfactory. It's a sensible but potentially confusing aspect of the design that will require good documentation, and optimizing its name is only going to do so much good here.

*Commit pushed by Dan*

> how to get access to a 'versioned namespace' of the plugins

We will support these ways of getting, for example, `StatsPlugin_V34`:

1. I have the base class; I want a specific version.

```py
In [2]: from ophyd import select_version

In [3]: from ophyd import StatsPlugin

In [4]: select_version(StatsPlugin, (3, 4))
Out[4]: ophyd.areadetector.plugins.StatsPlugin_V34
```

2. I want to directly import a class from the module where it is defined.

```py
In [5]: from ophyd.areadetector.plugins import StatsPlugin_V34 
# not available from ophyd or ophyd.areadetector though
```

3. I have a versioned `PluginNamespace_V34`; I want a specific plugin from it.

```py
In [6]: from ophyd import PluginNamespace_V34

In [7]: PluginNamespace_V34.StatsPlugin
Out[7]: ophyd.areadetector.plugins.StatsPlugin_V34
```

4.I have `PluginNamespace`, and I want to get a specific version and plugin.

```
In [8]: from ophyd import PluginNamespace

In [9]: select_version(PluginNamespace, (3, 4)).StatsPlugin
Out[9]: ophyd.areadetector.plugins.StatsPlugin_V34
```

This is a lot of ways to do it, but I can think of arguments for why each one is a reasonable way to do it, so I don't think there is a very good reason to making any of them not possible.

*Commit coming by Ken*

> how to expose access to the `*_VXXX` classes

They will be importable from `ophyd.areadetector.plugins` but they will not be hoisted into `ophyd.areadetector` or `ophyd`. in `ophyd/areadetector/__init__.py` we will explicitly hoist the old (1.9.1) classes for back-compat. We will rely on documentation to make the new classes discoverable.

*Commit coming by Ken*

> should `CommonPlugins` include the file-writing ones?

To be clear, the `CommonPlugins` currently _do_ contain file plugins --- but those file plugins _do not_ include the asset registry integration. The original reasons for keep the asset registry integration separate no longer apply, and it can probably be pushed into the file plugins contained in CommonPlugins now. This will require significant additional tests and likely some discussion. Let's handle it in a separate PR.

*No action required for this PR*

PR #675, opened 2019-01-19
MNT: import ABCs from collections.abc

tacaswell:


klauer:
Apparently these are deprecated imports in 3.7+.

PR #677, opened 2019-01-24
ENH: NDDerivedSignal

### LINKED ISSUE #673: derived signal can't handle type change
{{@klauer added}} An example from LIX where a string `Signal` gets converted to a float `DerivedSignal`:

```python
class _MaxiPressure(DerivedSignal):
    def inverse(self, value):
        msg = value
        if msg == "UNDER":
            return 1.0e-3
        elif msg == "OVER":
            return 1.1e3
        else:
            return float(msg)

    def describe(self):
        desc = super().describe()
        desc[self.name]['dtype'] = 'float'
        return desc

    def put(self, value, **kwargs):
        raise Exception("gauge is read only")
```

teddyrendahl:
# Description
This introduces `NDDerivedSignal` a subclass of `DerivedSignal` to reshape a flat array into a shaped one based on either static values or other components of the class. This was made with the purpose of the supporting shaped image signals, but I believe it is general enough to apply to any situation where you have a flat array this is intended to be reshaped.

# Motivation
This is useful for a variety of reasons, my major motivation is for `typhon` which can not tell a flat array from an image without checking if we are looking at an `AreaDetector` class explicitly. However, this certainly makes things like a `LiveImage` callback far easier to write. 

Also closes #673, before `DerivedSignal.describe` used most of the original signals description. We now infer the `data_type` and `shape` from the value of our `DerivedSignal` so changing types and shapes will be properly reflected.

# Tests
Tests on both the base `NDDerivedSignal` implementation and on `ImagePlugin` itself.

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -24,11 +24,39 @@ class NDDerivedSignal(DerivedSignal):
     """
     DerivedSignal to shape a flattened array
 
-    The purpose of this class is to take a flattened array and shape in its'
+    The purpose of this class is to take a flattened array and shape in its
     proper form. The shape of the final array may be static in which case the
     shape and number of dimensions can be set as static integers. Otherwise,
-    other signals from this classes parent can inform what the proper shape of
-    the array
+     other signals from this classes parent can inform the proper shape of
+    the array.
+
+    Parameters
+    ----------
+    derived_from : str, ``ophyd.Signal``
+        Either a `str` that specifies the attribute of the parent we will get
+        the unshaped array or an ``ophyd.Signal`` itself.
+
+    shape: tuple
+        A tuple containing integers in the case of a static array or links to
+        ``Signal`` objects. The specifications of the signals follows the same
+        rules as the ``derived_from`` parameter
+
+    num_dimensions: int, str, ``Signal``
+        The number of dimensions of the array. This can either be a static
+        ``int`` or a ``Signal`` itself.
+
+    Example
+    -------
+    .. code:: python
+
+        class TwoDimensionalDetector(Device):
+
+            flat_array = Cpt(EpicsSignal, ':Array')
+            width = Cpt(EpicsSignalRO, ':Width')
+            height = Cpt(EpicsSignalRO, ':Height')
+            shaped_array = Cpt(NDDerivedSignal('flat_array',
+                                               shape=('height', 'width'),
+                                               num_dimensions=2)
     """
     def __init__(self, derived_from, shape, num_dimensions,
                  parent=None, **kwargs):
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -41,7 +41,7 @@ class NDDerivedSignal(DerivedSignal):
         ``Signal`` objects. The specifications of the signals follows the same
         rules as the ``derived_from`` parameter
 
-    num_dimensions: int, str, ``Signal``
+    num_dimensions: int, str, ``Signal``, optional
         The number of dimensions of the array. This can either be a static
         ``int`` or a ``Signal`` itself.
 
@@ -58,7 +58,7 @@ class NDDerivedSignal(DerivedSignal):
                                                shape=('height', 'width'),
                                                num_dimensions=2)
     """
-    def __init__(self, derived_from, shape, num_dimensions,
+    def __init__(self, derived_from, shape, num_dimensions=None,
                  parent=None, **kwargs):
         super().__init__(derived_from, parent=parent, **kwargs)
         # Assemble our shape of signals
@@ -69,6 +69,8 @@ class NDDerivedSignal(DerivedSignal):
             self._shape.append(dim)
         self._shape = tuple(self._shape)
         # Assemble ndims
+        if not num_dimensions:
+            num_dimensions = len(self._shape)
         if isinstance(num_dimensions, str):
             num_dimensions = getattr(parent, num_dimensions)
         self._num_dimensions = num_dimensions
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -47,7 +47,7 @@ class NDDerivedSignal(DerivedSignal):
 
         # Ensure callbacks are fired when array is reshaped
         for dim in self._shape + (self._num_dimensions, ):
-            if not isinstance(dim, (int, float)):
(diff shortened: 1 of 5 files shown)


ophyd/areadetector/base.py
```diff
@@ -19,6 +20,77 @@ def __init__(self, prefix, **kwargs):
         super().__init__(prefix + '_RBV', write_pv=prefix, **kwargs)
 
 
+class NDDerivedSignal(DerivedSignal):
+    """
+    DerivedSignal to shape a flattened array
+
+    The purpose of this class is to take a flattened array and shape in its'
```
tacaswell: Can you please write a numpydoc style docstring?  The `shape` input should be carefully explained.

ophyd/areadetector/base.py
```diff
@@ -19,6 +20,77 @@ def __init__(self, prefix, **kwargs):
...
+    shape and number of dimensions can be set as static integers. Otherwise,
+    other signals from this classes parent can inform what the proper shape of
+    the array
+    """
+    def __init__(self, derived_from, shape, num_dimensions,
+                 parent=None, **kwargs):
+        super().__init__(derived_from, parent=parent, **kwargs)
+        # Assemble our shape of signals
+        self._shape = []
+        for dim in shape:
```
tacaswell: This looks like it only grabs these once at init.  If these change later I don't think it will be reflected here.  The ability to track the shape seems important for looking at ex, the output of the ROI plugin.

ophyd/signal.py
```diff
@@ -453,9 +454,11 @@ def derived_from(self):
 
     def describe(self):
         '''Description based on the original signal description'''
-        desc = self._derived_from.describe()[self._derived_from.name]
+        desc = super().describe()[self.name]  # Description of this signal
```
tacaswell: Does this fix #673 as well?

ophyd/signal.py
```diff
@@ -453,9 +454,11 @@ def derived_from(self):
...
         desc['derived_from'] = self._derived_from.name
-        return {self.name: desc}
+        # Description of the derived signal 
+        derived_desc = self._derived_from.describe()[self._derived_from.name]
+        return {self.name: dict(ChainMap(desc, derived_desc))}
```
tacaswell: do we need to flatten the chainmap?  If so, might as well do `derived_desc.update(desc)`

ophyd/signal.py
tacaswell: :sheep: reading comprehension of the PR message body fail.

ophyd/areadetector/base.py
klauer:
```suggestion
    The purpose of this class is to take a flattened array and shape in its
```

teddyrendahl: Will do.

ophyd/areadetector/base.py
```diff
@@ -19,6 +20,77 @@ def __init__(self, prefix, **kwargs):
...
 
+class NDDerivedSignal(DerivedSignal):
+    """
+    DerivedSignal to shape a flattened array
+
+    The purpose of this class is to take a flattened array and shape in its'
+    proper form. The shape of the final array may be static in which case the
+    shape and number of dimensions can be set as static integers. Otherwise,
+    other signals from this classes parent can inform what the proper shape of
+    the array
```
klauer: Docstring trails off here

ophyd/areadetector/base.py
```diff
@@ -19,6 +20,77 @@ def __init__(self, prefix, **kwargs):
...
+            self._shape.append(dim)
+        self._shape = tuple(self._shape)
+        # Assemble ndims
+        if isinstance(num_dimensions, str):
+            num_dimensions = getattr(parent, num_dimensions)
+        self._num_dimensions = num_dimensions
+
+        # Ensure callbacks are fired when array is reshaped
+        for dim in self._shape + (self._num_dimensions, ):
+            if not isinstance(dim, (int, float)):
```
klauer:
```suggestion
            if not isinstance(dim, int):
```

Float is not a valid type to reshape with

ophyd/areadetector/base.py
teddyrendahl: If you pass in signals every call to `derived_shape` will be be updated, they are only static if you pass in integers. We will also be updated of any changes to the array shape via the `_array_shape_callback`

ophyd/areadetector/base.py
```diff
@@ -19,6 +20,77 @@ def __init__(self, prefix, **kwargs):
...
+        self._shape = []
+        for dim in shape:
+            if isinstance(dim, str):
+                dim = getattr(parent, dim)
+            self._shape.append(dim)
+        self._shape = tuple(self._shape)
+        # Assemble ndims
+        if isinstance(num_dimensions, str):
+            num_dimensions = getattr(parent, num_dimensions)
+        self._num_dimensions = num_dimensions
```
klauer: I think `num_dimensions` could be optional, inferring from `shape` when not given.

ophyd/areadetector/plugins.py
```diff
@@ -283,6 +283,12 @@ class ImagePlugin(PluginBase):
     _plugin_type = 'NDPluginStdArrays'
 
     array_data = C(EpicsSignal, 'ArrayData')
+    shaped_image = C(NDDerivedSignal, 'array_data',
+                     shape=('array_size.height',
+                            'array_size.width',
+                            'array_size.depth')
```
klauer:
```suggestion
                            'array_size.depth'),
```

ophyd/areadetector/base.py
tacaswell: Sorry, I was reading this as the train was pulling into Jamacia and miss-understood :disappointed: .

tacaswell:
The test failures look real and caused by this PR.

I am a bit concerned that this automatically subscribes to the array PV an init. May be out of scope for this PR, but would it be better to push the subscription at https://github.com/NSLS-II/ophyd/blob/312da6721783d43debc9cbf5a797e5abb8920d33/ophyd/signal.py#L442-L447 to be when the user actually subscribes to the derived pv?

klauer:
Go for it, @tacaswell

teddyrendahl:
@tacaswell I think that is a very valid point and in the scope of this PR

teddyrendahl:
Power cycling to pickup #679

PR #679, opened 2019-01-25
CI: Split CL builds on Travis

ZLLentz:
Currently we run tests using `pyepics` and `caproto` in a build matrix of three python versions. This expands the build matrix to be 3 python versions x 3 control layers (`pyepics`, `caproto`, `p4p`). 

`caproto` is set to allowed fail because that's how the travis configuration is currently, and `p4p` is also allowed to fail because it's still a PR and somewhat experimental. Hopefully as these tests stabilize, they can be moved out of allowed failures.

Advantages:
- Easy to quickly differentiate control layer problems vs general ophyd problems
- Obvious to see when changes break specific control layers rather than all of them
- Control layers with long logs (caproto) don't break the 10,000 line travis limit for the other layers
- Easy to expand if we add support for `tango`, etc.

Disadvantages:
- The build matrix is getting big
- Not sure if we can combine coverage from the different control layers

Questions:
- How does codecov respond when 9 builds are associated with the same PR?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,41 @@ cache:
 matrix:
   include:
     - python: 3.6
-      env: BUILD_DOCS=true
+      env:
+        - BUILD_DOCS=true
+        - TEST_CL=pyepics
+    - python: 3.6
+      env:
+        - BUILD_DOCS=
+        - TEST_CL=caproto
+    - python: 3.6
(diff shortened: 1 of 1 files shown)


klauer:
Thaaaank youuu, @ZLLentz!

PR #681, opened 2019-01-29
 My suggestion for reporting missing self.value

PierreSchnizer:
If the member 'value' of :class:`ophyd.signal.Signal` has never been set, 
its value is None. This value can thus not be turned to a JSON representation.
The user is presented with a message informing that None can not be converted
to value useful in bluesky.

I propose to add a check in the describe method if value is not None. Then the
user can be informed  which variable is the culprit.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -301,6 +301,12 @@ class Signal(OphydObject):
             with the ``event_model.event_descriptor.data_key`` schema.
         """
         val = self.value
+        if val is None:
+            raise AssertionError(
+                f'Description can not yet been made for {self.name} as '
+                f'self.value == None. This typically means, that the '
+                f'value was never set.'
+                )
         return {self.name: {'source': 'SIM:{}'.format(self.name),
                             'dtype': data_type(val),
                             'shape': data_shape(val)}}
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -303,7 +303,7 @@ class Signal(OphydObject):
         val = self.value
         if val is None:
             raise NonPVValue(
-                f'Description can not yet been made for "{self.name}" as '
+                f'Description can not yet been made for {self.name!r} as '
                 f'self.value = None. This typically means, that the '
                 f'value was never set.'
                 )
@@ -311,7 +311,7 @@ class Signal(OphydObject):
         try:
             dtype = data_type(val)
         except Exception as ex:
-            msg = f'Conversion error for self.value for variable "{self.name}".'
+            msg = f'Type inference failed for {self.value!r} for {self.name!r}.'
             raise ex.__class__(msg) from ex
 
         return {self.name: {'source': 'SIM:{}'.format(self.name),
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -302,14 +302,21 @@ class Signal(OphydObject):
         """
         val = self.value
         try:
-            dtype =  data_type(val)
-        except NonPVValue as npv:
-            if val is None:
+            dtype = data_type(val)
+        except Exception as npv:
+            if isinstance(npv, NonPVValue) and val is None:
+                msg, = npv.args
                 raise NonPVValue(
                     f'Description can not yet been made for "{self.name}" as '
                     f'self.value = None. This typically means, that the '
                     f'value was never set.'
                 )
+
+            # val was not None so lets give all the knowledge to the user
+            txt = f'Conversion error for self.value for variable "{self.name}":'
+            args = (txt,) + npv.args
+            npv.args = args
+            raise npv
 
         return {self.name: {'source': 'SIM:{}'.format(self.name),
                             'dtype': dtype,
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -301,22 +301,18 @@ class Signal(OphydObject):
             with the ``event_model.event_descriptor.data_key`` schema.
         """
         val = self.value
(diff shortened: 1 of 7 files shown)


PierreSchnizer:
First time using pytest. Tests converted

@mrakitin Is there a more specifc ophyd exception to raise?

ophyd/tests/test_missing_readback.py
```diff
@@ -0,0 +1,32 @@
...
+    """Just to create a device
+    """
+    signal = Cpt(sim.FakeEpicsSignalRO, 'a_fake_signal')
+
+
+def _get_fake_signal():
+    return ADevice(name = "fake_device")
+
+def test_describe_fail():
+    """If readback was never set, does descirbe fail?
```
mrakitin:
```suggestion
    """If readback was never set, does describe fail?
```

ophyd/tests/test_missing_readback.py
```diff
@@ -0,0 +1,32 @@
...
+
+    I guess that's a bug
+    """
+    # self.assertRaises(ValueError, self.fake_signal.describe)
+    fs = _get_fake_signal()
+    with pytest.raises(AssertionError):
+        fs.describe()
+
+def test_describe_after_set():
+    """If readback was set does descirbe fail?
```
mrakitin:
```suggestion
    """If readback was set does describe fail?
```

ophyd/tests/test_missing_readback.py
```diff
@@ -0,0 +1,32 @@
...
+
+    SHould work
```
mrakitin:
```suggestion
    Should work
```

PierreSchnizer:
I made modifications to suggest a dedicated error. Furthermore I now try the conversion first and 
catch the error if it occurs. 

Personally I am not very comfortable with excepting ValueError. It could be raised further down the
line. Then in signal.Signal.describe one can not be sure if the error was raised by ValueError

ophyd/utils/epics_pvs.py
```diff
@@ -319,7 +319,7 @@ def data_type(val):
         if isinstance(val, py_types):
             return json_type
 
-    raise ValueError(
+    raise NonPVValue(
```
tacaswell: Might want a second new sub-class of ValueError here.

ophyd/signal.py
```diff
@@ -301,8 +301,25 @@ def describe(self):
             with the ``event_model.event_descriptor.data_key`` schema.
         """
         val = self.value
+        try:
+            dtype = data_type(val)
+        except Exception as npv:
```
tacaswell: I suggest doing this as

```python
if val is None:
    raise NonPVValue(...)
try:
    dtype = data_type(val)
except ValueError as ex:
   raise ValueError(f'...{self.name}...') from ex
```

tacaswell: We might also want our own sentinel rather than None?

teddyrendahl:
Is this the better route than having `ophyd.Signal` default to 0?

klauer:
@teddyrendahl 0 is an integer, `None` is not (as you well know), which may lead to the type of the `Signal` changing inadvertently when it does actually get initialized.

I'd be more inclined to push the responsibility of handling a `None` value up to the caller somehow.  Is it really ophyd's problem that `None` isn't always serializable? I'd guess that's not possible within the constraints of the current NSLS-II event/data model?

teddyrendahl:
A good point. I'd argue that the value is already switching type from `NoneType` to whatever is is set afterwards, but I do see that `NoneType` has a more significant connotation. 

I would tend to agree however that `ophyd` should report a `NoneType` in describe and the consumer of the `describe` document should be able to decide what to do with it.

tacaswell:
> I'd guess that's not possible within the constraints of the current NSLS-II event/data model?

The responsibility for generating the correct data type information lives in `describe` method, I think the buck stops with ophyd on this one.  I don't think we have an sensible type to label `None` as and Iwould rather not add one because it does not have a pair in EPICS and not all languages we may want to target as consumers may have the notion.

You could also have a 'latch' on Signal that doing anything to it before the first set would raise, but that would not prevent people from setting `None` which does not get us around the actual problem. 

> Is this the better route than having `ophyd.Signal` default to 0?

Agree with @klauer on this one, if the application really cares about a default value, it should explicitly set one when creating the `Signal`.

This ideally this would only apply to the 'soft' signals, but I think that if an `EpicsSignal` has `None` as its value there are bigger problems so I'm not super worried about that (and I have advocated for a while (but not putting in the effort to _do_ it) that `EpicsSignal` should not be inferring the datatype by looking at the python side value but by asking the underlying readpv which also side-steps the problem there). 

If we do this in `Signal` we should also do it in the 3 or 4 other places we use `datatype` in `signal.py`.

Thanks for you work on this @PierreSchnizer !

tacaswell:
sorry my prose is terrible, it has been a long day....

ophyd/utils/epics_pvs.py
PierreSchnizer: Sorry for the late comment. I am new to GitHub...

I am a bit hesitant on creating exception class names. My impression is that a one which is wisely chosen facilitates usage.

klauer: Subclassing built-in Exceptions for a package is good practice, as we can make - and catch - very specific errors.

`ValueError` could be raised just about anywhere, but `NonPVValue` could only come from here.

PR #682, opened 2019-02-04
API/REF: Make simulated hw into Devices.

awalter-bnl:
I have added the `configure` method to SynSignal, so that each of the simulated detectors in `hw` have this method. 

This method is required so that we can deploy a 'plan' that includes ` yield from bluesky.plan_stubs.configure`. This allows for the testing of callbacks (like the suitcase Serializers) with plans that have mulitple descriptors per stream.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,7 @@ pip-delete-this-directory.txt
 # Unit test / coverage reports
 htmlcov/
 .tox/
-.coverage
+.coverage*
 .cache
 .pytest_cache
 nosetests.xml
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -144,7 +144,7 @@ class SynSignal(Signal):
                   d: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         '''Configure the device for something during a run
         '''
-        if not d:
+        if d:
             warnings.warn('The `configure` method on `ophyd.sim.SynSignal` '
                           'devices is a no-op method, all information'
                           'included in the dictionary `d` is ignored')
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -19,7 +19,7 @@ from tempfile import mkdtemp
 from .signal import Signal, EpicsSignal, EpicsSignalRO
 from .areadetector.base import EpicsSignalWithRBV
 from .status import DeviceStatus, StatusBase
-from .device import (Device, Component, Component as C,
+from .device import (Device, Component, Component as Cpt,
                      DynamicDeviceComponent as DDC, Kind)
 from types import SimpleNamespace
 from .pseudopos import (PseudoPositioner, PseudoSingle,
@@ -857,14 +857,14 @@ class InvariantSignal(SynSignal):
 
 
 class SPseudo3x3(PseudoPositioner):
-    pseudo1 = C(PseudoSingle, limits=(-10, 10), egu='a', kind=Kind.hinted)
-    pseudo2 = C(PseudoSingle, limits=(-10, 10), egu='b', kind=Kind.hinted)
-    pseudo3 = C(PseudoSingle, limits=None, egu='c', kind=Kind.hinted)
-    real1 = C(SoftPositioner, init_pos=0)
-    real2 = C(SoftPositioner, init_pos=0)
-    real3 = C(SoftPositioner, init_pos=0)
+    pseudo1 = Cpt(PseudoSingle, limits=(-10, 10), egu='a', kind=Kind.hinted)
+    pseudo2 = Cpt(PseudoSingle, limits=(-10, 10), egu='b', kind=Kind.hinted)
+    pseudo3 = Cpt(PseudoSingle, limits=None, egu='c', kind=Kind.hinted)
+    real1 = Cpt(SoftPositioner, init_pos=0)
+    real2 = Cpt(SoftPositioner, init_pos=0)
+    real3 = Cpt(SoftPositioner, init_pos=0)
 
-    sig = C(Signal, value=0)
+    sig = Cpt(Signal, value=0)
 
     @pseudo_position_argument
     def forward(self, pseudo_pos):
@@ -884,10 +884,10 @@ class SPseudo3x3(PseudoPositioner):
 
 
 class SPseudo1x3(PseudoPositioner):
-    pseudo1 = C(PseudoSingle, limits=(-10, 10), kind=Kind.hinted)
-    real1 = C(SoftPositioner, init_pos=0)
-    real2 = C(SoftPositioner, init_pos=0)
-    real3 = C(SoftPositioner, init_pos=0)
+    pseudo1 = Cpt(PseudoSingle, limits=(-10, 10), kind=Kind.hinted)
+    real1 = Cpt(SoftPositioner, init_pos=0)
+    real2 = Cpt(SoftPositioner, init_pos=0)
+    real3 = Cpt(SoftPositioner, init_pos=0)
 
     @pseudo_position_argument
     def forward(self, pseudo_pos):
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -437,7 +437,8 @@ class SynGauss(SynSignal):
 
         def func():
             m = self._motor.read()[self._motor_field]['value']
-            v = self.Imax * np.exp(-(m - self.center) ** 2 / (2 * self.sigma ** 2))
+            v = self.Imax * np.exp(-(m - self.center) ** 2 /
+                                   (2 * self.sigma ** 2))
             if self.noise == 'poisson':
                 v = int(self.random_state.poisson(np.round(v), 1))
             elif self.noise == 'uniform':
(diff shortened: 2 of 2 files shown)


danielballan:
I always thought of `configure` as a Device thing, like `stage` and `unstage`, not a Signal thing. I would like to hear from other contributors how they think about it before deciding this either way.

Another way to achieve what you need on the suitcase side would be to make a Device that contains these Signals and adding a `configure` method to _that_.

awalter-bnl:
I don't disagree with you @danielballan, but currently all of our simulated detectors inherit from `SynSignal`, not `Device` and this is the easiest path to get the required outcome. I thought I would implement this, and open the PR to generate discussion on how best to solve this, so I am open to other options.

tacaswell:
This ties into https://github.com/NSLS-II/ophyd/issues/678 about the distinction between Signal and Device. 

In the tests we are using `SynSignal` as a stand-in for when we would have a `Device` on the floor (because from bluesky's point of view they are 'the same'), but in this case there is something bluesky knows about that is a way to tell `Devices` and `Signals` apart.  

Given that it all synthetic we can not apply [the teddy criteria](https://github.com/NSLS-II/ophyd/issues/678#issuecomment-458669448) to sort out if this should be a device or a signal, this is the path of least resistance.

ophyd/sim.py
```diff
@@ -139,6 +140,14 @@ def get(self):
         # example.
         return super().get()
 
+    def configure(self,
+                  d: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+        '''Configure the device for something during a run
+        '''
+        old = self.read_configuration()
```
tacaswell: Maybe add a check that the dictionary is empty (as there is nothing _to_ configure)?

awalter-bnl: done

awalter-bnl:
In addition to adding the empty dict warning as requested I have also made some changes to pre-exisitng flake8 issues.

klauer:
`configure` is not part of the [`BlueskyInterface`](https://github.com/NSLS-II/ophyd/blob/442271c06e8384c18f67176a9cef691edf154631/ophyd/device.py#L398) and appears to lie in somewhat of a grey area.

We should review the methods that bluesky is expecting and make them a part of the `BlueskyInterface` as either required (or optional) members. If `configure` is deemed an optional one, then it's OK - albeit strange - that `SynSignal` implements it whereas all of the other `Signal`s classes in the entirety of ophyd do not.

My opinion: If this work is blocking progress, merge away. The above should be addressed sooner rather than later, though.

danielballan:
I have set up suitcase-utils to [use this branch as a test dependency](https://github.com/NSLS-II/suitcase-utils/pull/3/files#diff-f4c265073ea4274a80eca85440875e56R12) so this is not _blocking_ progress but we should aim to resolve it in the next couple weeks at most. Also, be sure not to delete this branch @awalter-bnl until suitcase-utils has been corresponding updated to rely on ophyd `master` or the next ophyd release.

awalter-bnl:
thanks for the heads up @danielballan I will try to remember not to delete the branch.

awalter-bnl:
Thanks @mrakitin for the update, this new device will be useful to add to the `detector_list` `pytest.fixture` in `suitcase-utils.tests.conftest.py`

ophyd/sim.py
```diff
@@ -1240,6 +1257,10 @@ def hw(save_path=None):
     # area detector that directly stores image data in Event
     direct_img = SynSignal(func=lambda: np.array(np.ones((10, 10))),
                            name='img', labels={'detectors'})
+
+    direct_img_list = SynSignal(func=lambda: [[1] * 10] * 10,
+                                name='img', labels={'detectors'})
```
awalter-bnl:
```suggestion
                                name='img_list', labels={'detectors'})
```

awalter-bnl: @mrakitin I think we should have a different name for this detector to avoid confusion (see above)

mrakitin: Good point, @awalter-bnl, I don't mind to change it, however I followed the pattern as 2 detectors had the same name attribure (see the `img` det below). I think we then need to make them all unique.

awalter-bnl: @mrakitin you are correct, I made that suggestion in your PR to `suitcase-utils` :-)

tacaswell:
I think there are two issues here

 1. we probably should have included `configure` in the bluesky interface (at least as an optional method and may provide a reasonable way to squeeze setting the EGU and PREC into EpicsSignal API) as the RunEnigne has special behavior responding to the configure message (discarding the descriptor cache)
2. in practice we have ended up with very few 'bare' Signals being used in plans (the only one I can think of that is used commonly is the ring current, everything else (AD, scalar, quadem, any motor, ...) have _some_ sort of user configurable behavior and hence end up being a Device.  We (at least I) have been thinking of the thing is `hw` as a "toy beamline" but most of the things in there are actually `Signal` subclasses rather `Device` subclasses so the issue this is fixing is coming from that impedance miss-match.

We could re-do the contents of `hw` to 'promote' `SynGauss` and friends to be devices (at the cost of changing their types, confident we could do this without changing anything else about the API around them) or add a bunch more objects which are the Device version of them (but then we have 2 for not great reasons) or make a new `hw` fixture in suitcase-utils that has just `Device` subclasses, however none of those options are great.

There is already some really wacky stuff in here (such as [`collect_asset_docs`](https://github.com/NSLS-II/ophyd/blob/6cab3d2f2adbca2e4aab557bda80c1e2db175bfb/ophyd/sim.py#L773) on a `Signal` subclass) which is a legacy of the fact that much of this code was copied from bluesky where we were just duck-typing ophyd for testing.

I think this is the most pragmatic path forward.   I think of this module as a 'for-testing-only' space (hence being pretty cavalier about it), are these classes being used in production?

ophyd/sim.py
```diff
@@ -139,6 +140,19 @@ def get(self):
         # example.
         return super().get()
 
+    def configure(self,
+                  d: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+        '''Configure the device for something during a run
+        '''
+        if d:
+            warnings.warn('The `configure` method on `ophyd.sim.SynSignal` '
```
tacaswell: This should probably raise rather than warning.

klauer:
@tacaswell I don't believe you can set EGU and PREC except through accessing the corresponding record fields. That is to say, if you have `EpicsSignal('PV.VAL')` there is no `caput` that you could do to change precision or units, outside of making an additional channel/signal `EpicsSignal('PV.EGU')`.

Regardless of the above, I personally think the right path is making them into `Device`s. 

We use the simulated devices fairly extensively here at SLAC. @teddyrendahl and @ZLLentz would know better if such a change would break things for us.

teddyrendahl:
> I think this is the most pragmatic path forward. I think of this module as a 'for-testing-only' space (hence being pretty cavalier about it), are these classes being used in production?

Not used in production anywhere but we do use them in a few of our other test suites.

> Regardless of the above, I personally think the right path is making them into Devices.

I agree with @klauer. I am fine making them into full devices and am willing to pay the cost of updating our test suites if it ends up making a difference. If we go this route I think it would be interesting to have all of the parameters of `SynGauss` as `Signal` components in spirit of https://github.com/NSLS-II/ophyd/pull/655 as well. (largely so it will make a pretty `typhon` screen :smile:)

tacaswell:
> @tacaswell I don't believe you can set EGU and PREC except through accessing the corresponding record fields. That is to say, if you have `EpicsSignal('PV.VAL')` there is no `caput` that you could do to change precision or units, outside of making an additional channel/signal `EpicsSignal('PV.EGU')`.

`EpicsSignal` already holds 1 or 2 channels (when there are separate write / readback PVs) so I don't think it is too much of a stretch to do something like

```python
def configure(self, egu=None, prec=None):
    if prec is not None:
         prec_pv = self.pvname.partition('.')[0] + '.PREC'
         self.cl.get_pv(prec_pv).put(prec)
    ... 
``` 

modulo any caching / error handling logic.

I'm going to start pushing commits at this branch to do the Device conversion.

danielballan:
I think that the recent changes to this branch have broken all suitcases which (against our usual practice...) depend on this branch. https://travis-ci.org/NSLS-II/suitcase-cookiecutter/builds/512748508#L1970

.gitignore
```diff
@@ -35,7 +35,7 @@ pip-delete-this-directory.txt
 # Unit test / coverage reports
 htmlcov/
 .tox/
-.coverage
+.coverage*
```
danielballan: This change would cuase `.coveragerc` to be ignored. It should be reverted I think.

danielballan:
What's the status of this work? Reviewing the comments above, it looks like we're all on board with "rewrite test objects as Devices" but that there is more to do.

jklynch:
It would be great if this PR could be revived. It would make developing suitcases easier.

awalter-bnl:
Sorry @danielballan it appears I missed your comment on this from May. If the solution is to "rewrite test objects as Devices"  I don't have the time to undertake that amount of work, and history shows that having me attempt such wide-ranging PR's into Bluesky related repo's is a sure way to kill them. If someone else would like to take over that task I would be happy with that, or if there is a preference to just adds this with minimal changes I am happy to work through the minimal changes.

tacaswell:
Closing in favor of #762 which has a rebased and slightly extended version of this code and https://github.com/bluesky/suitcase-utils/pull/29 which de-couples suitcase-utils from the exact ophyd version.

PR #688, opened 2019-02-25
FIX: None-ified components in Device subclasses

### LINKED ISSUE #687: BUG: cannot use component_name=None to remove the component from a Device
@tacaswell and I met this issue while working on the PE detector for [NSLS-II QAS](https://github.com/NSLS-II-QAS/profile_collection/pull/2). When we attempt to list all of the available components of the `cam` component with `pe1.cam.summary()`, it throws an error that one of the PVs is not available. They use the ADCore v3.3.2, which has some PVs removed/added (see https://github.com/areaDetector/ADCore/commit/47c70bf10644c63955c00646eaffd881f8a1a9bc), but ophyd does not have this change, `pool_max_buffers` is still in the code: https://github.com/NSLS-II/ophyd/blob/442271c06e8384c18f67176a9cef691edf154631/ophyd/areadetector/cam.py#L54

An attempt of setting `pool_max_buffers = None` did not resolve the problem.

klauer:
Closes #687

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -785,6 +785,7 @@ class Device(BlueskyInterface, OphydObject):
         cls._sig_attrs = OrderedDict((attr, cpt)
                                      for base in base_devices
                                      for attr, cpt in base._sig_attrs.items()
+                                     if getattr(cls, attr) is not None
                                      )
 
         # map component classes to their attribute names from this class
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -879,7 +879,7 @@ class Device(BlueskyInterface, OphydObject):
             #  - Always include non-lazy components
             #  - Include a lazy if already instantiated OR requested with
             #    include_lazy
-            lazy_ok = cpt.lazy and (include_lazy or attr in self.__dict__)
+            lazy_ok = cpt.lazy and (include_lazy or attr in self._signals)
             should_walk = not cpt.lazy or lazy_ok
 
             if not should_walk:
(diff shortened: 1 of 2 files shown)


mrakitin:
Is it a problem with epics on travis?
```
cp: cannot stat '/home/travis/epics/R7.0.1.1/support/areadetector/3-2/ADCore/iocBoot/EXAMPLE_commonPlugins.cmd': No such file or directory
```

klauer:
@mrakitin sometimes the `curl` of the supporting libraries fails - expand the log a couple lines above (`$ bash "${CI_TOP}/install-from-release.sh" "${EPICS_ON_TRAVIS_URL}"`) and you'll see 'unexpected EOF'.

klauer:
Piggybacking another fix on this PR related to `walk` functions and laziness. As noted in the commit message:    
```__get__ is always called for descriptors. To see if a lazy signal has been instantiated, the `_signals` dictionary can be checked instead.``` and importantly, `__dict__` cannot be used in that case.

mrakitin:
> @mrakitin sometimes the `curl` of the supporting libraries fails - expand the log a couple lines above (`$ bash "${CI_TOP}/install-from-release.sh" "${EPICS_ON_TRAVIS_URL}"`) and you'll see 'unexpected EOF'.

Got it. Thanks, @klauer!

tacaswell:
Pushed a fix to the failing test.

mrakitin:
The tests pass after @tacaswell's fix. Merging.

PR #689, opened 2019-02-26
BUG: Fix infinite recursion observed from reading shaped_image

ronpandolfi:
As discussed in NIKEA: https://nikea.slack.com/archives/C7PNB3BUG/p1551207337008300

### DIFF
--- a/ophyd/areadetector/base.py
+++ b/ophyd/areadetector/base.py
@@ -105,8 +105,7 @@ class NDDerivedSignal(DerivedSignal):
         if not any(array_shape):
             raise RuntimeError(f"Invalid array size {self.derived_shape}")
 
-        array = self._derived_from.value
-        return np.array(array).reshape(array_shape)
+        return np.array(value).reshape(array_shape)
 
     def subscribe(self, callback, event_type=None, run=True):
         cid = super().subscribe(callback, event_type=event_type, run=run)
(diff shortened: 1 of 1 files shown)


ophyd/areadetector/base.py
```diff
@@ -105,7 +105,7 @@ def inverse(self, value):
         if not any(array_shape):
             raise RuntimeError(f"Invalid array size {self.derived_shape}")
 
-        array = self._derived_from.get()
+        array = self._derived_from.value
```
tacaswell: I think you want this to be `array = value` (the one that is passed in)

ophyd/areadetector/base.py
```diff
@@ -105,8 +105,7 @@ def inverse(self, value):
         if not any(array_shape):
             raise RuntimeError(f"Invalid array size {self.derived_shape}")
 
-        array = self._derived_from.get()
-        return np.array(array).reshape(array_shape)
+        return np.array(value).reshape(array_shape)
```
ronpandolfi: A bit simpler, then

teddyrendahl:
It wasn't you, but there is a rogue `print` statement in this function. Mind nixing it?

tacaswell:
Calling `.read()` on it should be enough?

mrakitin:
I guess so, reading a shaped image would be fine as it was the motivation for this PR.

PR #690, opened 2019-02-27
FIX: ScalerCH

prjemian:
`chname` is the EpicsSignal, `name` is a str and raises an exception here


tacaswell:
Thanks @prjemian !  Either we never actually ran this or there was a copy-paste bug when migrating this from a BL profile...

PR #693, opened 2019-02-27
FIX: monitor + get*() issue

klauer:
This PR aims to fix a callback-related issue created by the latest Signal refactor.

The Signal refactor intended to keep metadata information and values up-to-date with individual `.get()` and `.get_setpoint()` requests in `EpicsSignal`.  It also ran callbacks for value and metadata upon receiving new information.

DBR_CTRL Updates
----------------
First, note that EpicsSignal monitors using DBR_TIME on the control layer. If the `get()` requested a DBR_CTRL type by specifying `get(..., form='ctrl')`, for example, units and such would be updated internally.  There is no `DBR_` type which encapsulates all of the possible metadata.

The issue
---------
Callbacks being run by `.get()` on every call meant that an infinite recursionscenario was very easy to run into.

I discovered this while subscribing to `Plugin.nd_array_port` and calling `detector.get_asyn_digraph()`, which internally calls `plugin.nd_array_port.get()`, which spawned a callback, which caused my subscription to be run, ..., and so on and so forth.

The fix (1)
-------
* For consistency, update EpicsSignal internals regardless of the source of new information.
* For monitor callbacks (i.e., `_read_changed(..., from_monitor=True)`) subscriptions will be run.
* For `get()` calls (i.e., `_read_changed(..., from_monitor=False)`), subscriptions will not be run.

The fix (2)
-----------
An additionally important fix is the following:
* Metadata callbacks must be guaranteed to run in one thread

This is handled by the dispatcher 'metadata' thread.  Some new useful API is exposed through the dispatcher and helper `DispatcherThreadContext` objects.

Future
------
* Should we consider a `_run_subs_if_changed`-sort of utility method such that subscriptions are only run if anything has changed since the last update? This can easily be compared with the subscription argument cache stored on every ophyd object.
* Metadata + value should be stored in the same dictionary for better atomic updating...

### DIFF
--- a/ophyd/_dispatch.py
+++ b/ophyd/_dispatch.py
@@ -59,11 +59,52 @@ class _CallbackThread(threading.Thread):
         self.context = None
 
 
+class DispatcherThreadContext:
+    '''
+    A thread context associated with a single Dispatcher event type
+
+    Parameters
+    ----------
+    dispatcher : Dispatcher
+    event_type : str
+
+    Attributes
+    ----------
+    dispatcher : Dispatcher
+    event_type : str
+    event_thread : _CallbackThread
+    '''
+
+    def __init__(self, dispatcher, event_type):
+        self.dispatcher = dispatcher
+        self.event_type = event_type
+        self.event_thread = None
+
+    def run(self, func, *args, **kwargs):
+        '''
+        If in the correct threading context, run func(*args, **kwargs) directly,
+        otherwise schedule it to be run in that thread.
+        '''
+        if self.event_thread is None:
+            self.event_thread = self.dispatcher._threads[self.event_type]
+
+        current_thread = threading.currentThread()
+        if current_thread is self.event_thread:
+            print('calling', func, args, kwargs)
+            func(*args, **kwargs)
+        else:
+            print('scheduling', func, args, kwargs)
+            self.event_thread.queue.put((func, args, kwargs))
+
+    __call__ = run
+
+
 class EventDispatcher:
     def __init__(self, *, context, logger, timeout=0.1,
                  thread_class=_CallbackThread, debug_monitor=False,
                  utility_threads=4):
         self._threads = {}
+        self._thread_contexts = {}
         self._thread_class = thread_class
         self._timeout = timeout
 
@@ -135,13 +176,9 @@ class EventDispatcher:
         'Schedule `callback` with the given args and kwargs in a util thread'
         self._utility_queue.put((callback, args, kwargs))
 
-    def run_in_thread(self, event_type, callback, *args, **kwargs):
-        event_thread = self._threads[event_type]
-        current_thread = threading.currentThread()
-        if current_thread is event_thread:
-            callback(*args, **kwargs)
-        else:
-            event_thread.queue.put((callback, args, kwargs))
+    def get_thread_context(self, name):
+        'Get the DispatcherThreadContext for the given thread name'
+        return self._thread_contexts[name]
 
     def _start_thread(self, name, *, callback_queue=None):
         'Start dispatcher thread by name'
@@ -152,6 +189,7 @@ class EventDispatcher:
                                                  logger=self.logger,
                                                  daemon=True,
                                                  callback_queue=callback_queue)
+        self._thread_contexts[name] = DispatcherThreadContext(self, name)
         self._threads[name].start()
 
 
--- a/ophyd/_dispatch.py
+++ b/ophyd/_dispatch.py
@@ -90,10 +90,8 @@ class DispatcherThreadContext:
 
         current_thread = threading.currentThread()
         if current_thread is self.event_thread:
-            print('calling', func, args, kwargs)
             func(*args, **kwargs)
         else:
-            print('scheduling', func, args, kwargs)
             self.event_thread.queue.put((func, args, kwargs))
 
     __call__ = run
--- a/ophyd/_dispatch.py
+++ b/ophyd/_dispatch.py
@@ -135,6 +135,14 @@ class EventDispatcher:
         'Schedule `callback` with the given args and kwargs in a util thread'
         self._utility_queue.put((callback, args, kwargs))
 
+    def run_in_thread(self, event_type, callback, *args, **kwargs):
+        event_thread = self._threads[event_type]
+        current_thread = threading.currentThread()
+        if current_thread is event_thread:
+            callback(*args, **kwargs)
+        else:
+            event_thread.queue.put((callback, args, kwargs))
+
     def _start_thread(self, name, *, callback_queue=None):
         'Start dispatcher thread by name'
         self._threads[name] = self._thread_class(name=name, dispatcher=self,
--- a/ophyd/_dummy_shim.py
+++ b/ophyd/_dummy_shim.py
@@ -1,8 +1,38 @@
 import threading
 
+
+class DummyDispatcherThreadContext:
+    dispatcher = None
+    event_type = None
+    event_thread = None
+
+    def run(self, *args, **kwargs):
+        ...
+
+    __call__ = run
+
+
+class DummyDispatcher:
+    context = None
+    logger = None
+    stop_event = None
+    timeout = 0.1
+    threads = {}
+
+    def stop(self):
+        ...
+
+    def schedule_utility_task(self, callback, *args, **kwargs):
+        ...
+
+    def get_thread_context(self, name):
+        return DummyDispatcherThreadContext()
+
+
 thread_class = threading.Thread
 pv_form = 'time'
 name = 'dummy'
+_dispatcher = DummyDispatcher()
 
 
 def setup(logger):
@@ -26,4 +56,4 @@ def release_pvs(*args, **kwargs):
 
 
 def get_dispatcher():
-    raise NotImplementedError
+    return _dispatcher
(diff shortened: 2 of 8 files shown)


tacaswell:
What is the bug that you are chasing here?

klauer:
@tacaswell I updated the original PR message. I think it's closer to being 'correct'.

ophyd/_dispatch.py
```diff
@@ -135,6 +174,10 @@ def schedule_utility_task(self, callback, *args, **kwargs):
         'Schedule `callback` with the given args and kwargs in a util thread'
         self._utility_queue.put((callback, args, kwargs))
 
+    def get_thread_context(self, name):
```
danielballan: It may not be worth paying for a function call here when exposing the dict would accomplish the same thing. Is the intent to protect `_thread_contexts` from some evil code that might mutate it, or to make space for the possibility that the process of getting a thread context might grow more complex in the future?

ophyd/_pyepics_shim.py
```diff
@@ -129,13 +134,16 @@ def clear_callbacks(self):
...
+    # Run _release_pvs in the 'monitor' thread, assuring that the CA context is correct
+    def _release_pvs():
+        for pv in pvs:
+            pv.clear_callbacks()
+            pv.clear_auto_monitor()
+        event.set()
+
+    event = threading.Event()
+    _dispatcher.get_thread_context('monitor').run(_release_pvs)
+    event.wait()
```
danielballan: Should this have a default timeout?

ophyd/ophydobj.py
```diff
@@ -310,6 +311,7 @@ def cb(*args, obj: OphydObject, sub_type: str, **kwargs) -> None:
 
         # wrapper for callback to snarf exceptions
         def wrap_cb(cb):
+            @functools.wraps(cb)
```
danielballan: :+1:

ophyd/_dispatch.py
klauer: It's not likely to get more complicated in the future. Assuming all users are benevolent, I think it comes down to a stylistic choice - which I don't feel strongly enough about to argue about. If you do, I can definitely nuke the function and expose it directly.

ophyd/_pyepics_shim.py
klauer: It could - but I'm not sure it's necessary here. The only possibility for this to timeout would be in the case of a massively backed-up event queue or deadlock.

In the former case, the caller would want to ensure that the PVs were fully released before proceeding. Without the wait, one could imagine a scenario where there is a race to clear the callbacks here or add callbacks in some other function.

ophyd/_dummy_shim.py
```diff
@@ -23,3 +23,7 @@ def get_pv(*args, **kwargs):
 
 def release_pvs(*args, **kwargs):
     raise NotImplementedError
+
+
+def get_dispatcher():
+    raise NotImplementedError
```
tacaswell: This will break the dummy backend as we unconditionally call this in `Signal.__init__`.

klauer: Looks like we need some dummy backend tests...

ophyd/signal.py
```diff
@@ -851,13 +862,19 @@ def fix_value(fixer_function, value):
             metadata['timestamp'] = time.time()
         return metadata
 
-    def _read_changed(self, value=None, **kwargs):
+    def _read_changed(self, value=None, from_monitor=True, **kwargs):
```
tacaswell: kwarg only?

ophyd/signal.py
```diff
@@ -1182,52 +1202,66 @@ def _pv_access_callback(self, read_access, write_access, pv):
...
+                severity=self._metadata['setpoint_severity'],
+                precision=self._metadata['setpoint_precision'],
+                lower_ctrl_limit=self._metadata['lower_ctrl_limit'],
+                upper_ctrl_limit=self._metadata['upper_ctrl_limit'],
+                units=self._metadata['units'],
+            )
         return metadata
 
-    def _write_changed(self, value=None, timestamp=None, **kwargs):
+    def _write_changed(self, value=None, timestamp=None, from_monitor=True,
```
tacaswell: the default of `from_monitor` is not given in `_metadata_changed` and True for `_read_changed`, should these be standardized?  (I would lean to no default in all 3 places).

ophyd/_dispatch.py
tacaswell: I'm :+1: on leaving this as-is.

ophyd/signal.py
klauer: Your comments made me rethink my previous attempt, nuking `write_changed`/`read_changed` `from_monitor=False`.

That also means there's more to review - sorry...

ophyd/signal.py
```diff
@@ -1182,58 +1215,67 @@ def _pv_access_callback(self, read_access, write_access, pv):
...
+        'CA monitor: callback indicating the setpoint PV value has changed'
         if timestamp is None:
             timestamp = time.time()
 
+        metadata = self._metadata_changed(self.setpoint_pvname, kwargs,
+                                          require_timestamp=True,
+                                          from_monitor=True,
+                                          update=False)
+
+        # Update values internally for consistency, but only run callbacks if this
```
tacaswell: This comment is now wrong.

ophyd/signal.py
```diff
@@ -1182,58 +1215,67 @@ def _pv_access_callback(self, read_access, write_access, pv):
...
     def _write_changed(self, value=None, timestamp=None, **kwargs):
-        '''A callback indicating that the write value has changed'''
+        'CA monitor: callback indicating the setpoint PV value has changed'
         if timestamp is None:
             timestamp = time.time()
 
+        metadata = self._metadata_changed(self.setpoint_pvname, kwargs,
+                                          require_timestamp=True,
+                                          from_monitor=True,
+                                          update=False)
```
tacaswell: Why not `update=True`?

klauer: Good call!

ophyd/_dispatch.py
danielballan: Was thinking more about (marginal) performance than style. Fine with leaving it as is.

ophyd/signal.py
```diff
@@ -1154,26 +1172,35 @@ def get_setpoint(self, *, as_string=None, connection_timeout=1.0,
...
         value = info.pop('value')
         if as_string:
             value = waveform_to_string(value)
 
-        # The following will update all metadata, run subscriptions, and
-        # also update self._readback such that this value can be accessed
-        # through EpicsSignal.value
-        self._write_changed(value=value, **info)
+        has_monitor = self._monitors[self.setpoint_pvname] is not None
+        if form != 'time' or not has_monitor:
```
tacaswell: Why does the monitor enter into this logic?  I understand updating this if it ctrl from, but why update it if the form in 'time' if there is not a monitor?

ophyd/signal.py
```diff
@@ -1182,58 +1209,65 @@ def _pv_access_callback(self, read_access, write_access, pv):
...
     def _write_changed(self, value=None, timestamp=None, **kwargs):
-        '''A callback indicating that the write value has changed'''
+        'CA monitor: callback indicating the setpoint PV value has changed'
         if timestamp is None:
             timestamp = time.time()
 
+        metadata = self._metadata_changed(self.setpoint_pvname, kwargs,
+                                          require_timestamp=True,
+                                          from_monitor=True,
+                                          update=False)
```
tacaswell: @klauer left a comment suggesting changing to `update=True` was a good idea and resolving the conversation, but this is still `update=False`.

ophyd/signal.py
klauer: The comment below touches on it:
```
# A different form (such as 'ctrl') holds additional data not available
# through the DBR_TIME channel access monitor
```
Part 1 of the `if form != 'time'`: It may include an updated 'units' or something else not in the dbr_time request. So, update the metadata because the next monitor event will not include it.

Part 2 of the `if not has_monitor`: means that no CA monitor is setup, so metadata will never be updated by a monitor callback. This is the only place it will be updated.

If you can find a more concise way to express that, I'd be happy to implement it.

ophyd/signal.py
klauer: Sloppy, @klauer, sloppy! Addressed in a follow-up commit.

ophyd/signal.py
tacaswell: I read that comment as only applying to part 1.  If there is a monitor, is there harm in updating the meta-data anyway?

Maybe just add the line "if there is not a monitor, this is the only chance to update the meta-data so do it here"?  It is possible that we lost a EventResponse so that this `get` has meta-data we have not heard about yet.

If the subscription is of a ctrl type (does that make sense?) are we painted in to a weird corner?

klauer: I get your hesitation, and I'm not entirely happy with this just yet - despite it being a few iterations from my first Signal refactor PR. The main issue is that these things remain racy, so updating the metadata at the wrong moment could wipe away what came from the monitor.

I don't think we have the right level of calls to the control layer to lock this appropriately without causing deadlock. That said, I'd like to see if we can address that raciness along with atomic updates of value + metadata (as `dict.update` is atomic in Python) in follow-up PRs.

PR #696, opened 2019-03-04
FIX: correct logger name

mrakitin:
I guess when the `log.py` was ported from bluesky via #664, the logger name wasn't changed accordingly. This PR corrects it.


mrakitin:
Thank you for reviewing/merging, @klauer!

PR #697, opened 2019-03-04
FIX: Setting Capture has no effect in Single Mode.

mrakitin:
Ophyd unconditionally tries to set the `capture` component of the TIFF plugin of the AD to `0` in https://github.com/NSLS-II/ophyd/blob/18de9724f0d9cf78f1a0e3abe3dd8da9408b6260/ophyd/areadetector/filestore_mixins.py#L427

However, this action is not supported when the TIFF plugin is in the "Single" file write mode, so that it results in the error `ERROR: capture not supported in Single mode`:
tiff-plugin

This PR attempts to workaround this situation.

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -424,7 +424,8 @@ class FileStorePluginBase(FileStoreBase):
         filename, read_path, write_path = self.make_filename()
 
         # Ensure we do not have an old file open.
-        set_and_wait(self.capture, 0)
+        if self.file_write_mode != 'Single':
+            set_and_wait(self.capture, 0)
         # These must be set before parent is staged (specifically
         # before capture mode is turned on. They will not be reset
         # on 'unstage' anyway.


klauer:
I don't have a better solution (or an answer to @danielballan's question, for that matter).

👍

prjemian:
@MarkRivers  Can you explain why Capture Mode is not applicable in Single Mode.  This issue is looking at the AD TIFF plugin

kgofron:
Is this inconsistency of TIFF settings, AD release dependent? I know that Mark has been modifying the stream/ capture functionality in recent releases.

For this sppecific camera:
base-7-0-1-1
ADProsilica: R2-5-6-g8a179ee
ADCore: R3-3-2
asyn: 4-33-15-g7550ea6

MarkRivers:
> I think this should be merged but (as discussed on private channels) it would be nice to get a clearer understanding of why Capture Mode is not applicable in Single Mode.

There are 3 modes, SingleMode, CaptureMode, and StreamMode.  The Capture control only applies in CaptureMode and StreamMode. 

Capture and Stream modes are intended for collecting multiple NDArrays into memory (Capture) or appending to single disk file (Stream).  Setting Capture to 0 in Single mode really should be a warning, not an error because it does not actually do anything.
 
> Is this inconsistency of TIFF settings, AD release dependent? 

This has nothing to do with TIFF, it is in the base NDPluginFile class.

> I know that Mark has been modifying the stream/ capture functionality in recent releases.

I have been discussing changing the stream/capture settings, but have not made any changes.  Here is the issue I created:
https://github.com/areaDetector/ADCore/issues/73

danielballan:
That all makes sense to me. Thank you, @MarkRivers!

danielballan:
I have changed the title of this PR as there is nothing TIFF-specific about either the problem or the fix here (which affects all file plugins).

mrakitin:
Thanks for the clarification, @MarkRivers!

PR #701, opened 2019-03-12
TST: Hide caproto imports.

danielballan:
This makes it possible for other libraries (e.g. suitcases) to reuse some
ophyd fixtures (e.g. ``hw``) without installing all of the optional caproto
test dependencies. This is no small thing: the set of of depedencies we avoid
having to manage includes curio, trio, special attr verison for trio, asv,
pytest-benchmark.

### DIFF
--- a/ophyd/tests/conftest.py
+++ b/ophyd/tests/conftest.py
@@ -9,7 +9,6 @@ from types import SimpleNamespace
 from ophyd import (set_cl, EpicsMotor, Signal, EpicsSignal, EpicsSignalRO,
                    Component as Cpt)
 from ophyd.utils.epics_pvs import (AlarmSeverity, AlarmStatus)
-from caproto.tests.conftest import run_example_ioc
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tacaswell:
Is there a way to use skiptest in fixtures?

awalter-bnl:
@tacaswell , while it is possible to use `pytest.skip()` in a fixture it is not so easy to pass in the command to skip or not. Below is how I achieved this in suitcase-utils by using an argument called 'ignore'. I am not immediately seeing how it applies in this case though.

https://github.com/NSLS-II/suitcase-utils/blob/136f3190b890f89ffd4f6f988d5d630e24b96dbd/suitcase/utils/tests/conftest.py#L70

ophyd/tests/conftest.py
```diff
@@ -109,6 +108,7 @@ def fake_motor_ioc(prefix, request):
                step_size=f'{prefix}step_size',
                )
 
+    from caproto.tests.conftest import run_example_ioc
```
tacaswell: we should use `pytest.importorskip` here so if you have tests that use this they fail gracefully rather than failing with ugly exceptions.

tacaswell:
Restarted the failed CI, looks like sim detector failed to start properly...

PR #702, opened 2019-03-12
ENH: Add register_instance.

danielballan:
At various times it has been proposed to "auto-register" ophyd objects
in order to maintain a global registery of devices. We have avoided
making OphydObj aware of all its own instances because this could have
weird implications and generally feels like an unusual design in Python.
But we could allow downstream code to sign up to be notified each time
an OphydObj is instantiated. Each subscriber will get one notification,
and they can keep a strong or weak reference if they like, as needed.

<hr />

This suggestion is fully due to @tacaswell and I may or may not have
implemented it exactly as he envisioned. I need to refer to it in
another discussion so I have sketched it here.

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -135,7 +135,7 @@ class OphydObject:
     # This is set to True when the first OphydObj is instiated. This may be of
     # interest to code that adds something to instantiation_callbacks, which
     # may want to know whether it has already "missed" any instances.
-    __instantiated = False
+    __any_instantiated = False
 
     def __init__(self, *, name=None, attr_name='', parent=None, labels=None,
                  kind=None):
@@ -183,7 +183,7 @@ class OphydObject:
         # Instantiate logger
         self.log = logging.getLogger(base_log + '.' + name)
 
-        if not self.__instantiated:
+        if not self.__any_instantiated:
             self.log.debug("This is the first instance of OphydObject. "
                            "name={self.name}, id={id(self)}")
             self.__mark_as_instantiated()
@@ -191,7 +191,7 @@ class OphydObject:
 
     @classmethod
     def __mark_as_instantiated(cls):
-        cls.__instantiated = True
+        cls.__any_instantiated = True
 
     @classmethod
     def add_instantiation_callback(cls, callback, fail_if_late=False):
@@ -207,7 +207,7 @@ class OphydObject:
             ``RuntimeError`` if it has, as a way of verify that no instances will
             be "missed" by this registry. False by default.
         """
-        if fail_if_late and OphydObject.__instantiated:
+        if fail_if_late and OphydObject.__any_instantiated:
             raise RuntimeError(
                 "OphydObject has already been instantiated at least once, and "
                 "this callback will not be notified of those instances that "
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -1,5 +1,6 @@
 import functools
 from itertools import count
+import weakref
 
 import time
 import logging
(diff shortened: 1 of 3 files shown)


ophyd/ophydobj.py
```diff
@@ -121,6 +124,14 @@ def __init__(self, *, name=None, attr_name='', parent=None, labels=None,
             name = self.name
         # Instantiate logger
         self.log = logging.getLogger(base_log + '.' + name)
+        self.__register_instance(self)
+
+    def __register_instance(self):
```
tacaswell: make this a `@classmethod` to make sure we don't get different callbacks on different objects?

ophyd/ophydobj.py
```diff
@@ -121,6 +124,14 @@ def __init__(self, *, name=None, attr_name='', parent=None, labels=None,
             name = self.name
         # Instantiate logger
         self.log = logging.getLogger(base_log + '.' + name)
+        self.__register_instance(self)
```
tacaswell: this should go in `__init_subclass__` so sub-classes can not escape :smiling_imp:

tacaswell:
This is both simpler and more flexible than what I was thinking, however it leaves open an issue where we have

```python
OphydObj.new_instance_callbacks.append(some_callback)
obj = Obj1()
OphydObj.new_instance_callbacks.append(some_other_callback)
obj2 = Obj2()
```

which results in the two callbacks seeing different sets of objects which could lead to weird consistency issues.  On the other hand this could be a feature (for example to set up same 'base' devices that register with a "all-the-things" registry, and then install a "for-the-gui" registry and instantiate the rest of your devices).  On the other-other hand, relying on instantiation order is super brittle and a sign that you should probably be explicitly building that registry.  This is the 'simpler' part as the locking logic is non-trivial.

The more flexible part is I was thinking this would be a single callback not a list, list is definitely better.

ophyd/ophydobj.py
danielballan: Hm. `__init_subclass__` doesn't have access to the instance, does it? Are you envisioning redefining `__init__` in `__init_subclass__`?

tacaswell: ah, nvm

klauer:
So the argument against keeping track of ophyd objects in a standard way: "it's weird... so let the user do it themselves through callbacks"? **That** seems weird to me.

In any case:
I think rather than the callback mechanism which pushes the burden onto the subscriber to keep track of lists of objects in a safe way (e.g., with weak references), it would be advisable to make some helper that correctly does the bookkeeping for you.

While there could be 2 flavors of weak reference vs strong reference helpers (as mentioned in your PR description), in my opinion the latter should not even exist as it could very easily lead to unintended consequences.

tacaswell:
> So the argument against keeping track of ophyd objects in a standard way: "it's weird... so let the user do it themselves through callbacks"? **That** seems weird to me.

There are a bunch of decisions here that I don't see clear answer to.  What are the keys we use (likely obj.name, but that can change after instatitation, maybe the prefix (but that has even more collisions) )? Do we provide filtering by class (`reg[type(obj)][obj.name] == obj`) (this has problems as some classes are dynamically generated (DCC and friends))?  What do we do if there are collisions? Do we include all of the instances or just 'root' instances that have no parent?

When we talked about this sort of thing initially we fell back to "if you want a registry make it your self" which requires more discipline than we have actually observed in beamline profiles so this was a hook to guarantee you get a bite at seeing each obphyd object at least once.

klauer:
@tacaswell those are all great questions - which should be solved by the library and not by the end-user.

danielballan:
The "end user" I was imagining for this feature is "downstream library author" not end user. Should we brainstorm how we imagine this might be used?

klauer:
@danielballan - that sounds reasonable. What were the use cases that drove this development?
Largely, I'd assume, keeping track of all objects of a specific type?

danielballan:
Belatedly documenting my recollection of our consensus from the last call:

* Provide several potentially-useful functions that could be useful with `register_instance` in some situation.
* Do not register anything by default.
* Do not add "error if `register_instance` called after the first instance is made."

To that third point, we _could_ add a flag that flips from `False` to `True` when the first instance is made, which the caller of `register_instance` could _check_ if they want to be sure that they know about all instances. Any feelings about that?

ZLLentz:
A flag for the first `register_instance` call seems appropriate to me, some may find it useful. Can we also store information about the first object registered? This would help us identify where the mysterious pre-register ophyd objects are coming from.

danielballan:
That seems useful but I'm unsure how to implement it. Should we have `first_instance = weak.ref(...)`?

ZLLentz:
I'm not sure what would be an optimal implementation. Even something as simple as `instance.log.debug('I was the first ophyd object created')` would be very useful as a debug tool for the registry timing.

danielballan:
Great, that's a good idea. Issuing a one-off DEBUG log message is safer than wading into reference-holding.

danielballan:
This is ready for another round of discussion. Needs tests.

danielballan:
Example usage:

```py
>>> import ophyd
>>> import ophyd.ophydobj
>>> s = ophyd.ophydobj.register_instances_in_weakset()
>>> signal = ophyd.Signal(name='signal')
>>> list(s)
[Signal(name='signal', value=0.0, timestamp=1553886512.2564445)]
```

danielballan:
Add some tests and fixes.

ophyd/ophydobj.py
```diff
@@ -74,7 +126,15 @@ class OphydObject:
...
 
+    # Any callables appended to this mutable class variable will be notified
+    # one time when a new instance of OphydObj is instantiated. See
+    # OphydObject.add_instantiation_callback().
+    __instantiation_callbacks = []
     _default_sub = None
+    # This is set to True when the first OphydObj is instiated. This may be of
+    # interest to code that adds something to instantiation_callbacks, which
+    # may want to know whether it has already "missed" any instances.
+    __instantiated = False
```
tacaswell: maybe name this `__any_instantiated`?  `__instatiated` feels like it is a `per-instance` name...

ophyd/ophydobj.py
```diff
@@ -49,6 +50,57 @@ class UnknownSubscription(KeyError):
     ...
 
 
+def register_instances_keyed_on_name(fail_if_late=False):
+    """Register OphydObj instances in a WeakValueDictionary keyed on name.
+
+    Be advised that ophyd does not require 'name' to be unique so this should
```
tacaswell:
```suggestion
    Be advised that ophyd does not require 'name' to be unique and is
    configurable by the user at run-time so this should
```

danielballan:
Renamed `__instantiated` -> `__any_instantiated` per @tacaswell request.

PR #703, opened 2019-03-14
FIX: correct path semantics behavior

EliotGann:
This is found with @mrakitin at NSLS-II RSOXS beamline when the collection is done on the Windows machine, and the detector is on a Linux one.

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -24,7 +24,7 @@ import os
 import logging
 import warnings
 import uuid
-from pathlib import PurePath
+from pathlib import Path, PurePath, PurePosixPath
 
 from datetime import datetime
 from collections import defaultdict
@@ -34,6 +34,8 @@ from ..device import GenerateDatumInterface, BlueskyInterface, Staged
 from ..utils import set_and_wait
 
 logger = logging.getLogger(__name__)
+
+
 
 
 def new_uid():
@@ -46,7 +48,7 @@ def new_short_uid():
     return '-'.join(new_uid().split('-')[:-1])
 
 
-def _ensure_trailing_slash(path):
+def _ensure_trailing_slash(path, path_semantics=None):
     """
     'a/b/c' -> 'a/b/c/'
 
@@ -54,7 +56,10 @@ def _ensure_trailing_slash(path):
     setpoint filepath to match the readback filepath, we need to add the
     trailing slash ourselves.
     """
-    return os.path.join(path, '')
+    if path_semantics is 'posix':
+        return f'{PurePosixPath(path)}/'
+    else:
+        return f'{PurePath(path)}{os.path.sep}'
 
 
 def resource_factory(spec, root, resource_path, resource_kwargs,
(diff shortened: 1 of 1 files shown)


mrakitin:
Before making it ready for review, it will be tested with the following scenarios:
- [ ] Linux collection machine + Linux IOC machine (mostly any detector on the NSLS-II floor)
- [ ] Linux collection machine + Windows IOC machine (e.g. PE detector, e.g. at QAS or XPD)
- [x] Windows collection machine + Linux IOC machine (the GreatEyes detector at NSLS-II RSOXS beamline, see #704) (already tested)

klauer:
Looks like this could be a good step in the right direction! 👍

ophyd/areadetector/filestore_mixins.py
```diff
@@ -46,15 +48,18 @@ def new_short_uid():
...
 
     EPICS adds the trailing slash itself if we do not, so in order for the
     setpoint filepath to match the readback filepath, we need to add the
     trailing slash ourselves.
     """
-    return os.path.join(path, '')
+    if path_semantics is 'posix':
+        return f'{PurePosixPath(path)}/'
+    else:
+        return f'{PurePath(path)}{os.path.sep}'
```
danielballan: Consider the scenario where ophyd is running on a Linux workstation but the area detector driver is running on Windows, as is the case at some beamlines. In that situation, `os.path.sep` will return `/` but the driver needs to be set to `\`. I suggest something along the lines of:

```py
if path_semantics == 'posix':
    return f'{PurePosixPath(path)}/'
elif path_semantics == 'windows':
    return  f'{PurePath(path)}\'
elif path_semantics is None:
    # We are forced to guess which path semantics to use.
    # Guess that the AD driver is running on the same OS as this client.
    return f'{PurePath(path)}{os.path.sep}'
else:
    # This should never happen, but just for the sake of future-proofing....
    raise ValueError(f"Cannot handle path_semantics={path_semantics}")
```

gwbischof:
You may want to see @danielballan fix for this also
https://github.com/danielballan/ophyd/tree/use-valid-path-semantics

ophyd/areadetector/filestore_mixins.py
mrakitin: @danielballan, I like the more explicit conditions.
For the Windows part it should be either `f'{PurePath(path)}\\\\'` or `fr'{PurePath(path)}\\'` to have `\\` returned at the end. Also, I think it should better be `PureWindowsPath`.

mrakitin: I played with it a bit, and `f'{PurePath(path)}\\'` should be OK (we need two backslashes to escape a special backslash symbol).

mrakitin:
I added the suggested improvement and expanded it to be used in the `write_path_template` method via d916b18.

danielballan:
This languished for far too long! Sorry, @mrakitin, @EliotGann, and @RussBerg. This will go out in ophyd v1.4.1 on Monday.

mrakitin:
Thank you, @danielballan.

klauer:
Just one year? Seems about right...

mrakitin:
As a Russian proverb says, "one has to wait for three years for what is promised" 😄

PR #704, opened 2019-03-14
Add GreatEyes detector

EliotGann:
Implemented with @mrakitin at NSLS-II RSOXS beamline.

### DIFF
--- a/ophyd/areadetector/detectors.py
+++ b/ophyd/areadetector/detectors.py
@@ -127,7 +127,7 @@ class FirewireWinDetector(DetectorBase):
 
 
 class GreatEyesDetector(DetectorBase):
-    _html_docs = ['greateyesDoc.html']  # the documentation is not public
+    _html_docs = []  # the documentation is not public
     cam = C(cam.GreatEyesDetectorCam, 'cam1:')
 
 
--- a/ophyd/areadetector/detectors.py
+++ b/ophyd/areadetector/detectors.py
@@ -19,6 +19,7 @@ __all__ = ['DetectorBase',
            'DexelaDetector',
            'FirewireLinDetector',
            'FirewireWinDetector',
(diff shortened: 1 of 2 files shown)


ophyd/areadetector/detectors.py
```diff
@@ -125,6 +126,11 @@ class FirewireWinDetector(DetectorBase):
     cam = C(cam.FirewireWinDetectorCam, 'cam1:')
 
 
+class GreatEyesDetector(DetectorBase):
+    _html_docs = ['greateyesDoc.html']  # the documentation is not public
```
klauer:
```suggestion
    _html_docs = []  # the documentation is not public
```

The documentation is used internally to get docstrings for AreaDetector components, so clearing this list would be appropriate.

Everything else looks fine. Thanks!

mrakitin: Thanks @klauer! I tried to apply the suggestion, but don't seem to have permissions to push to @EliotGann's PR branch.

@EliotGann, can you please accept Ken's suggestions on this page? That will go as a separate commit into your branch.

mrakitin:
For those who are interested in the implementation at the beamline, here is the link: https://github.com/NSLS-II-SST/profile_collection/commit/1883eb5496662adf65fa5373cc21af6edfd2cc09.

ophyd/areadetector/detectors.py
EliotGann: I tried to accept the suggestion ... not sure if I did it correctly

mrakitin: Yeah, it was correctly done, we can see 583bb9b1181905fe04c088d8db37272670fb402d commit now. Thanks!

mrakitin:
I assume it's good to go given that we have 2 approvals.

PR #705, opened 2019-03-22
FIX: Use a valid path semantics choice.

gwbischof:
The event-model expects path semantics to be one of
``{'posix', 'windows'}`` but ``os.name`` returns ``'nt'`` on Windows.


stuartcampbell:
Looks ok to me

Issue #708, opened 2019-04-13
ophyd master pyepics control layer `get_pv` PV reuse needs work


klauer:
The path forward will likely be one of the following, in order of preference:
* Fixing pyepics `epics.PV` (another PR) to more easily handle running callbacks on already-created instances
* Making `pyepics_shim.PV` rely only on pyepics `epics.ca` (i.e., not the `epics.PV` class)
* Vendoring `epics.PV` and making it ophyd-specific

Problem area:
https://github.com/NSLS-II/ophyd/blob/1ca2a260e990fe406d718f68f3738cf9d1085ce3/ophyd/_pyepics_shim.py#L173-L191

klauer:
Closed by #720

PR #710, opened 2019-04-13
ENH: directly run the connection and access callbacks if PV exists

tacaswell:
Attempts to address #708


klauer:
PR attempts to address #708

tacaswell:
Closing in favor of #720 

(@klauer in general assume you can push to my branches and force-push with a comment!)

PR #721, opened 2019-05-13
CI: Publish to bluesky.github.io. Always build.

danielballan:


.travis.yml
```diff
@@ -117,18 +119,17 @@ script:
   - coverage report
   # Make sure only the primary build uploads to codecov
   - |
-    if [ $BUILD_DOCS ]; then
+    if [ $PUBLISH_DOCS ]; then
```
mrakitin: Should not it be something like
```suggestion
    if [ ! -z "$PUBLISH_DOCS" ]; then
```
?

danielballan: As written the docs are published if `PUBLISH_DOCS` is _set_, so we turn it off by simply not setting that variable. But I agree that this is clearer.

.travis.yml
```diff
@@ -117,18 +119,17 @@ script:
...
+  - pip install -r docs-requirements.txt
+  - make -C doc html
   - |
-    if [ $BUILD_DOCS ]; then
-      pip install -r docs-requirements.txt
-      pushd doc
-      make html
-      popd
-      # Publish docs.
+    if [ $PUBLISH_DOCS ]; then
```
mrakitin:
```suggestion
    if [ ! -z "$PUBLISH_DOCS" ]; then
```

.travis.yml
mrakitin: Indeed, they work identically. Tested it with this script and changing the `VAR`:
```bash
#!/bin/bash

VAR=test

echo "Old approach"
if [ $VAR ]; then
    echo "VAR set to $VAR"
else
    echo "VAR is empty"
fi

echo "New approach"
if [ ! -z "$VAR" ]; then
    echo "VAR set to $VAR"
else
    echo "VAR is empty"
fi
```
Outputs:
```bash
$ bash test.sh
Old approach
VAR is empty
New approach
VAR is empty
# Update the 'VAR' to be 'test'
$ bash test.sh
Old approach
VAR set to test
New approach
VAR set to test
```

PR #723, opened 2019-05-16
CI: Silence codecov comments.

danielballan:
This adds a configuration file that we have had in bluesky and other
repos for awhile now. It silences codecov giant comments, while still
leaving codecov's X's or checkmarks in the block "checks" above the
Merge button.


mrakitin:
Thanks for the update, @danielballan!

PR #725, opened 2019-05-21
ENH: reconfigure hints for sim motors

mrakitin:
This is to make `no_hints` tests passing in Bluesky.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -259,7 +259,7 @@ class SetpointSignal(Signal):
         return self.parent.sim_state['setpoint_ts']
 
 
-class SynAxisNoHints(Device):
+class SynAxis(Device):
     """
     A synthetic settable Device mimic any 1D Axis (position, temperature).
 
@@ -375,8 +375,17 @@ class SynAxisNoHints(Device):
         return self.readback.get()
 
 
-class SynAxis(SynAxisNoHints):
-    readback = Component(ReadbackSignal, value=None, kind=Kind.hinted)
+class SynAxisEmptyHints(SynAxis):
+    @property
+    def hints(self):
+        return {}
+
+
+class SynAxisNoHints(SynAxis):
+    readback = Component(ReadbackSignal, value=None, kind='omitted')
+    @property
+    def hints(self):
+        raise AttributeError
 
 
 class SynGauss(SynSignal):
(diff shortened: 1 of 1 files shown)


ophyd/sim.py
```diff
@@ -375,8 +375,17 @@ def position(self):
         return self.readback.get()
 
 
-class SynAxis(SynAxisNoHints):
-    readback = Component(ReadbackSignal, value=None, kind=Kind.hinted)
+class SynAxisEmptyHints(SynAxis):
```
teddyrendahl: It isn't enough to just lower all the Components to `omitted` in the test?

`motor.readback.kind = 'omitted'`

Seems strange to add a class to the library just for this test?

mrakitin: Yes, it sounds reasonable. I just wanted to follow the existing approach when the `SynAxisNoHints` class already existed (however was wrongly defined.) I tried `motor.readback.kind = 'omitted'`, however it fails on Bluesky's BestEffortCallback, as in this case `motor.hints` returns `{'field': []}`, not `{}`. We will probably have to change how it's handled by Bluesky. I'll post the code and traceback into the linked issue there.

teddyrendahl:
@mrakitin If this is slowing your development please move ahead. I am not intending to be a stick in the mud. This will not break anybody's code.

I do however see a slightly worrying pattern that `ophyd.sim` is getting picked at as needed for `bluesky` tests (see https://github.com/bluesky/ophyd/pull/682/files as well) and wanted to just make a public note.

ophyd/sim.py
teddyrendahl: Isn't that the more realistic behavior? It appears to me the only way you'll get a completely empty dictionary is a malformed subclass with `hints` defined.

https://github.com/bluesky/ophyd/blob/be880e1bcc87c05668bf300845e32416537e647b/ophyd/device.py#L1261

danielballan:
> I do however see a slightly worrying pattern that ophyd.sim is getting picked at as needed for bluesky tests 

That's a fair concern. I propose that we keep @mrakitin's fix to `SynAxisNoHints` here, since the class exists, but that we move `SynAxisEmptyHints` into the bluesky test file where it is used.

As to #682, I have advocated for defining those classes in suitcase-utils (where they are used by test fixtures) and I think it's worth revisiting that suggestion now. I'll move further comment on that point to that PR.

Going forward, I think ophyd should provide realistic Device/Signal subclasses suitable for simulation and testing, but that if bluesky wants to check that it works against objects that provide some subset of the "bluesky interface" _without_ `ophyd.Device` or some broken variant of the bluesky interface, those classes ought be written in the bluesky test suite, not in `ophyd.sim`. Agreed, @teddyrendahl? Thoughts, anyone else?

prjemian:
Right. Now you have described the real world situation you wish to test.
You can test by example(s) but not by exhaustion.

ophyd/sim.py
```diff
@@ -375,8 +375,17 @@ def position(self):
...
+    @property
+    def hints(self):
+        return {}
+
+
+class SynAxisNoHints(SynAxis):
+    readback = Component(ReadbackSignal, value=None, kind='omitted')
+    @property
+    def hints(self):
+        raise AttributeError
```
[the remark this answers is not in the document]
mrakitin: It was seen in the code below for a different class, and simultaneously found at stackoverflow 🙂

PR #727, opened 2019-05-22
fixes #726

### LINKED ISSUE #726: EpicsMotor won't set_current_position() outside of current limits
Reported by @cpchuang:

Another interesting finding of “EpicsMotor” device,

    samx.set_current_position([number])

won’t let me set position to a value outside current low/high-limit. Not sure if this is a bug or a feature. In my opinion, this action shouldn’t check the high/low limit since this is to re-define “value” of current physical position. In fact, the limit should change accordingly, which it did if the value is within this limit.

 
```
In [167]: samx.high_limit                                                                                                                                

Out[167]: 9.0

 

In [168]: samx.low_limit                                                                                                                                

Out[168]: -11.0

 

In [169]: samx.position                                                                                                                                 

Out[169]: -10.0

 

In [170]: samx.set_current_position(-15)

LimitError: Value -15 outside of range: [-11.0, 9.0]
```

prjemian:
Skip limits check in EpicsMotor.set_current_position()

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -179,7 +179,7 @@ class EpicsMotor(Device, PositionerBase):
 
         '''
         self.set_use_switch.put(1, wait=True)
-        self.user_setpoint.put(pos, wait=True)
+        self.user_setpoint.put(pos, wait=True, force=False)
         self.set_use_switch.put(0, wait=True)
 
     @raise_if_disconnected


tacaswell:
Can you adjust the test in ophyd/tests/test_epicsmotor.py to exercise this?

tacaswell:
Good catch on finding / fixing this!

tacaswell:
Good catch on finding / fixing this!

danielballan:
The failure is just on the Python nightly.

PR #729, opened 2019-05-24
Support for AD file writer CreateDirectory PV

### LINKED ISSUE #728: TimeoutError when setting areaDetector file path
When using `yield from bps.mv(areadetector.tiff1.file_path, "/tmp/one/two/three/four/five")`, a TimeoutError is received (eventually).  Note that `abs_set` returns with no error.  Same behavior is observed for the HDF5 plugin.  Assume it's a common problem for all file writer plugins.

The problem is that `$(P):TIFF1:CreateDirectory` is set to 0 (default setting) so the AD TIFF plugin does not create the additional directory paths.

First off, this should not cause a TimeoutError since AD has a PV that describes if setting the file path succeeded (`$(P):TIFF1:FilePathExists_RBV`).  This may need custom handling in ophyd `.set()` or `.put()` (which one?).

Second, the user should take control of this PV (needs an interface in ophyd) and set it to `-len(new_file_path.split("/"))` (negative number and then number of subdirectories to be created).

prjemian:
AreaDetector file writer plugins (TIFF, HDF5, ...) are intended to write the image buffer to a directory and file.  The plugin can create the directory path, given certain parameters.  One parameter that is not yet interfaced in ophyd is the `CreateDirectory` PV which can, if set to a negative value, create up to that many subdirectory levels.  It's default setting is zero (no directories created).  This PV will fix #728.

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -785,6 +785,7 @@ class FilePlugin(PluginBase, GenerateDatumInterface, version=(1, 9, 1), version_
     auto_increment = Cpt(SignalWithRBV, 'AutoIncrement', kind='config')
     auto_save = Cpt(SignalWithRBV, 'AutoSave', kind='config')
     capture = Cpt(SignalWithRBV, 'Capture')
+    create_directory_depth = Cpt(SignalWithRBV, suffix="CreateDirectory")
     delete_driver_file = Cpt(SignalWithRBV, 'DeleteDriverFile', kind='config')
     file_format = Cpt(SignalWithRBV, 'FileFormat', kind='config')
     file_name = Cpt(SignalWithRBV, 'FileName', string=True, kind='config')


prjemian:
Is there a way to set a default value?  We have found (at APS) that `-5` is a reasonable default for us.

ophyd/areadetector/plugins.py
```diff
@@ -785,6 +785,7 @@ class FilePlugin(PluginBase, GenerateDatumInterface, version=(1, 9, 1), version_
     auto_increment = Cpt(SignalWithRBV, 'AutoIncrement', kind='config')
     auto_save = Cpt(SignalWithRBV, 'AutoSave', kind='config')
     capture = Cpt(SignalWithRBV, 'Capture')
+    create_directory_depth = Cpt(SignalWithRBV, suffix="CreateDirectory")
```
tacaswell: We already have this pv, but it is in the class for 2-2+ ADs (see https://github.com/bluesky/ophyd/blob/949efc0fa560a4affc26482276177f26daf7d4e8/ophyd/areadetector/plugins.py#L1042-L1047 )

tacaswell:
The `stage_sigs` are the place to put the configuration you want the detector to be in during data collection.

ophyd/areadetector/plugins.py
prjemian: Here's my setup.  Why didn't I get it then?

```In [1]: 
   ...: class MySingleTriggerSimDetector(SingleTrigger, SimDetector): 
   ...:        
   ...:     image = Component(ImagePlugin, suffix="image1:")
   ...:     hdf1 = Component(
   ...:         MyHDF5Plugin,
   ...:         suffix='HDF1:', 
   ...:         root='/',                               # for databroker
   ...:         
   ...:         # note: path MUST, must, MUST have trailing "/"!!!
   ...:         #  ...and... start with the same path defined in root (above)
   ...:         write_path_template="/tmp/simdet/%Y/%m/%d/",    # for EPICS AD
   ...:     )
   ...:     tiff1 = Component(
   ...:         MyTIFFPlugin,
   ...:         suffix='TIFF1:', 
   ...:         write_path_template="/tmp/simdet/%Y/%m/%d/",    # for EPICS AD
   ...:     )
   ...: 
   ...: 
   ...: _ad_prefix = "vm7SIM1:"        # IOC prefix

In [2]: adsimdet = MySingleTriggerSimDetector(_ad_prefix, name='adsimdet')

In [3]: hasattr(adsimdet.tiff1, "create_directory")
Out[3]: False

In [5]: TIFFPlugin??
Init signature: TIFFPlugin(*args, **kwargs)
Docstring:      AreaDetector plugin base class
Source:        
class TIFFPlugin(FilePlugin):
    _default_suffix = 'TIFF1:'
    _suffix_re = 'TIFF\d:'
    _html_docs = ['NDFileTIFF.html']
    _plugin_type = 'NDFileTIFF'
File:           ~/Apps/anaconda/envs/bluesky/lib/python3.6/site-packages/ophyd/areadetector/plugins.py
Type:           ComponentMeta
```

Need a custom TIFFPlugin that is otherwise identical except subclass of `FilePlugin_V22`?

prjemian:
So this must come *before* the stage sig for the file path.  Got it.

ophyd/areadetector/plugins.py
prjemian: Aha.  That feature is in a PR, not a release.  That's why I can't find it locally.

tacaswell:
Sorry, on a plane that is starting to land so no time, have a look at the PR where @klauer added those classes for how to use them.

The stage_sigs are an ordered dict exactly to support this sort of sequenced staging.

prjemian:
Thanks.  Happy landing.  Great support, even at 30k'!

klauer:
Yes, this is addressed in the other PR as @tacaswell mentions. We need to keep the PVs separated by version in their respective classes, so I'm afraid we can't merge this.

PR #730, opened 2019-05-27
BLD: Add python_requires.

danielballan:
This is now part of our standard configuration for new projects, but it was
never added to ophyd.


jklynch:
Reviewed and tested.

PR #731, opened 2019-06-04
Add array trimming before reshaping

ronpandolfi:
Following discussion of https://github.com/areaDetector/ADFastCCD/issues/13 with @mrakitin and @stuwilkins, it was decided that it is possible for EPICS to transmit more elements than the product of its shape, as with the FastCCD and "overscan rows".

To accommodate this, the image array must be trimmed before reshaping.


stuwilkins:
@ronpandolfi I wonder if this will blow up if the array is < the product of the array dimensions? I just have a question if we should trap that? This is very possible if the IOC is not configured correctly.

ophyd/areadetector/base.py
```diff
@@ -100,13 +100,11 @@ def forward(self, value):
...
     def inverse(self, value):
         """Shape the flat array to send as a result of ``.get``"""
-        print(self.derived_shape, self._shape, self.derived_ndims)
         array_shape = self.derived_shape[:self.derived_ndims]
         if not any(array_shape):
             raise RuntimeError(f"Invalid array size {self.derived_shape}")
 
-        array = self._derived_from.get()
-        return np.array(array).reshape(array_shape)
+        return np.array(value[:np.prod(array_shape)]).reshape(array_shape)
```
mrakitin: Shouldn't it be `np.array(array[...` using the `array` from L108? Not sure there the input `value` was used before.

danielballan: What type is `value` here? If it already an array (at least sometimes, in some circumstances) using `np.asarray` in place of `np.array` will save us from making a copy.

ronpandolfi:
Qingteng Zhange has also encountered this issue with a Pilatus detector:
Screen Shot 2019-06-19 at 4 29 45 PM

danielballan:
@ronpandolfi I'm up for tagging a patch release to get this out. Can you look at @mrakitin's and my comments above when you get a chance?

ophyd/areadetector/base.py
ronpandolfi: > Shouldn't it be `np.array(array[...` using the `array` from L108? Not sure there the input `value` was used before.

The array value is actually passed into this method as the `value` arg; there's no need to get it again, so I've removed that line.

ronpandolfi:
> @ronpandolfi I wonder if this will blow up if the array is < the product of the array dimensions? I just have a question if we should trap that? This is very possible if the IOC is not configured correctly.

@stuartcampbell I think that case should raise, indicating that misconfiguration. The exception raised currently is `ValueError: cannot reshape array of size Z into shape (X,Y)`. I'll add something suggesting to check the IOC.

ronpandolfi:
@danielballan Above comments have been addressed.

tacaswell:
Restarted the failed test.

@mrakitin We don't have a good enough test rig for AD stuff to be worth reproducing this issue.

PR #732, opened 2019-06-05
HACK: determine ophyd object name from code frame

klauer:
5 minute fun test, inspired by https://github.com/bluesky/bluesky.github.io/pull/6

Not a serious PR


danielballan:
Ha, wild.

PR #739, opened 2019-06-19
no pickles for numpy saving/loading (numpy cve handling)

CJ-Wright:


danielballan:
Per https://www.numpy.org/devdocs/reference/generated/numpy.load.html, this was made `False` by default in numpy 1.16.3, so this change will only effect earlier versions of numpy where the default was `True`.

CJ-Wright:
Thank you!

PR #740, opened 2019-06-19
DOC: Mark 1.4.0 as pre-release.

danielballan:
If we plan to tag an rc1 today, I don't think the date should be the date of
the rc1---it should be the date of the eventual release. For now, should we
mark it as "pre-release"?


klauer:
Yes.

PR #746, opened 2019-07-05
Fix 'shape' reported by describe().

danielballan:
We noticed in https://github.com/bluesky/bluesky-browser/pull/36
that area detectors have been reporting the wrong shape in
``describe()``. For a given image ``arr`` referenced by ``read()[data_key]``
we expect ``describe()[data_key]['shape']`` == ``arr.shape``.

Currently ``describe()[data_key]['shape']`` is ``(x, y, 0)``. To match
the ``arr.shape``, it should be ``(num_images, y, x)``. This PR changes
the shape to make it correct.

Downstream tooling will need to cope with both, unless/until we sort out
how to do data migrations. The addition of version metadata to the
RunStart document in bluesky 1.6.0 may help with this.

### DIFF
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -73,9 +73,9 @@ class CamBase(ADBase):
 
     array_callbacks = ADCpt(SignalWithRBV, 'ArrayCallbacks')
     array_size = DDC(ad_group(EpicsSignalRO,
-                              (('array_size_x', 'ArraySizeX_RBV'),
+                              (('array_size_z', 'ArraySizeZ_RBV'),
                                ('array_size_y', 'ArraySizeY_RBV'),
-                               ('array_size_z', 'ArraySizeZ_RBV'))),
+                               ('array_size_x', 'ArraySizeX_RBV'))),
                      doc='Size of the array in the XYZ dimensions')
 
     array_size_bytes = ADCpt(EpicsSignalRO, 'ArraySize_RBV')
--- a/ophyd/areadetector/detectors.py
+++ b/ophyd/areadetector/detectors.py
@@ -73,7 +73,10 @@ class DetectorBase(ADBase):
 
     def make_data_key(self):
         source = 'PV:{}'.format(self.prefix)
-        shape = tuple(self.cam.array_size.get())
+        # This shape is expected to match arr.shape for the array.
+        shape = (self.cam.num_images.get(),
+                 self.cam.array_size.array_size_y.get(),
+                 self.cam.array_size.array_size_x.get())
         return dict(shape=shape, source=source, dtype='array',
                     external='FILESTORE:')
 
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -237,9 +237,9 @@ class PluginBase(ADBase, version=(1, 9, 1), version_type='ADCore'):
     height = Cpt(EpicsSignalRO, 'ArraySize1_RBV')
     depth = Cpt(EpicsSignalRO, 'ArraySize2_RBV')
     array_size = DDC_EpicsSignalRO(
+        ('depth', 'ArraySize2_RBV'),
         ('height', 'ArraySize1_RBV'),
         ('width', 'ArraySize0_RBV'),
-        ('depth', 'ArraySize2_RBV'),
         doc='The array size'
     )
 
@@ -314,9 +314,9 @@ class ImagePlugin(PluginBase, version=(1, 9, 1), version_type='ADCore'):
 
     array_data = Cpt(EpicsSignal, 'ArrayData')
     shaped_image = Cpt(NDDerivedSignal, derived_from='array_data',
-                       shape=('array_size.height',
-                              'array_size.width',
-                              'array_size.depth'),
+                       shape=('array_size.depth',
+                              'array_size.height',
+                              'array_size.width'),
                        num_dimensions='ndimensions',
                        kind='omitted')
 
@@ -326,8 +326,8 @@ class ImagePlugin(PluginBase, version=(1, 9, 1), version_type='ADCore'):
         if array_size == (0, 0, 0):
             raise RuntimeError('Invalid image; ensure array_callbacks are on')
 
-        if array_size[-1] == 0:
-            array_size = array_size[:-1]
+        if array_size[0] == 0:
+            array_size = array_size[1:]
 
         pixel_count = self.array_pixels
         image = self.array_data.get(count=pixel_count)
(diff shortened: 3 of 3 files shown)


prjemian:
100% agree with this change.  It is a minor change that **breaks** the API.  Should *not* rely on passive check that metadata might report the ophyd version.  Need to do something that identifies this change as the point at which the API changed, such as adding an attribute.  Any data without that attribute would have a postive, definite means of announcing the shape info is reversed.

danielballan:
@mrakitin Yeah, the issue was `tuple((y, x, z))` should just be `(y, x, z)`. :-D Already force-pushed a fix.

@prjemian I agree that a more targeted, guaranteed-to-be-present signal would be good. Let's think about where we could put it. Consumers of the document stream can't see the ophyd objects and thus can't check for a new attribute. We could add a key to the data_keys:

```python
>>> det.some_signal.describe()
{'...': {'shape': ..., 'dtype': '...', 'source': 'PV:...', 'control_system': {'versions': {'ophyd': '1.4.0'}}}
```

One could imagine reporting other things in there too, like pyepics version, whether a given signal is using pyepics or caproto or some yet-to-be-written control layer integration like Tango. This would make the payload from `describe()` significantly heavier though.

prjemian:
@danielballan : I see your point.  Stepping back a bit, is the document the place where data readers will retrieve this piece of information about the image?  Or is it the content produced by the EPICS ADCore plugin that provides the shape that is actually used?  If the content comes from the document, then we need an attribute (which ends up in the appropriate document) that describes the shape array is forward or reverse.  Whichever the choice, it will be opposite of what has been done until now.

tacaswell:
@prjemian Can you expand on that a bit more?  I don't see a difference between clients having to check a version string vs knowing to check some other key in the documents.  In either case consumers that are not currently checking for something and are relying on the buggy array shape are going to be broken.   

I am not sure we want to commit to maintaining, forever, a option for the order of the shape key.  Further, I am not sure that AD is the only source of multi-dimensional data that is in EventModel so in general lacking what ever key we use to identify the order of the dimensions in shape.

@danielballan I think the discussion of extending the `data_key` should get elevated to an enhancement proposal.

@prjemian The `Descriptor` _should_ be accurately reporting the data that will be coming in the `Event` document, however the handlers that fetch that data do not consult the `Descriptor` and have the ground truth.

The fundamental issue here is that `ophyd` was buggy and sourcing bogus data for the shape. Thus, in my judgement this is a bug-fix not an API change and we do not need to bend over-backwards to accommodate the broken behavior going forward (In either case should be documented in the release notes).

@prjemian do you have clients that rely on looking at the shape? Dan's work on the data browser is our first client that is actually trying to use it (hence why it lasted for so long in a broken state).

Having a data audit tool that checks that the descriptors are actually correct for existing data would be interesting / useful (and we probably should go back into the database of fix the buggy descriptors (!!)).

As we only have an RC for 1.6, I think this should be backported to that release.

klauer:
I agree with the point that it's a bugfix and not an API break, with the version check being sufficient.

Let's do a quick survey of how we handle array shapes:
1. `PluginBase` https://github.com/bluesky/ophyd/blob/d3cd9b942d1fe1fd7a77355f95e5b7cfe8bb6289/ophyd/areadetector/plugins.py#L238-L243

2. `ImagePlugin.shaped_image`
https://github.com/bluesky/ophyd/blob/d3cd9b942d1fe1fd7a77355f95e5b7cfe8bb6289/ophyd/areadetector/plugins.py#L315-L320

3. `CamBase`, pre-PR refers to this directly:
https://github.com/bluesky/ophyd/blob/d3cd9b942d1fe1fd7a77355f95e5b7cfe8bb6289/ophyd/areadetector/cam.py#L68-L72

4. `CamBase`, post-PR makes this tuple:
https://github.com/bluesky/ophyd/blob/e4891cb64d6db8456b140f545eceba599a2a2515/ophyd/areadetector/detectors.py#L75-L77

Where we choose one style vs another seems worth documenting, as the above has already confused me...

tacaswell:
Do we want to also change (3) as well then? Getting transposed array sizes from `d.cam` vs `d.plugin` is pretty bad and I think we want to have c-style row-major slow-to-fast shape/index everywhere.

prjemian:
@tacaswell : you mentioned

> ... difference between clients having to check a version string vs knowing to check some other key in the documents

The proposition was that clients *might* have a version string in the metadata.  My reply was to have a *positive, definite means of announcing the shape info *.

But, in truth, APS has no code of which I am aware that relies on this content -- we have no legacy software to support.  I can agree that it is, indeed, a bug fix, especially so given [this survey](https://github.com/bluesky/ophyd/pull/746#issuecomment-508805688) by @klauer.  Documentation will describe it, then.

danielballan:
Indeed, the version string will be not be present if an old version of bluesky is used. I think our best path forward is for consumers to do "Check for a version string in the RunStartt document; assume old ophyd+blueksy if version is missing _and_ `shape` looks like `(X, Y, 0)`." Until we get around to having a way to fix old data via a migration.

We can cnosider a path to better ways to handle this kind of thing in general in an enhancement proposal.

danielballan:
> Do we want to also change (3) as well then? 

It seems a bit weird to override what AD exposes though. Other clients would likely show x, y, z.

tacaswell:
I am :+1: on changing 3 as well.   I think it is more important to remain consistent with ourselves than a hypothetical screen.

danielballan:
Would we then flip y and x in `size` and `max_size` as well, and the many other places that give coordinates that file?

I would like to resolve this one way or another in time for 1.4.0 if we can so that we can stop patching descriptors in the future.

danielballan:
In a phone discussion last week, we decided to sort out what order the image array waveform comes back in and then make all sub-devices that impose an order of x/y/z signals use that same order.

danielballan:
@prjemian Can you take the lead on this?

danielballan:
@dmgav, Would you be up for this?

ronpandolfi:
Was chasing this issue for a while today. For posterity (until merged), I corrected it by setting the camera component's num_images kind to 'config'.

danielballan:
@AbbyGi made an executive decision, pending someone speaking up to argue a different view. Everywhere that ophyd imposes an ordering on array dimensions, we will use of these orderings:

* y, x
* z, y, x
* depth, height, width

As @tacaswell commented above, it is most important for us to be consistent with ourselves. This is a Python- and numpy-based wrapping of ophyd so wherever it imposes structure on logical PV groups the structure should respect numpy ordering.

danielballan:
Looking close with @AbbyGi, I see that several of these signals have the current ordering _in their name_, as in `max_xy`, `gain_xy`, and several others. Swapping the order of `x` and `y` to be "consistent" with array shape would then be inconsistent with the signal's own name. We could add new signals, `max_yx` but I'm not sure that makes anything actually better.

There are really two different kinds of consistency under consideration here.

(1) Consistency of a signal's metadata with its data --- its reported `shape` in the metadata and its actual `array.shape`
(2) Consistency of coordinate systems _between_ signals

(1) is necessary for activities like working with `dask.array` and pre-allocating resources to load an image into generally. (2) is helpful when working with signals together, such as plotting a centroid on an image.

Let us keep this PR focused on (1).

tacaswell (approved):

danielballan:
I intend to do an integration test with `SimDetector` all the way through databroker before we merge, to make sure this actually solves the problem of making images open-able without repairing the metadata at data access time.

danielballan:
@AbbyGi and I verified using https://gist.github.com/danielballan/3419df60330fb1543ac6306462a2e6c0#file-interactive_test-py that this results in correct metadata in the descriptor, and that it flows through databroker correctly.

This raises a separate but related issue described in #934, but that is distinct from this. Merging.

PR #749, opened 2019-08-02
ENH: add EpicsPathSignal

klauer:
Docstrings and cleanups would follow if this is deemed reasonable.

Basic idea is that on a `Device` level (specifically `FilePlugin`) we should be specifying a `Cpt(EpicsPathSignal, 'FilePath', path_semantics='nt')` instead of a regular `EpicsSignal`. This would then take care of `set_and_wait_path` for us.

I believe the base `EpicsSignal` should be more easily customizable to allow such things without massive copy/pasting, but opted to keep changes minimal for this proof-of-concept.

Notes
* Alternative to https://github.com/bluesky/ophyd/pull/703
* `_ensure_trailing_slash` could then be removed, and classes would be modified to use this.
* This would need changing as well: https://github.com/bluesky/ophyd/blob/master/ophyd/areadetector/filestore_mixins.py#L432 (not sure why we aren't calling `.set()` and then `wait(status)` there instead?)

### DIFF
--- a/ophyd/areadetector/paths.py
+++ b/ophyd/areadetector/paths.py
@@ -95,14 +95,22 @@ class EpicsPathSignal(EpicsSignal):
         if string is not True:
             raise ValueError('Specifying an EpicsPathSignal with string=False'
                              ' does not make sense')
-        if path_semantics not in OS_NAME_TO_PATH_CLASS:
-            raise ValueError(f'Unknown path semantics: {path_semantics}. '
-                             f'Options: {OS_NAME_TO_PATH_CLASS}')
 
         super().__init__(write_pv=write_pv,
                          read_pv=f'{write_pv}_RBV',
                          string=True,
                          **kwargs)
+
+    @property
+    def path_semantics(self):
+        return self.path_semantics
+
+    @path_semantics.setter
+    def path_semantics(self, value):
+        if value not in OS_NAME_TO_PATH_CLASS:
+            raise ValueError(f'Unknown path semantics: {value}. '
+                             f'Options: {OS_NAME_TO_PATH_CLASS}')
+        self.path_semantics = value
 
     def _repr_info(self):
         yield from super()._repr_info()
--- a/ophyd/areadetector/paths.py
+++ b/ophyd/areadetector/paths.py
@@ -1,11 +1,10 @@
-import os
+import logging
 import pathlib
 import time as ttime
 
 from .. import EpicsSignal
-from ..status import Status
 
-
+logger = logging.getLogger(__name__)
 OS_NAME_TO_PATH_CLASS = {
     'nt': pathlib.PureWindowsPath,
     'posix': pathlib.PurePosixPath,
@@ -41,15 +40,16 @@ def path_compare(path_a, path_b, semantics):
 def set_and_wait_path(signal, val, *, path_semantics, poll_time=0.01,
                       timeout=10):
     """
-    Set a signal to a value and wait until it reads correctly.
-    For floating point values, it is strongly recommended to set a tolerance.
-    If tolerances are unset, the values will be compared exactly.
+    Set a path signal to a value and wait until it reads back correctly.
 
     Parameters
     ----------
     signal : EpicsPathSignal (or any object with `get` and `put`)
+        The signal itself
     val : object
         value to set signal to
+    path_semantics : {'nt', 'posix'}
+        The OS name for the path
     poll_time : float, optional
         how soon to check whether the value has been successfully set
     timeout : float, optional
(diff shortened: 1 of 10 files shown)


klauer:
Have a few minutes for some input on the idea, @tacaswell @danielballan @teddyrendahl @ZLLentz?

danielballan:
When I saw this PR title go up I thought, "Why didn't we think of this before?" Having a custom type with its out `set()` method feels like the right place to put this logic. :+1:

prjemian:
👍

danielballan:
I see this is still a draft. Any reason to wait to merge this?

klauer:
Yes, a good reason - I think the signal itself could be refactored slightly making a `set` hook much easier. The `EpicsPathSignal` itself would then only be a few lines long.

danielballan:
@ronpandolfi and I discussed this on today's Pilot call. This may be the right place to solve an issue that he has been having.

klauer:
I don't have tests just yet to mark this as ready to review, but perhaps @ronpandolfi could give this a try to see if it fixes the issue he ran into?

danielballan:
Bump @ronpandolfi

klauer:
We can perhaps bring this into pcdsdevices and see how it functions.

tacaswell:
Can we keep pushing on this here?

klauer:
You're welcome to continue the development, @tacaswell

tacaswell:
There isn't parallel work going on in pcdsdevices?

danielballan:
This just needs tests, right?

klauer:
Tests + integration into ophyd.areadetector

danielballan:
Aiming to finally get this into a release in the next week or two.

ophyd/areadetector/paths.py
```diff
@@ -0,0 +1,126 @@
...
+        'pvname' being the setpoint and 'pvname_RBV' being the read-back path.
+
+        Operating system-specific path semantics are respected when confirming
+        that a :meth:`.set()` operation has completed.
+        '''
+        if write_pv.endswith('_RBV'):
+            # Strip off _RBV if it was passed in erroneously
+            write_pv = write_pv[:-4]
+
+        self.path_semantics = path_semantics
```
danielballan, line 94: Make this a (settable) `property` so we can configure it easily.

ophyd/areadetector/paths.py
```diff
@@ -0,0 +1,126 @@
...
+        if string is not True:
+            raise ValueError('Specifying an EpicsPathSignal with string=False'
+                             ' does not make sense')
+        if path_semantics not in OS_NAME_TO_PATH_CLASS:
```
danielballan, line 98: ...and do this check inside the setter.

ophyd/tests/test_signal.py
```diff
@@ -561,19 +561,34 @@ def test_epicssignal_pv_reuse(cleanup, pvname, count):
...
                          [('C:\\some\\path\\here'),
-                          ('D:\\here\\is\\another\\')
+                          ('D:\\here\\is\\another\\'),
+                          ('C:/yet/another/path'),
+                          ('D:/more/paths/here/')
                           ])
-def test_windows_paths(path):
-    signal_nt = EpicsPathSignal('TEST', path_semantics='nt')
-    signal_nt.set(path).wait(3)
+def test_windows_paths(paths, cleanup, signal_test_ioc):
```
danielballan, line 579: This can use the `path_signal` fixture. It doesn’t need to access `signal_test_ioc` directly or do its cleanup.

ophyd/tests/test_epicspath.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from ..areadetector.paths import EpicsPathSignal
+
+
+def test_path_semantics_exception():
```
danielballan, line 6: It's probably not worth making a new test faile just for this one test. Could this be moved to `test_signal.py`?

danielballan (approved):
I have one last suggestion about organizing the tests, noted in line. Otherwise, this seems ready.

danielballan:
One more thing (as said over a private channel but reiterated here for the record) we should remove `_ensure_trailnig_slash` as well before we merge. It is not needed and is known to be broken.

danielballan:
The CI failure is real. It seems that `EpicsPathSignal`'s custom `set_and_wait` logic doesn't yet work as intended.

```
E               TimeoutError: Attempted to set EpicsPathSignal(read_pv='13SIM1:HDF1:FilePath_RBV', name='det_hdf1_file_path', parent='det_hdf1', value='/tmp/data1/2020/09/17/', timestamp=1600362472.980279, auto_monitor=False, string=True, write_pv='13SIM1:HDF1:FilePath', limits=False, put_complete=False, path_semantics='posix') to value '/tmp/data1/2020/09/17' and timed out after 10 seconds. Current value is '/tmp/data1/2020/09/17/'.
``

danielballan:
I think the problem is that we are directly using `set_and_wait(signal, value)` here, rather than `signal.set(value).wait(timeout)`. This circumvents `EpicsPathSignal`'s ability to use `set_and_wait_path` rather than old `set_and_wait`.

https://github.com/bluesky/ophyd/blob/f0e5dd744a9f65cbfa3f9e02a59c3c1f8fb53a3a/ophyd/areadetector/filestore_mixins.py#L430-L432

PR #751, opened 2019-08-03
FIX: functools.partial has no __name__

### LINKED ISSUE #750: dispatcher callback failure; functools.partial has no __name__
```
      File "/Users/klauer/docs/Repos/ophyd/ophyd/_dispatch.py", line 41, in run
        self.current_callback = (callback.__name__, kwargs.get('pvname'))
    AttributeError: 'functools.partial' object has no attribute '__name__'
```

klauer:
Closes #750

### DIFF
--- a/ophyd/_dispatch.py
+++ b/ophyd/_dispatch.py
@@ -38,7 +38,9 @@ class _CallbackThread(threading.Thread):
                 ...
             else:
                 try:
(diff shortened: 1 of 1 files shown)


klauer:
It does seem like it could be (should be?) solved at the Python level...

PR #755, opened 2019-08-13
ENH: specify for which device another set call is issued

mrakitin:
This is to be more specific which device has a second `set` call.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -253,7 +253,8 @@ class Signal(OphydObject):
                 del th
 
         if self._set_thread is not None:
-            raise RuntimeError('Another set() call is still in progress')
+            raise RuntimeError('Another set() call is still in progress '
+                               f'for {self.name}')
 
         st = Status(self)
         self._status = st


prjemian:
Most of the time, the only recovery I know is to restart the kernel.  Will this be part of how to repair instead of restart?

mrakitin:
Same for us at the moment. I think something like `<status>._finished(success=False)` can be used if the `<status>` object is exposed somehow.

prjemian:
Would `{self.dotted_name}` be more useful?  That would make the useful name available for copy/paste.  Not sure how to get from there to the status object, though.

mrakitin:
I like the idea, @prjemian.

mrakitin:
I actually tried it at a beamline, and the `<det>.dotted_name` is not available for ophyd v1.3.3, as it was added in [1.4.0rc1](https://blueskyproject.io/ophyd/release_notes.html#v1-4-0-pre-release). I think we should stick to the current implementation for now.

danielballan:
Do you intend to backport this to the 1.3.x series?

mrakitin:
Would be great to have it on the floor after the next push.

danielballan:
OK, that's reason enough to merge this as is. I opened #757 to remind us to follow up on the other ideas suggested here.

tacaswell:
> Will this be part of how to repair instead of restart?

This is usually a sign either your plan is broken or your hardware is behaving badly.  To "fix" it you can do `obj._set_thread = None`.

tacaswell:
backported to v1.3.x as 623cea9

PR #758, opened 2019-08-22
ENH add Lambda750kCam

prjemian:
support for [ADLambda](https://github.com/areaDetector/ADLambda) [X-Spectrum Lambda 750K](https://x-spectrum.de/products/lambda-350k750k/) area detector

We also have support for [the plugin chain we use now](https://github.com/aps-8id-trr/ipython-8idiuser/blob/master/profile_bluesky/startup/25-areadetector.py#L23-L72).  That plugin chain is likely to be redeveloped in the next year and might not be useful in ophyd, other than pedagogical value.

### DIFF
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -18,6 +18,7 @@ __all__ = ['CamBase',
            'FirewireLinDetectorCam',
            'FirewireWinDetectorCam',
            'GreatEyesDetectorCam',
+           'Lambda750kCam',
            'LightFieldDetectorCam',
            'Mar345DetectorCam',
            'MarCCDDetectorCam',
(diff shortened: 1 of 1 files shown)


tacaswell:
The scatter-gather plugins could probably come upstream.  I would be surprised if those go away...

klauer:
@tacaswell https://github.com/bluesky/ophyd/blob/7f1452d41271d353c3d7d0f6b8a93adef7dc0858/ophyd/areadetector/plugins.py#L1919

https://github.com/bluesky/ophyd/blob/7f1452d41271d353c3d7d0f6b8a93adef7dc0858/ophyd/areadetector/plugins.py#L2251

tacaswell:
ah, forgot @prjemian was working from an older version of ophyd.

prjemian:
Thanks.  I had local overrides for UPPER case in scatter and gather

PR #763, opened 2019-09-29
REF/ENH: do not require `self` for FormattedComponent

klauer:
Just an idea - see docstring in diff for the example.


prjemian:
FWIW: I'd enjoy this simplification.  It makes more sense.

klauer:
Seems like we're up +1.5. OK to merge?

tacaswell:
Rebased to patch up documentation fix, will merge when CI passes.

PR #769, opened 2019-11-04
Do not depend on caproto's experimental dependencies.

danielballan:
Some of caproto's *optional* dependencies are very experimental:
in particular, ``curio``, ``trio``, and ``asks``. They sometimes
introduce dependencies on bleeding-edge versions of things (most
recently, ``attrs``) that can cause packaging headaches in our other
downstream libraries. These optional requirements should not propagate
into ophyd.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,7 +2,7 @@ codecov
 matplotlib
 pytest
 databroker>=1.0.0b1
-caproto[standard]
+caproto[standard] >=0.4.2rc1
 h5py
 pytest-faulthandler
 pyepics>=3.4.0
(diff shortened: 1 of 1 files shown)


danielballan:
This needs https://github.com/caproto/caproto/pull/541

danielballan:
Caproto v0.4.2rc1 has been released with the caproto/caproto#541. I have updated this PR to pin to the RC. If CI, passes we should:

* Release caproto v0.4.2 final.
* Update this PR to pin to that instead of the RC.
* Merge this PR.
* Do an ophyd back-port release.

PR #771, opened 2019-11-04
Pin caproto requirement to separate experimental dependencies.

jklynch:
Back port of #769.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,6 +1,6 @@
 codecov
 coveralls
-caproto[standard]>=0.4.2rc1
+caproto[standard] >=0.4.2rc1
 matplotlib
-pytest>=3.6
+pytest >=3.6
 pytest-cov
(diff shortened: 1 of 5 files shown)


test-requirements.txt
```diff
@@ -1,5 +1,6 @@
 codecov
 coveralls
+caproto[standard]>=0.4.2rc1
 matplotlib
-pytest >=3.6
```
danielballan: I believe the spaces should stay, as [it can help with conda inter-operability](https://github.com/intake/intake/blob/aa83ccd41ed32f90f6bef5dbbe0e8a9e67c6d781/requirements.txt#L1). Is it causing a problem somewhere?

danielballan: Good point. I just checked the [approached we settled on in the cookiecutter](https://github.com/NSLS-II/scientific-python-cookiecutter/blob/a3e3da71b331d03565f50a23f96f911ea2cb16aa/%7B%7B%20cookiecutter.repo_name%20%7D%7D/setup.py#L32-L33) which was created well after ophyd. Looks like `f.read().splitlines()` would also work.

PR #774, opened 2019-11-14
remove deprecated kwarg "context"

### LINKED ISSUE #773: Backport a fix for pyepics context kwarg warning
This warning

```
/srv/conda/envs/notebook/lib/python3.7/site-packages/epics/pv.py:110: UserWarning: The `context` kwarg for epics.get_pv() is deprecated. New PVs will _not_ be created in the requested context.
```

mentioned also by @prjemian in #765 should be back-ported to 1.3.x since we still have some work to do before releasing 1.4.0.

jklynch:
This PR resolves #773.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ cache:
 
 matrix:
   include:
-    - python: 3.5
-      env: CONDA_ENV=1
     - python: 3.6
       env: BUILD_DOCS=true CONDA_ENV=1
     - python: 3.6
@@ -65,7 +63,7 @@ before_install:
 install:
   - |
     if [ $CONDA_ENV ]; then
-       conda create -n testenv python=$TRAVIS_PYTHON_VERSION numpy "pyepics>=3.4.0" "pytest>=3.6" pytest-cov databroker networkx -c lightsource2-tag -c conda-forge -c soft-matter -c defaults --override-channels
+       conda create -n testenv python=$TRAVIS_PYTHON_VERSION numpy pyepics "pytest>=3.6" pytest-cov databroker networkx -c lightsource2-tag -c conda-forge -c soft-matter -c defaults --override-channels
        source activate testenv
        export PYEPICS_LIBCA=$CONDA_PREFIX/lib/libca.so
        conda install readline -c lightsource2-tag
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ before_install:
 install:
   - |
     if [ $CONDA_ENV ]; then
-       conda create -n testenv python=$TRAVIS_PYTHON_VERSION numpy pyepics "pytest>=3.6" pytest-cov databroker networkx -c lightsource2-tag -c conda-forge -c soft-matter -c defaults --override-channels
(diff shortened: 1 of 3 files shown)


tacaswell:
I am a bit concerned about this as I think this is required to make ophyd work with older pyepics and I'm not sure we want to force that forward here.

jklynch:
On Nov 5 I asked @prjemian and @ZLLentz if dropping support for python 3.5 from ophyd 1.3.x would affect them and they said they have no python 3.5 deployments.

jklynch:
@tacaswell I guess you were talking about the deprecated keyword arg

jklynch:
There is some evidence this PR is not compatible with pyepics 3.3.0, but it is easy to upgrade to 3.4.0.

danielballan:
@jklynch If we want ophyd to _work_ with old pyeipcs (satisfying @tacaswell's concern above) but avoid issuing a warning with new pyepics (resolving #773) then we could remove the pin on `pyepics` and instead do a check to see which version of pyepics we have to decide whether to use the `context` kwarg or not.

```py
from distutils.version import LooseVersion    
LooseVersion(epics.__version__) >= LooseVersion('3.4.0')
```

ophyd/_pyepics_shim.py
```diff
@@ -19,8 +20,8 @@
 
 
 def get_pv(*args, **kwargs):
-    import epics
-    kwargs.setdefault('context', epics.ca.current_context())
+    if not LooseVersion(epics.__version__) >= LooseVersion('3.4.0'):
```
danielballan: Minor readability improvement by avoiding the double-negative.

```suggestion
    if LooseVersion(epics.__version__) < LooseVersion('3.4.0'):
```

danielballan: Might be worth linking, in a comment, to this commit where the change we are accounting for here was introduced. https://github.com/pyepics/pyepics/commit/1ca46afa5afe285fecd4ea1d12a7eeee0376c773

h/t @tacaswell for digging that up.

.travis.yml
```diff
@@ -22,8 +22,6 @@ cache:
 
 matrix:
   include:
-    - python: 3.5
```
tacaswell: do we have to drop 3.5 testing?  Although I initially suggested we do that, I thought we settled on removing the one f-string in a backport.

tacaswell:
pyepics 3.4 is the right version to check, came in via https://github.com/pyepics/pyepics/commit/1ca46afa5afe285fecd4ea1d12a7eeee0376c773

.travis.yml
danielballan: It's not about language features we're using; it's about packages available from conda. (Note that this is a `CONDA_ENV=1` build.) However, now that we have (rightly!) removed the pyepics pin that was once in this PR, perhaps we can restore the 3.5 build and remove the `python_requires`.

danielballan: To underline the point: thanks for the close read, @tacaswell!

tacaswell: fair cop.

PR #779, opened 2019-12-07
Changes from NSLS-II SIX for clearer error messages about timeouts

mrakitin:
This is a very volatile, work-in-progress PR to the **v1.3.x** branch, to save the state of the code used at the NSLS-II SIX beamline in the recent attempts to mitigate the random timeout errors. See https://github.com/NSLS-II-SIX/profile_collection/issues/24 (and https://github.com/NSLS-II-SIX/profile_collection/issues/20?) for the context.

### DIFF
--- a/ophyd/utils/epics_pvs.py
+++ b/ophyd/utils/epics_pvs.py
@@ -200,7 +200,7 @@ def raise_if_disconnected(fcn):
     return wrapper
 
 
-def set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
+def set_and_wait(signal, val, poll_time=0.01, timeout=60, rtol=None,
                  atol=None):
     """Set a signal to a value and wait until it reads correctly.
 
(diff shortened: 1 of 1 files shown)


ophyd/utils/epics_pvs.py
```diff
@@ -265,15 +266,18 @@ def set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
...
     if enums:
         # convert enum values to strings if necessary first:
         if not isinstance(a, str):
             a = enums[a]
         if not isinstance(b, str):
             b = enums[b]
         # then compare the strings
-        return a == b
+        ret = a == b
+        if ret and signal_name == 'rixscam_hdf5_capture':
```
mrakitin: I know, this should be parametrized or removed.

klauer:
Is this worthwhile to keep open @mrakitin?

mrakitin:
@klauer, I think this can be closed, but I'd like to keep the branch around. The fix is not general enough, so in case of a follow-up update there will be a separate PR.

PR #781, opened 2019-12-11
FIX: Restore SignalRO

### LINKED ISSUE #780: SignalRO - Compatibility Broken
The changes added with https://github.com/bluesky/ophyd/commit/7e8d56b71122f2d8a12399f37824be0326a2e666 ended up breaking code at Typhon where `SignalRO` is being used.

hhslepicka:
This PR restores SignalRO which was being heavily used by Typhon for tests.
Closes #780 

Attn. @klauer

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -409,7 +409,7 @@ class SignalRO(Signal):
         raise ReadOnlyError("The signal {} is readonly.".format(self.name))
 
     def set(self, value, *, timestamp=None, force=False):
-        raise ReadOnlyError("The signal {} is readonly".format(self.name))
+        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
 
 
 class DerivedSignal(Signal):
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -86,6 +86,21 @@ class EnumSignal(Signal):
         return desc
 
 
+class SignalRO(Signal):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._metadata.update(
+            connected=True,
+            write_access=False,
+        )
+
+    def put(self, value, *, timestamp=None, force=False):
+        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
+
+    def set(self, value, *, timestamp=None, force=False):
+        raise ReadOnlyError("The signal {} is readonly".format(self.name))
+
+
 class SynSignal(Signal):
     """
     A synthetic Signal that evaluates a Python function when triggered.


hhslepicka:
@mrakitin nope... We need just the most simple signal, no fancy generation of value via function... 
SignalRO basically received the value and that was it.

hhslepicka:
@mrakitin no idea.

mrakitin:
@hhslepicka, maybe it's a good chance to move it to the "proper" location? The read-only version of a generic `Signal` seems to be useful outside of the simulation framework.

Do you mind to move the function before this PR gets merged? Would it be much work to change the imports in typhon?

hhslepicka:
@mrakitin not much work just a bunch of try/except for the ImportErrors to keep backwards compatibility.

danielballan:
I put a PR against your PR to suggest another possibility: _define_ in it `ophyd.signal` to keep things organized but expose it in `ophyd.sim` to un-break PyDM without imposing a bunch of import updates. https://github.com/hhslepicka/ophyd/pull/1

hhslepicka:
Merged... it looks perfect @danielballan . 
Thanks for looking into it...

ophyd/signal.py
```diff
@@ -397,6 +397,21 @@ def _run_metadata_callbacks(self):
...
+        self._metadata.update(
+            connected=True,
+            write_access=False,
+        )
+
+    def put(self, value, *, timestamp=None, force=False):
+        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
+
+    def set(self, value, *, timestamp=None, force=False):
+        raise ReadOnlyError("The signal {} is readonly".format(self.name))
```
mrakitin:
```suggestion
        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
```

mrakitin:
Great solution, @danielballan! I added a minor punctuation suggestion to have a consistent error message across the methods.

ophyd/signal.py
hhslepicka: Addressed with latest commit.

mrakitin: Thanks!

hhslepicka:
I don't have commit privileges here...

ophyd/signal.py
```diff
@@ -397,6 +397,21 @@ def _run_metadata_callbacks(self):
...
+        self._metadata.update(
+            connected=True,
+            write_access=False,
+        )
+
+    def put(self, value, *, timestamp=None, force=False):
+        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
+
+    def set(self, value, *, timestamp=None, force=False):
+        raise ReadOnlyError("The signal {} is readonly.".format(self.name))
```
mrakitin: Can it be implemented as placeholders, something like the snippet below?
```py
    def put(self, *args, **kwargs):
         raise ReadOnlyError("The signal {} is readonly.".format(self.name))

    def set(self, *args, **kwargs):
        self.put(*args, **kwargs)
```

This way one does not need to sync error messages in multiple places. @hhslepicka, @danielballan, @klauer?

klauer: `set = put`

mrakitin: That works for me too.

hhslepicka: I will just merge it now and we can discuss that later on...

mrakitin: Sounds good.

PR #783, opened 2019-12-13
resolve EpicsMotor stalls sometimes at end of move

### LINKED ISSUE #782: motor motion fails to complete #1
As reported https://github.com/prjemian/ipython-vm7/issues/1, 

> Sometimes, motor motion fails to complete. First noticed that recently when testing `bec.LivePlot` problems for the APS USAXS instrument, running a specific sequence of activities within an infinite loop. At random time, the sequence waits indefinitely for the motor. The motor is an EPICS soft motor record running in a self-hosted linux VM on the same workstation as the bluesky session.
> 
> In EPICS, the motor has reached position and the DMOV is 1. In bluesky, the move is waiting to complete but has reached 100%.

This test code replicates the problem (and assume there is an EPICS motor record `sky:m1` available):

```
#!/usr/bin/env python

from bluesky import RunEngine
from bluesky import SupplementalData
from bluesky.callbacks.best_effort import BestEffortCallback
from bluesky.utils import get_history
from bluesky.utils import ProgressBarManager
import bluesky.plan_stubs as bps
from databroker import Broker
import epics
import logging
from ophyd import EpicsMotor
import time

# logging.basicConfig(level=logging.DEBUG)
REPETITIONS = 10
CYCLES = 250
DELAY_S = 1e-6

bec = BestEffortCallback()
db = Broker.named("mongodb_config")
sd = SupplementalData()
pbar_manager = ProgressBarManager()

RE = RunEngine(get_history())
# RE.subscribe(db.insert)
RE.subscribe(bec)
# RE.preprocessors.append(sd)
# RE.waiting_hook = pbar_manager

m1 = EpicsMotor("sky:m1", name="m1")
while not m1.connected:
    time.sleep(.05)

def move(motor, label, dest, delay_s):
    yield from bps.checkpoint()
    yield from bps.mv(m1, dest)
    print(f"{label}:   {dest} {motor.position}")
    yield from bps.sleep(delay_s)

def ping_pong(motor, v1, v2, cycles=100, delay_s=1e-2):
    for i in range(cycles):
        yield from move(m1, f"ping {i+1}", v1, delay_s)
        yield from move(m1, f"pong {i+1}", v2, delay_s)

if __name__ == "__main__":
    RE(ping_pong(m1, 1, -1, cycles=CYCLES, delay_s=DELAY_S))
```

The `ping_pong()` plan usually stalled after a few cycles, at random either on ping or pong.

prjemian:
fixes #782

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -145,8 +145,18 @@ class EpicsMotor(Device, PositionerBase):
             If motion fails other than timing out
         '''
         self._started_moving = False
+        status = None
 
+        def dmov_callback(old_value, value, **kwargs):
+            self.log.debug(f'DEBUG: dmov_callback(): new={value} old={old_value}')
+            if self._started_moving and value == 1:
+                self.motor_done_move.clear_sub(dmov_callback)
+                status._finished(success=True)
+                status.done = True
+
+        self.motor_done_move.subscribe(dmov_callback)
         status = super().move(position, **kwargs)
+
         self.user_setpoint.put(position, wait=False)
         try:
             if wait:


ophyd/epics_motor.py
```diff
@@ -145,8 +145,18 @@ def move(self, position, wait=True, **kwargs):
             If motion fails other than timing out
         '''
         self._started_moving = False
+        status = None
```
mrakitin: Is it needed here? `None` is not valid within the callback below, and overwritten on line 158 anyway.

danielballan: I don't think we need this line. Closures are allowed to refer to variables that are defined below them, as in:

```py
In [1]: def f(): 
   ...:     print(a) 
   ...:                                                                                                                                                                                       

In [2]: a = 1                                                                                                                                                                                 

In [3]: f()                                                                                                                                                                                   
1
```

ophyd/epics_motor.py
```diff
@@ -145,8 +145,18 @@ def move(self, position, wait=True, **kwargs):
...
 
+        def dmov_callback(old_value, value, **kwargs):
+            self.log.debug(f'DEBUG: dmov_callback(): new={value} old={old_value}')
```
danielballan:
```suggestion
            self.log.debug('dmov_callback(): new=%f, old=%f', value, old_value)
```
Two things here:
- The logging formatter will add the "DEBUG:" part for us; it shouldn't be in the message body. (It's very possible we aren't consistent about this in ophyd. The logging overall needs a thorough review.)
- A bit of Python logging best practice... Although we love and encourage f-strings in general, for log messages in particular the old-style string templating is recommended because the logging framework can ensure that the templating will only be performed if that log message is actually issued. That is, if there is nothing handling DEBUG-level log messages, then you won't pay the performance overhead of building this string. Note that the values are passed as arguments to `debug`, like `debug('...', a, b)` not `'debug(...' % (a, b))`. The formatting is deferred and only happens if it needs to. "If a {Python log message} falls in a forest and there's no one there to hear it...."

ophyd/epics_motor.py
```diff
@@ -145,8 +145,18 @@ def move(self, position, wait=True, **kwargs):
...
+            if self._started_moving and value == 1:
+                self.motor_done_move.clear_sub(dmov_callback)
+                status._finished(success=True)
+                status.done = True
```
danielballan: Calling `_finished` sets this. It need not and probably should not be done explicitly. We should make this a property so that that is unambigous.

ophyd/epics_motor.py
```diff
@@ -145,8 +145,18 @@ def move(self, position, wait=True, **kwargs):
...
+
+        self.motor_done_move.subscribe(dmov_callback)
```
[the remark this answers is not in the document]
tacaswell: You also have to remember to un-subscribe this or we will have as many callbacks as we have had motor moves!

tacaswell:
Much of this logic is already in https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/epics_motor.py#L224-L271

I suspect that this is going to interact badly with that logic.

----

It is subscribed a bit differently on the 1.3.x branch 

https://github.com/bluesky/ophyd/blob/9f21ca2690a4a8a1db86d2eeb9a5628e9fc03123/ophyd/epics_motor.py#L81 and https://github.com/bluesky/ophyd/blob/9f21ca2690a4a8a1db86d2eeb9a5628e9fc03123/ophyd/epics_motor.py#L222-L267

danielballan:
Haha, I see three of us pounced in parallel.Thanks for the contribution, @prjemian, and good luck sifting through our feedback. :upside_down_face:

ophyd/epics_motor.py
prjemian: I agree.  Only added it to suppress warnings when logging was set to DEBUG.

ophyd/epics_motor.py
prjemian: Agreed.  It's gone.

ophyd/epics_motor.py
prjemian: I hear you.  `DEBUG: ` is out and back to old style strings.

ophyd/epics_motor.py
prjemian: not needed here, deleted it

ophyd/epics_motor.py
```diff
@@ -145,8 +145,18 @@ def move(self, position, wait=True, **kwargs):
             If motion fails other than timing out
         '''
         self._started_moving = False
+        status = None
 
+        def dmov_callback(old_value, value, **kwargs):
+            self.log.debug(f'DEBUG: dmov_callback(): new={value} old={old_value}')
+            if self._started_moving and value == 1:
```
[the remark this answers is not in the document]
prjemian: Good question.  This is tangly code.

ophyd/epics_motor.py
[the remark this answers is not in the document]
prjemian: Isn't this the matching unsubscribe?  `self.motor_done_move.clear_sub(dmov_callback)`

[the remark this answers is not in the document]
prjemian: Good move: `run=False`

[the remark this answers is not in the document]
prjemian: What about this?

```
                self.motor_done_move.clear_sub(dmov_callback)
                self.motor_done_move.unsubscribe(cbid)
```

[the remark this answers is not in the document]
prjemian: with:

```
        cbid = self.motor_done_move.subscribe(dmov_callback, run=False)
```

prjemian:
> 
> 
> Much of this logic is already in
> 
> https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/epics_motor.py#L224-L271
> 
> I suspect that this is going to interact badly with that logic.
> 
> It is subscribed a bit differently on the 1.3.x branch
> 
> https://github.com/bluesky/ophyd/blob/9f21ca2690a4a8a1db86d2eeb9a5628e9fc03123/ophyd/epics_motor.py#L81
> and
> 
> https://github.com/bluesky/ophyd/blob/9f21ca2690a4a8a1db86d2eeb9a5628e9fc03123/ophyd/epics_motor.py#L222-L267

This might be the better place for the logic in the `dmov_callback()`.  This logic:

```
        was_moving = self._moving
        self._moving = (value != 1)
```

should change to:

```
        was_moving = self._moving
        self._moving = not self.motor_is_done.value
```


This logic is problematic:

```
        if was_moving and not self._moving:
```

since it requires the transition to declare the move complete.  Suggest relocating the done_moving()` call thus:

```
        if not self._started_moving and not self._moving:
            self._done_moving(success=success, timestamp=timestamp,
                              value=value)
```

ophyd/epics_motor.py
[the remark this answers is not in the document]
prjemian: It's in the `_move_changed()` method:

```
        if not self._started_moving:
            started = self._started_moving = (not was_moving and self._moving)
```

prjemian:
So, removing the `dmov_callback()` from `move()` and adding to the `_move_changed()` method as described results in the motor stall re-appearing.  Where is the status message declared done?  It's not directly in `positioner._done_moving()`.  Very tangly code.  Is it one of the `st._finished()` lines in `signal.py`?

prjemian:
In the end, I restored the `dmov_callback()` to *ensure* the status message is finished.  There is no problem with calling this more than once, the code has an early bailout for

```
        if self.done:
            return
```

Is there a better way to be certain that `st._finished()` gets called?

ophyd/epics_motor.py
danielballan: Done here: https://github.com/bluesky/ophyd/pull/784

ophyd/epics_motor.py
```diff
@@ -146,7 +146,17 @@ def move(self, position, wait=True, **kwargs):
         '''
         self._started_moving = False
 
+        def dmov_callback(old_value, value, **kwargs):
+            self.log.debug('dmov_callback(): new=%d old=%d', value, old_value)
+            if self._started_moving and value == 1:
+
+                self.motor_done_move.clear_sub(dmov_callback)
```
tacaswell: Only one of these should be required.

tacaswell:
> since it requires the transition to declare the move complete

My understanding is that the motor record will always flash DMOV low and then high to indicate that it is moving (even if it already at it's target destination).

I am still :-1: on this as-is because it is short-circuiting a whole bunch logic in `_move_changed` which I assume is there for a reason.  If there are cases where that logic is currently correctly _not_ reporting the motion done, this extra callback will flip the status and erroneously report that motion is done when it is not.

Do we have a clear understanding of the source of the problem?  @prjemian do you have any timing diagrams of what the various PVs are doing when it does and does not work correctly?

My current guess is that we are hitting race-conditions on some of the internal state of `EpicsMotor` between the code running on the monitor thread and the code running on the main thread that is causing the next `m.move()` execution to overlap with the previous `_move_changed` execution.  Thus, we miss the falling edge of DMOV to indicate that we started moving for the move the user requested and then ignore the rising edge because we think that does not belong to us.

----

In https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/positioner.py#L195-L198

we subscribe the status object to `self._SUB_REQ_DONE` on the ophyd object and then in https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/positioner.py#L200-L208 process those callbacks which is in turn called from https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/epics_motor.py#L270-L271

prjemian:
@tacaswell : You have described the actions of the .MOVN field of the motor record.  The .DMOV transitions from 1 to 0 at the start of a move and transitions from 0 to 1 at the end, after the completion of the primary move plus any additional move complexities (such as backlash corrections or retries) controlled by the motor record.  See this documentation for the details:  https://epics.anl.gov/bcda/synApps/motor/motorRecord.html#Fields_status

About the .DMOV field, the documentation states:

> This field is set to 0 when the motor record begins a motion, and remains 0 through any retries and backlash corrections that may be required until the motor record has completely finished that motion, whereupon the field is set to 1. DMOV is guaranteed to execute and post a 1/0/1 pulse when the motor is commanded to move--even if no motion actually occurs because the motor was commanded to move to its current position.

prjemian:
It is clear from the structure of the plan used to test this problem (when executed with logging level set to DEBUG):

```
def move(motor, label, dest, delay_s):
    yield from bps.checkpoint()
    yield from bps.mv(m1, dest)
    print(f"{label}:   {dest} {motor.position}")
    yield from bps.sleep(delay_s)
```

that the next move has not yet begun.  If the next move command had been issued, the RunEngine logging would indicate a `checkpoint` message had been processed since the most recent move activity.  What makes this obvious is that no `checkpoint` message is logged at the time the motor stalls.  The logging indicates the motor has finished its most recent move yet the RunEngine is waiting.

prjemian:
Is there a better way to be certain that st._finished() gets called?

prjemian:
I agree that this change has implications of change for additional methods such as `_move_changed()` and/or `_done_moving()`.  What tools are there to diagnose further the handling of the status objects?

tacaswell:
What is the behavior of DMOV is the motor is already moving and then we instruct it to move again?

The monitor subscriptions are run on a background thread (one sourced via libca and then trampolined to python).

The conflict is between the current move and the _previous_ move's clean up code.

prjemian:
If the .VAL field is changed while the motor is moving (.DMOV = 0), the .DMOV field will remain at zero until the new move is finished.

prjemian:
> The monitor subscriptions are run on a background thread (one sourced via libca and then trampolined to python).

What is *trampolined*?

> The conflict is between the current move and the _previous_ move's clean up code.

How can I diagnose the code in the current master at this level?

prjemian:
Here's an example of the timing of changes in motor VAL, MOVN, and DMOV fields using this command: `camonitor sky:m1.{VAL,MOVN,DMOV}`

**before the move**

```
sky:m1.VAL                     2019-12-14 06:23:54.200602 -0.1  
sky:m1.MOVN                    2019-12-14 06:23:54.200602 0  
sky:m1.DMOV                    2019-12-14 06:23:54.200602 1  
```

**step by 1.0**

```
sky:m1.DMOV                    2019-12-14 06:23:54.200602 0  
sky:m1.VAL                     2019-12-16 17:07:04.776139 0.9  
sky:m1.MOVN                    2019-12-16 17:07:04.776439 1  
sky:m1.MOVN                    2019-12-16 17:07:06.048334 0  
sky:m1.DMOV                    2019-12-16 17:07:06.048334 1
```

**step by 1.0, then step by 1.0 during the first move**

```
sky:m1.DMOV                    2019-12-16 17:07:06.048334 0  
sky:m1.VAL                     2019-12-16 17:07:23.618395 1.9  
sky:m1.MOVN                    2019-12-16 17:07:23.620621 1  
sky:m1.VAL                     2019-12-16 17:07:24.163468 2.9  
sky:m1.VAL                     2019-12-16 17:07:24.840445 2.9  
sky:m1.MOVN                    2019-12-16 17:07:24.840445 0  
sky:m1.MOVN                    2019-12-16 17:07:24.840492 1  
sky:m1.MOVN                    2019-12-16 17:07:26.143724 0  
sky:m1.DMOV                    2019-12-16 17:07:26.143724 1
```

**step by -1.0 (with backlash correction)**

```
sky:m1.DMOV                    2019-12-16 17:11:47.487712 0  
sky:m1.VAL                     2019-12-16 17:12:13.615191 1.9  
sky:m1.MOVN                    2019-12-16 17:12:13.615351 1  
sky:m1.MOVN                    2019-12-16 17:12:14.883530 0  
sky:m1.MOVN                    2019-12-16 17:12:14.883580 1  
sky:m1.MOVN                    2019-12-16 17:12:15.184805 0  
sky:m1.DMOV                    2019-12-16 17:12:15.184805 1  
```

**step by -1.0 (with backlash correction), then step by -1.0 during the first move**

```
sky:m1.DMOV                    2019-12-16 17:12:15.184805 0  
sky:m1.VAL                     2019-12-16 17:12:26.397074 0.9  
sky:m1.MOVN                    2019-12-16 17:12:26.397250 1  
sky:m1.VAL                     2019-12-16 17:12:26.997775 -0.1  
sky:m1.VAL                     2019-12-16 17:12:27.622810 -0.1  
sky:m1.MOVN                    2019-12-16 17:12:27.622810 0  
sky:m1.DMOV                    2019-12-16 17:12:27.622810 0  
sky:m1.MOVN                    2019-12-16 17:12:27.622865 1  
sky:m1.MOVN                    2019-12-16 17:12:28.925954 0  
sky:m1.MOVN                    2019-12-16 17:12:28.926168 1  
sky:m1.MOVN                    2019-12-16 17:12:29.226441 0  
sky:m1.DMOV                    2019-12-16 17:12:29.226441 1  
```

tacaswell:
> What is _trampolined_?

The code for this now lives in https://github.com/bluesky/ophyd/blob/master/ophyd/_dispatch.py .  Due to how contexts work in libca/pyepics (this is the c-level structure that holds the various sockets / threads etc required to handle the asynchronous part of the epics protocol and manages the socket <-> virtual circuit <-> channel mapping) running python code in response to a monitor that does further caput/caget will fail.  I do not remember exactly why this is off the top of my head (maybe @klauer can remind me ;) ).  Instead the function we actually register with pyepics just "bounces" the values to a second queue that we process in a python thread which gets around this problem.

---

I think we just have to start putting in logging statements and sort out the timing between the things happening on different threads and at the IOC level.  I don't know any shortcut to tracking down race conditions, but at least we have a way to reliably reproduce this one!

prjemian:
Thanks for the explanation, @tacaswell !  The next thing to do is to return to the master branch and begin diagnosis with that code, to determine what  needs to be changed so the motor stall condition does not recur.  I'll pause further work on this branch until that diagnosis is fruitful.

prjemian:
Will make a VM image available to the DAMA team.

prjemian:
> Do we have a clear understanding of the source of the problem? @prjemian do you have any timing diagrams of what the various PVs are doing when it does and does not work correctly?

`Virtual Circuit Unresponsive` from EPICS `libca`: See the three adjacent comments starting with: https://github.com/bluesky/ophyd/pull/787#issuecomment-568296988

prjemian:
Possibly related to #788?  Will test new ophyd release 1.4.0 to see if this remains a problem.

prjemian:
Need to stress test this with current release to see if still a problem.

PR #784, opened 2019-12-16
Deprecate settable done attribute.

danielballan:
It should never have been possible to set `status.done` directly. This
PR makes doing so issue a warning. In the N+1 release we should remove
the setter entirely.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -46,7 +46,7 @@ class StatusBase:
         self._tname = None
         self._lock = RLock()
         self._callbacks = deque()
-        self.done = done
+        self._done = done
         self.success = success
         self.timeout = None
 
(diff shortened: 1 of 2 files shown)


ophyd/status.py
```diff
@@ -69,6 +69,32 @@ def __init__(self, *, timeout=None, settle_time=None, done=False,
             self._timeout_thread = thread
             self._timeout_thread.start()
 
+    @property
+    def done(self):
+        """
+        Boolean indicating whether associated operation has completed.
+
+        This be set to True at __init__ time or by calling `_finished()`. Once
```
mrakitin:
```suggestion
        This is set to True at __init__ time or by calling `_finished()`. Once
```

ophyd/status.py
```diff
@@ -69,6 +69,32 @@ def __init__(self, *, timeout=None, settle_time=None, done=False,
...
+        True, it can be never become False.
```
mrakitin:
```suggestion
        True, it can never become False.
```

mrakitin:
Power-cycled the build on CI, so it passes now.

PR #785, opened 2019-12-16
add logging to Signal

jklynch:
This PR proposes
  - automatic logging configuration on import of ophyd/log.py - taken from caproto with one change described below
  - logging calls in the Signal class

The logging configuration in caproto is used as a template, but the file handler has been changed to TimedRotatingFileHandler set to rollover to a new file every week and to keep the last four log files.

### DIFF
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     curses = None
 
-__all__ = ('set_handler',)
+__all__ = ('config_ophyd_logging', 'get_handler', 'set_handler',)
 
 
 def _stderr_supports_color():
@@ -24,8 +24,7 @@ def _stderr_supports_color():
                 if curses.tigetnum("colors") > 0:
                     return True
             elif colorama:
-                if sys.stderr is getattr(colorama.initialise, 'wrapped_stderr',
-                                         object()):
+                if sys.stderr is getattr(colorama.initialise, 'wrapped_stderr', object()):
                     return True
     except Exception:
         # Very broad exception handling because it's always better to
@@ -128,20 +127,34 @@ class LogFormatter(logging.Formatter):
 plain_log_format = "[%(levelname)1.1s %(asctime)s.%(msecs)03d %(module)s:%(lineno)d] %(message)s"
 color_log_format = ("%(color)s[%(levelname)1.1s %(asctime)s.%(msecs)03d "
                     "%(module)s:%(lineno)d]%(end_color)s %(message)s")
+
+
+def validate_level(level) -> int:
+    """
+    Return aN int for level comparison
+    """
+    if isinstance(level, int):
+        levelno = level
+    elif isinstance(level, str):
+        levelno = logging.getLevelName(level)
+
+    if isinstance(levelno, int):
+        return levelno
+    else:
+        raise ValueError("Your level is illegal, please use 'ERROR', 'WARNING', 'INFO', 'DEBUG', OR 'TRACE'.")
+
+
 logger = logging.getLogger('ophyd')
 
 
 current_handler = None  # overwritten below
 
 
-def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
+def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level='WARNING'):
     """
     Set a new handler on the ``logging.getLogger('ophyd')`` logger.
-
-    This function is run at import time with default paramters. If it is run
-    again by the user, the handler from the previous invocation is removed (if
-    still present) and replaced.
-
+    If this is called more than once, the handler from the previous invocation
+    is removed (if still present) and replaced.
     Parameters
     ----------
     file : object with ``write`` method or filename string
@@ -150,42 +163,44 @@ def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
         Date format. Default is ``'%H:%M:%S'``.
     color : boolean
         Use ANSI color codes. True by default.
-
+    level : str or int
+        Python logging level, given as string or corresponding integer.
+        Default is 'WARNING'.
     Returns
     -------
     handler : logging.Handler
         The handler, which has already been added to the 'ophyd' logger.
-
     Examples
     --------
     Log to a file.
-
-    >>> set_handler(file='/tmp/what_is_happening.txt')
-
+    >>> config_ophyd_logging(file='/tmp/what_is_happening.txt')
     Include the date along with the time. (The log messages will always include
     microseconds, which are configured separately, not as part of 'datefmt'.)
-
-    >>> set_handler(datefmt="%Y-%m-%d %H:%M:%S")
-
+    >>> config_ophyd_logging(datefmt="%Y-%m-%d %H:%M:%S")
     Turn off ANSI color codes.
-
-    >>> set_handler(color=False)
+    >>> config_ophyd_logging(color=False)
+    Increase verbosity: show level INFO or higher.
+    >>> config_ophyd_logging(level='INFO')
     """
     global current_handler
     if isinstance(file, str):
         handler = logging.FileHandler(file)
     else:
         handler = logging.StreamHandler(file)
+    levelno = validate_level(level)
+    handler.setLevel(levelno)
     if color:
-        format = color_log_format
+        log_format = color_log_format
     else:
-        format = plain_log_format
+        log_format = plain_log_format
     handler.setFormatter(
-        LogFormatter(format, datefmt=datefmt))
+        LogFormatter(log_format, datefmt=datefmt))
     if current_handler in logger.handlers:
         logger.removeHandler(current_handler)
     logger.addHandler(handler)
     current_handler = handler
+    if logger.getEffectiveLevel() > levelno:
+        logger.setLevel(levelno)
     return handler
 
 
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -151,11 +151,16 @@ logger = logging.getLogger('ophyd')
 current_handler = None  # overwritten below
 
 
-def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level='WARNING'):
+def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level='INFO', backupCount=4):
     """
     Set a new handler on the ``logging.getLogger('ophyd')`` logger.
     If this is called more than once, the handler from the previous invocation
     is removed (if still present) and replaced.
+
+    If a file path is specified a TimedRotatingLogHandler will be used. It will be
+    configured so that a new log file starts every Monday and 4 log files are kept.
+    Log files older than 4 weeks will be deleted.
+
     Parameters
     ----------
     file : object with ``write`` method or filename string
@@ -166,7 +171,9 @@ def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level=
         Use ANSI color codes. True by default.
     level : str or int
         Python logging level, given as string or corresponding integer.
-        Default is 'WARNING'.
+        Default is 'INFO'.
+    backupCount : int
+        Number of historical log files to keep. Default is 4.
     Returns
     -------
     handler : logging.Handler
@@ -180,8 +187,8 @@ def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level=
     >>> config_ophyd_logging(datefmt="%Y-%m-%d %H:%M:%S")
     Turn off ANSI color codes.
     >>> config_ophyd_logging(color=False)
-    Increase verbosity: show level INFO or higher.
-    >>> config_ophyd_logging(level='INFO')
+    Increase verbosity: show level DEBUG or higher.
+    >>> config_ophyd_logging(level='DEBUG')
     """
     global current_handler
     if isinstance(file, str):
@@ -208,9 +215,11 @@ def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level=
         logger.setLevel(levelno)
     return handler
 
+
 # Add a handler with the default parameters at import time.
 config_ophyd_logging()
 set_handler = config_ophyd_logging  # for back-compat
+
 
 def get_handler():
     """
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -2,6 +2,7 @@
 # Apache 2.0. See other_licenses/ in the repository directory.
 
 import logging
+import logging.handlers
 import sys
 try:
     import colorama
@@ -184,7 +185,11 @@ def config_ophyd_logging(file=sys.stdout, datefmt='%H:%M:%S', color=True, level=
     """
     global current_handler
     if isinstance(file, str):
-        handler = logging.FileHandler(file)
+        handler = logging.handlers.TimedRotatingFileHandler(
+            filename=file,
+            when="W0",                 # rollover every Monday
+            backupCount=backupCount,   # keep the most recent 4 log files
+        )
     else:
         handler = logging.StreamHandler(file)
     levelno = validate_level(level)
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -261,8 +261,6 @@ def config_ophyd_logging(
     return handler
 
 
-# Add a handler with the default parameters at import time.
-config_ophyd_logging()
 set_handler = config_ophyd_logging  # for back-compat
 
 
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -191,9 +191,9 @@ def config_ophyd_logging(
     If this is called more than once, the handler from the previous invocation
     is removed (if still present) and replaced.
 
-    If a file path is specified a TimedRotatingLogHandler will be used. It will be
-    configured so that a new log file starts every Monday and 4 log files are kept.
-    Log files older than 4 weeks will be deleted.
+    If a file path is specified a TimedRotatingLogHandler will be used. By default
+    it will be configured so that a new log file starts every Monday and 4 log files
+    are kept. Log files older than 4 weeks will be deleted.
 
     Parameters
     ----------
@@ -207,7 +207,9 @@ def config_ophyd_logging(
         Python logging level, given as string or corresponding integer.
         Default is 'WARNING'.
     backupCount : int
-        Number of historical log files to keep. Default is 4.
+        Number of historical log files to keep. Default is 4. Use 0 to
+        keep all historical log files. The argument name is chosen to match
+        TimedRotatingFileHandler's backupCount.
     Returns
     -------
     handler : logging.Handler
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -183,16 +183,12 @@ current_handler = None  # overwritten below
 
 
 def config_ophyd_logging(
-    file=sys.stdout, datefmt="%H:%M:%S", color=True, level="WARNING", backupCount=4
+    file=sys.stdout, datefmt="%H:%M:%S", color=True, level="WARNING"
 ):
     """
     Set a new handler on the ``logging.getLogger('ophyd')`` logger.
     If this is called more than once, the handler from the previous invocation
     is removed (if still present) and replaced.
-
-    If a file path is specified a TimedRotatingLogHandler will be used. By default
-    it will be configured so that a new log file starts every Monday and 4 log files
-    are kept. Log files older than 4 weeks will be deleted.
 
     Parameters
     ----------
@@ -205,10 +201,6 @@ def config_ophyd_logging(
     level : str or int
         Python logging level, given as string or corresponding integer.
         Default is 'WARNING'.
-    backupCount : int
-        Number of historical log files to keep. Default is 4. Use 0 to
-        keep all historical log files. The argument name is chosen to match
-        TimedRotatingFileHandler's backupCount.
     Returns
     -------
     handler : logging.Handler
@@ -227,11 +219,7 @@ def config_ophyd_logging(
     """
     global current_handler
     if isinstance(file, str):
-        handler = logging.handlers.TimedRotatingFileHandler(
-            filename=file,
-            when="W0",  # roll over every Monday
-            backupCount=backupCount,  # keep the most recent "backupCount" log files
-        )
+        handler = logging.FileHandler(file)
     else:
         handler = logging.StreamHandler(file)
     levelno = validate_level(level)
--- a/ophyd/log.py
+++ b/ophyd/log.py
@@ -132,7 +132,7 @@ color_log_format = ("%(color)s[%(levelname)1.1s %(asctime)s.%(msecs)03d "
 
 def validate_level(level) -> int:
     """
-    Return aN int for level comparison
+    Return an int for level comparison
     """
     if isinstance(level, int):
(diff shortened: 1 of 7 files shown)


danielballan:
By the way, :+1: for using `TimedRotatingFileHandler` as our default file handler. I would support moving that into caproto as well.

ophyd/log.py
```diff
@@ -128,20 +127,34 @@ def format(self, record):
...
+    """
+    if isinstance(level, int):
+        levelno = level
+    elif isinstance(level, str):
+        levelno = logging.getLevelName(level)
+
+    if isinstance(levelno, int):
+        return levelno
+    else:
+        raise ValueError("Your level is illegal, please use 'ERROR', 'WARNING', 'INFO', 'DEBUG', OR 'TRACE'.")
```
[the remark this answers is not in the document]
jklynch: Ah, TRACE. I knew it well.

TRACE is no more.

danielballan:
- [ ] Make `self.log` on `OphydObj` a `LoggerAdapter` that refers to one global logger and wraps up the object name, rather than separate `Logger` instances for each object name. https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/ophydobj.py#L184
- [ ] Log `set_and_wait` timeout at higher than DEBUG level. Maybe ERROR or exception. https://github.com/bluesky/ophyd/pull/785/files#diff-0aa5a9ffa6e34986979d3e945c34f14cR236
- [ ] Log at INFO level next to all calls to `self._write_pv.put`, `.get`, and `.get_with_metadata`. Same for `self._read_pv.get` and `.get_with_metadata`.
- [ ] Inside the value property, distinguish between returning the cached value `self._readback` and a fresh value from calling `self.get()`.
- [ ] Log inside `StatusBase._finished()` when an action completes. Do not log when a status object is returned; this is not really actionable/impactful. https://github.com/bluesky/ophyd/blob/3cbf73b8f71b9c2494ffb097ac1a5e9b5c0f1e23/ophyd/status.py#L119

ophyd/signal.py
```diff
@@ -158,6 +156,7 @@ def _repr_info(self):
 
     def get(self, **kwargs):
         '''The readback value'''
+        logger.info('%s get() returning ', self, self._readback)
```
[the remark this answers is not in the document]
danielballan: On a phone call, @jklynch and I realized that OphydObj is creating a separate logger for every instance of OphydObj, a pattern we have since learned to avoid, in favor of using `LoggerAdapter` to wrap _one_ `Logger` with some additional context (in this case, `self.name`). This can kind of thing could be changed to

```py
self.log('get() returning %s', self._readback)
```

where `self.log` is a `LoggerAdapter` that wraps up the relevant context (`self.name`, maybe other stuff too...). Downstream, a `Formatter` can incorporate that extra context in the final log message.

ophyd/log.py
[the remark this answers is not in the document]
mrakitin: Maybe it's worth to get all available level names programmatically?
```py
In [8]: logging._levelToName.values()
Out[8]: dict_values(['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG', 'NOTSET'])
```

ophyd/signal.py
mrakitin:
```suggestion
        logger.info('%s get() returning %s', self, self._readback)
```

ophyd/signal.py
```diff
@@ -214,6 +213,7 @@ def put(self, value, *, timestamp=None, force=False, metadata=None,
         if 'timestamp' not in self._metadata_keys:
             md_for_callback['timestamp'] = timestamp
 
+        logger.info('%s put(value=%s, timestamp=%s, force=%s, metadata=%s)', self, value, timestamp, force, metadata)
```
mrakitin: `self.name` can be used here as well, as @danielballan pointed out above.

ophyd/signal.py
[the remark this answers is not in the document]
danielballan: ...To be clear, all of the above was just to illustrate how to thread `self.name` into log messages. In this _particular_ case (logging on every `get()`) I think we should probably strip this out and rely on messages right next to where the I/O occurs... `self._read_pv.get` and similar.

mrakitin:
Many Linux programs having rotating logs have 10 last logs saved. I think I'd make sense to keep 10 last logs for ophyd as well, but it should be configurable.

Also, I don't see where `TimedRotatingFileHandler` is used. Was everything committed?

ophyd/signal.py
danielballan: Fair, but I think we'll just remove this logging call. Too much noise.

danielballan:
@jklynch [This comment](https://github.com/bluesky/ophyd/pull/788#issuecomment-572734042) notes some issues that are relevant to the changes here. Might as well address them while we are at it.

jklynch:
I found and fixed the problem referenced in #788 as line 188 of ophydobj.py. The similar one referenced as line 682 of signal.py is not present in the version I'm working with. I did find and fix another error of that type at line 792.

ophyd/signal.py
```diff
@@ -839,8 +853,12 @@ def get(self, *, as_string=None, connection_timeout=1.0, form='time',
             as_string = self._string
 
         self.wait_for_connection(timeout=connection_timeout)
-        info = self._read_pv.get_with_metadata(as_string=as_string, form=form,
-                                               **kwargs)
+        self.log.info(
```
danielballan: There are two categories of log messages added in this PR:
1. Things that ophyd is doing
2. Things that ophyd is asking the control layer to do.

The log messages in category (2), such as this one, might better be done upstream in the control layers. (Caproto already goes it; pyepics could be made to do it.) _But_ since the immediate motivation for this additional logging is to debug a tricky pseudopositioners issue that may involve pyepics, including older versions of pyepics, I think we _should_ add these log messages, provisionally. In the longer term, we may submit a PR to pyepics and then strip these out after a couple of pyepics releases.

I propose issuing all these kinds of log messages from a separate logger named something like `'ophyd_control_layer'` not under the `ophyd.` logger namespace. That will serve two purposes: it will keep the noise down for users because the control layer-related messages will have to be separately turned up, and it will make it easier to specifically remove just these log messages in the future.

ophyd/log.py
```diff
@@ -24,8 +24,7 @@ def _stderr_supports_color():
                 if curses.tigetnum("colors") > 0:
                     return True
             elif colorama:
-                if sys.stderr is getattr(colorama.initialise, 'wrapped_stderr',
-                                         object()):
+                if sys.stderr is getattr(colorama.initialise, 'wrapped_stderr', object()):
```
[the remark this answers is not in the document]
danielballan: re-upping this

ophyd/status.py
```diff
@@ -133,6 +142,7 @@ def _finished(self, success=True, **kwargs):
         success : bool, optional
            if the action succeeded.
         """
+        self.log.info('finished')
```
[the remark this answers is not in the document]
jklynch: oops

ophyd/ophydobj.py
```diff
@@ -102,6 +99,14 @@ def register(instance):
     return weak_set
 
 
+class OphydObjectLoggerAdapter(LoggerAdapter):
+    """
+    A LoggerAdapter for use by OphydObject.
+    """
+    def process(self, msg, kwargs):
```
danielballan: [I neglected to comment on this in the last review, sorry.]

I believe that [`Formatters.format`](https://github.com/python/cpython/blob/9af0e47b1705457bb6b327c197f2ec5737a1d8f6/Lib/logging/__init__.py#L651-L681) is the intended place to do this kind of string formatting. I imagine that the rationale is that richer consumers, such as Elastic which can consume and index structured data, might not want or need the structure to be flattened into one message string, so that work should only be done for the handlers that require flat strings.

See [example from previous work](https://github.com/caproto/caproto/blob/334679b45fef52e998a4cbb398d95b5708b82c5b/caproto/_log.py#L107-L117).

I have no problem punting `Formatter` work to a separate PR, but I don't think we should grow a `LoggerAdapter` subclass in the mean time for the sake of doing work that fits well in a `Formatter`.

ophyd/signal.py
```diff
@@ -267,9 +281,10 @@ def set_thread():
     def value(self):
         '''The signal's value'''
         if self._readback is not None:
-            return self._readback
-
-        return self.get()
+            val = self._readback
```
danielballan: :+1: for making it easier to add debugging prints here when needed

ophyd/status.py
```diff
@@ -26,6 +25,14 @@ def f(self, *args, **kwargs):
     return f
 
 
+class StatusBaseLoggerAdapter(LoggerAdapter):
+    """
+    A LoggerAdapter for use by StatusBase and derived classes.
+    """
+    def process(self, msg, kwargs):
+        return f"[{str(self.extra['status'])}] {msg}", kwargs
```
danielballan: Same comments as above regarding using `Formatter` instead of a custom subclass.

ophyd/log.py
```diff
@@ -150,44 +194,70 @@ def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
...
-    handler.setFormatter(
-        LogFormatter(format, datefmt=datefmt))
+        log_format = plain_log_format
+    handler.setFormatter(LogFormatter(log_format, datefmt=datefmt))
+
     if current_handler in logger.handlers:
         logger.removeHandler(current_handler)
     logger.addHandler(handler)
+
+    if current_handler in control_layer_logger.handlers:
```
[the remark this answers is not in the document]
jklynch: I would like to use the same handler for all ophyd-related logging and by default the same logging level, presuming the control_layer_logger is a temporary measure. I would prefer to leave it to developers to turn it down with setLevel().

ophyd/log.py
```diff
@@ -150,44 +194,70 @@ def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
         Date format. Default is ``'%H:%M:%S'``.
     color : boolean
         Use ANSI color codes. True by default.
-
+    level : str or int
+        Python logging level, given as string or corresponding integer.
+        Default is 'INFO'.
+    backupCount : int
```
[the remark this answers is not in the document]
jklynch: Maksim mentioned configuring this particular value in one of his comments. It is one of the defining features of the TimedRotatingFileHandler so I think it makes sense.

ophyd/log.py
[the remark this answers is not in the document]
danielballan: That is OK with me. In that case, I don't see any reason to give it a name outside the `ophyd.*` logger namespace. Confining ourselves to `ophyd.*` seems good behavior so that someone wishing to _silence_ our logs only has one top-level logger to touch. I only suggested going outside `ophyd.*` in order to side-step this mechanism, so if we decide not to side-step it let's call the logger `ophyd.control_layer`. It can of course have its level separately set, as you suggest. Also, I think all control layer messages should be DEBUG or maybe a custom unnamed level lower than debug (i.e. 0 < level < 10), up to you.

ophyd/log.py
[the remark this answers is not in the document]
danielballan: Good enough for me.

ophyd/log.py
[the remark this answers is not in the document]
jklynch: I agree that control layer messages should be DEBUGs.

ophyd/log.py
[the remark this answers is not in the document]
jklynch: There could be a whole new mechanism to provide a handler to `configure_ophyd_logging` but in the interest of keeping this PR smaller I punted.

jklynch:
CI is timing out on this unit test:
```
ophyd/tests/test_pseudopos.py::test_multi_sequential[pyepics] 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```
The function `test_multi_sequential()` is the first in `test_pseudopos.py` to call `wait_for_connection()`.

jklynch:
Starting `caproto-spoof-beamline` and running `test_pseudopos.py` seems to reproduce the behavior in CI - the test function `test_multi_sequential()` hangs. Here is the test hanging:
```
(venv) jlynch@thorisdottir:~/local/ophyd-logging$ pytest ophyd/tests/test_pseudopos.py 
==================================================================================================================================================== test session starts =====================================================================================================================================================
platform linux -- Python 3.6.9, pytest-5.3.2, py-1.8.0, pluggy-0.13.1
rootdir: /home/jlynch/local/ophyd-logging, inifile: pytest.ini
collected 130 items                                                                                                                                                                                                                                                                                                          

ophyd/tests/test_pseudopos.py ..
```
spoof-beamline gets made about something as well:
```
(venv) jlynch@thorisdottir:/mnt/hgfs/project/caproto$ caproto-spoof-beamline 

*** WARNING ***
This script spawns an EPICS IOC which responds to ALL caget, caput, camonitor
requests.  As this is effectively a PV black hole, it may affect the
performance and functionality of other IOCs on your network.

The script ignores the --interfaces command line argument, always
binding only to 127.0.0.1, superseding the usual default (0.0.0.0) and any
user-provided value.
*** WARNING ***

Press return if you have acknowledged the above, or Ctrl-C to quit.



                         PV blackhole started


Namespace(async_lib='asyncio', connection_delay=0.0, interfaces=['0.0.0.0'], list_pvs=False, prefix='', quiet=False, verbose=None)
{'connection_delay': 0.0}
[I 08:47:02.597          server:  165] Asyncio server starting up...
[I 08:47:02.597          server:  178] Listening on 127.0.0.1:5064
[I 08:47:02.598          server:  264] Server startup complete.
[I 08:47:12.464          common:  968] Connected to new client at 127.0.0.1:49290 (total: 1).
[E 08:47:38.081          server:  202] BcastLoop received error
      File "/usr/lib/python3.6/asyncio/selector_events.py", line 1091, in sendto
        self._sock.send(data)
    OSError: [Errno 22] Invalid argument
[E 08:47:53.088          server:  202] BcastLoop received error
      File "/usr/lib/python3.6/asyncio/selector_events.py", line 1091, in sendto
        self._sock.send(data)
    OSError: [Errno 22] Invalid argument
[E 08:48:08.105          server:  202] BcastLoop received error
      File "/usr/lib/python3.6/asyncio/selector_events.py", line 1091, in sendto
        self._sock.send(data)
    OSError: [Errno 22] Invalid argument
[E 08:48:23.110          server:  202] BcastLoop received error
      File "/usr/lib/python3.6/asyncio/selector_events.py", line 1091, in sendto
        self._sock.send(data)
    OSError: [Errno 22] Invalid argument
...
```

danielballan:
Oh, excellent. Maybe we have stumbled into a way to reproduce this BMM issue is a more isolated context. This problem looks similar.

The caproto errors might be unrelated. Can you try a different async-lib when you start the server? See the -h help string for details on that.

ophyd/log.py
```diff
@@ -113,34 +128,63 @@ def format(self, record):
...
+        levelno = level
+    elif isinstance(level, str):
+        levelno = logging.getLevelName(level)
+
+    if isinstance(levelno, int):
+        return levelno
+    else:
+        raise ValueError(
+            "Your level is illegal, please use "
+            "'ERROR', 'WARNING', 'INFO', 'DEBUG', OR 'TRACE'."
```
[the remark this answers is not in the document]
danielballan: Bump.

ophyd/log.py
```diff
@@ -100,47 +115,85 @@ def __init__(self, fmt=DEFAULT_FORMAT, datefmt=DEFAULT_DATE_FORMAT,
...
+logger = logging.getLogger("ophyd")
+control_layer_logger = logging.getLogger("ophyd_control_layer")
 
 
 current_handler = None  # overwritten below
 
 
-def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
+def config_ophyd_logging(
+    file=sys.stdout, datefmt="%H:%M:%S", color=True, level="INFO", backupCount=4
```
danielballan: The default should be `"WARNING"`. That's Python's default level in the logging module, INFO is very verbose (as I just found trying to uses this branch), and it's consistent with https://github.com/caproto/caproto/blob/1a7356487fd930cb6dee6c26f1ce1f117cd87c95/caproto/_log.py#L353.

ophyd/log.py
```diff
@@ -150,44 +203,70 @@ def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
...
+
+    if control_layer_logger.getEffectiveLevel() > levelno:
+        control_layer_logger.setLevel(levelno)
+
     return handler
 
 
 # Add a handler with the default parameters at import time.
-current_handler = set_handler()
+config_ophyd_logging()
```
danielballan: We used to think it was convenient to set up a default handler at import time, but we have since learned differently. The Python logging documentation advises strongly against libraries adding handlers at import time. In practice it seems tog get in the way as often as it helps; it's better to just provide and advertise a convenience function for users to call to add special logging, and then set it up to suit at the application level.

For example, at NSLS-II, we plan to call set up handlers in the profiles, ensuring that log messages are always written to some standard location.

danielballan: (Note: this section was copied out of caproto, which used to set up a handler at import time but now correctly leaves it to the application to do.)

ophyd/log.py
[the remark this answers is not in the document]
danielballan: Bump on this too.

ophyd/log.py
prjemian: Two (at least) trains of thought on where to write the logs.

1. well-known location
1. user's working directory

Guiding the choice is the answer to *Who needs this logging?*  Initially, the developer who is expects a predictable location.  Longer term, the user needs a record of what they have typed and in what sequence.  Writing the logs to the user directory means they can copy that directory (which may contain other content such as *.py files with their custom plans).

If the developer can tolerate logs going into the user directory, that should be the default.  **But**, the location should be configurable.

Another configurable should be 

* un/limited number of logs
* un/limited length of one log file

danielballan: Agreed. Bluesky will not make any choice by default, leaving it up to the application (or the user's IPython profile) but should probably provide some recommendations. At a recent group meeting, we discussed that `$XDG_DATA_HOME/bluesky/logs` (i.e. `~/.local/share/bluesky/logs`) is probably the "proper" place. Unfortunately, since we at NSLS-II have very slow shared home directory performance, we will probably use a world-writable system location, maybe under `/var/log`, but only until our storage situation is fixed. As for limited log file length and number, we will probably rely on and recommend https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler.

danielballan: Another possibility is configuring _two_ handlers, one for the user's local directory and one into a system used by the facility support team, which might be a system location or a network-accessible log capture service like Elastic or Graylog.

jklynch: I agree writing the logs to the user's home directory is very reasonable.

The question of further configuration such as number of logs and length of one log files runs into another thing I have been thinking about which is combined configuration of bluesky and ophyd logging (and maybe others). Maybe we should supply a handler to the `config_ophyd_logging` method? Or have a `add_ophyd_logging_handler`?

danielballan: At that point, I think users should just use the standard Python logging API. Our config_X_logging methods are just shortcuts to help non-developers turn on some extra logging to the terminal or a file as a one-liner without knowing the details. For facility configurations, we can assume more background knowledge.

prjemian: This discussion motivates me to make changes in [stdlogpj](https://github.com/prjemian/stdlogpj). 

> we discussed that `$XDG_DATA_HOME/bluesky/logs` (i.e. `~/.local/share/bluesky/logs`) is probably the "proper" place.  

This sounds like a facility-dependency, not general for any facility.

danielballan: > This sounds like a facility-dependency, not general for any facility.

Yes. I believe this discussion has no bearing on what should go in the _library_ code. Facilities can make their own decisions and implement them using `import logging; logging.getLogger('ophyd').addHandler(...)` without any special code in ophyd itself. At most, we could decide to document some recommendations or examples to help folks navigate the Python logging framework, [as we have done in the caproto docs](https://caproto.github.io/caproto/loggers.html).

ophyd/log.py
jklynch: I agree.

danielballan:
CI failure is unrelated to this PR:

<details>

```
>           fh = pytest.importorskip('databroker.assets.handlers')
ophyd/tests/test_areadetector.py:417: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/databroker/__init__.py:2: in <module>
    import intake
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/intake/__init__.py:26: in <module>
    registry.update(autodiscover())
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/intake/source/discovery.py:143: in autodiscover
    drivers[entrypoint.name] = _load_entrypoint(entrypoint)
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/intake/source/discovery.py:248: in _load_entrypoint
    return entrypoint.load()
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/entrypoints.py:82: in load
    mod = import_module(self.module_name)
../../../virtualenv/python3.7.1/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/databroker/core.py:1923: in <module>
    from databroker._core import Header
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/databroker/_core.py:6: in <module>
    from pims import FramesSequence, Frame
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/pims/__init__.py:1: in <module>
    from pims.api import *
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/pims/api.py:21: in <module>
    from pims.tiff_stack import TiffStack_tifffile  # noqa
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/pims/tiff_stack.py:23: in <module>
    import tifffile
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/tifffile/__init__.py:3: in <module>
    from .tifffile import __doc__, __all__, __version__, main
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/tifffile/tifffile.py:10643: in <module>
    bitorder_decode = imagecodecs.bitorder_decode  # noqa
../../../virtualenv/python3.7.1/lib/python3.7/site-packages/imagecodecs/imagecodecs.py:374: in __getattr__
    module = importlib.import_module('._' + module_, 'imagecodecs')
../../../virtualenv/python3.7.1/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
```

</details>

PR #791, opened 2020-01-30
Resolve #773 again.

### LINKED ISSUE #773: Backport a fix for pyepics context kwarg warning

danielballan:
This is a second pass at https://github.com/bluesky/ophyd/pull/774 (which was reverted) addressing the review comments that didn't get addressed before the merge of #774.

### DIFF
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -1,6 +1,7 @@
 import atexit
 import ctypes
 import epics
+from distutils.version import LooseVersion
 import queue
 import threading
 import warnings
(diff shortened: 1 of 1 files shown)


danielballan:
The important difference from #774 is that we  _thought_ we had no choice but to drop 3.5 support on a microrelease, which we were prepared to do reluctantly, but in the end we found a solution that lets us retain 3.5 support.

PR #788, opened 2020-01-06
Make timeouts more configurable.

danielballan:
Per @prjemian, there is a decades-long bug in the EPICS servers that causes IOCs to sometimes ignore read requests, even though the specification requires them to respond. Reportedly, SPEC works around this by simply retrying. This is a little scary, because if the IOC is suddenly not abiding by the specification one could worry that it could be giving wrong data.

We have observed this intermittently at NSLS, typically in the context of Area Detector and at least once with a Scaler. At APS (particulartly 8ID) they have observed it significantly more. Anecdotally, this server issue seems to be easier to hit with hard IOCs than soft IOCs, which may explain why it is more common at APS than NSLS.

As a compromise, ophyd will allow users to opt in to retries. Default settings can be applied at the class level, via a new classmethod on `EpicsSignalBase`. Individual signals may be separately configured at instantiation time.

Basic demo and relevant docstring:

```py
In [1]: import ophyd                                                                                                                                                                          

In [2]: ophyd.EpicsSignal.set_default_timeout(timeout=60, read_retries=5, floor=10)                                                                                                           

In [3]: signal = ophyd.signal.EpicsSignal('simple:A')                                                                                                                                         

In [4]: signal.get()                                                                                                                                                                          
Out[4]: 1

In [5]: signal.get_setpoint()                                                                                                                                                                 
Out[5]: 1

In [6]: ophyd.EpicsSignal.set_default_timeout?                                                                                                                                                
Signature:
ophyd.EpicsSignal.set_default_timeout(
    *,
    timeout=2.0,
    connection_timeout=1.0,
    read_retries=0,
    floor=0.5,
)
Docstring:
Set the class-wide defaults for timeouts and read retries.

This may only be called before any instances of EpicsSignalBase are
made.

Parameters
----------
timeout: float, optional
    Total time budget (seconds) for connecting and reading, including
    all retries.
connection_timeout: float, optional
    Time (seconds) allocated for establishing a connection with the
    IOC.
read_retries: integer, optional
    Number of retries. The default value is 0, meaning one total
    read attempt will be made.
floor: float, optional
    Minimum time per read attempt. Each read attempt will be given
    ``max(timeout / read_retries, floor)`` seconds to succeed or
    timeout.

Raises
------
RuntimeError
    If called after :class:`EpicsSignalBase` has been instantiated for
    the first time.
File:      ~/Repos/bnl/ophyd/ophyd/signal.py
Type:      method
```

[...]

PR #792, opened 2020-01-31
HACK: set read timeout unconditionally to 10s

tacaswell:
In #788 we added the ability to control the default timeout (and
per-signal PV timeout).  For unclear reasons some IOCs can take an
unexpectedly long time to service read requests and this knob gives us
the ability to accept that delay rather than failing the `get` which
returns `None` which tends to cause issues with down stream code.

That enhancement was built on top of a number of other
changes (including the machinery to know if any Signals have been
instantiated yet and an overhaul of EpicsSignal internals) which makes
it impossible to directly back-port that change to the 1.3.x branch.

Instead, we hard-code the default timeout to 10s.  This will increase
how long it will take for us to fail if the server truly has become
non-responsive, but will greatly reduce the number of false-positive
failures.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -16,6 +16,8 @@ from . import get_cl
 
 logger = logging.getLogger(__name__)
 
+_DEFAULT_TIMEOUT = 10  # s
+
 
 class Signal(OphydObject):
     '''A signal, which can have a read-write or read-only value.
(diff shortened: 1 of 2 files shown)


danielballan:
@tacaswell 1.3.3 was the last release. Moving to 1.3.4, which has an rc0 out but has not been released yet.

prjemian:
I'm OK with the principle here, that a default timeout of 10 s is being applied to old versions where modifying the code base (for the old release series) would be problematic.  For the future, user-configurable timeout is desirable.

PR #793, opened 2020-02-07
Some Minor Fixes

hhslepicka:
FIX: Run callbacks for Signal at init.
- For Typhos we need to receive information that the Signal was connected in order to properly render the correct widget. With `EPICS Signal` all works great because we can get the metadata callback. For Signal it is not true. This PR fixes this issue.

FIX: Don't blow if SynSignal was not created with asyncio loop.
- We are instantiating some test devices in a separated thread and this blows when we try with SynSignals. This commit avoids it to break when creating the signal outside of an asyncio loop.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -127,7 +127,10 @@ class SynSignal(Signal):
             # Initialize readback with a None value
             self._readback = None
         if loop is None:
-            loop = asyncio.get_event_loop()
+            try:
+                loop = asyncio.get_event_loop()
+            except RuntimeError:
+                loop = None
         self._func = func
         self.exposure_time = exposure_time
         self.precision = precision
@@ -260,6 +263,7 @@ class _ReadbackSignal(Signal):
             connected=True,
             write_access=False,
         )
+        self._run_metadata_callbacks()
 
     def get(self):
         return self.parent.sim_state['readback']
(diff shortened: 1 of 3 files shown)


hhslepicka:
`test_soft_derived` is failing due to the fact that now `Signal` on creation runs the metadata callback which makes it run more than just once. I will adjust the test now to reflect the change.

ophyd/sim.py
```diff
@@ -127,7 +127,10 @@ def __init__(self, func=None, *,
             # Initialize readback with a None value
             self._readback = None
         if loop is None:
-            loop = asyncio.get_event_loop()
+            try:
+                loop = asyncio.get_event_loop()
```
tacaswell: under what conditions does this fail?

hhslepicka: If you try to instantiate it inside of a QThread or regular python Thread.

klauer:
There's a bit of a problem here that arises from us not having a `SignalBase` and `Signal` (what really should be called a `SoftSignal` of sorts) separately.

With this PR, `Signal` will have a callback indicating `connected=True` before a subclass of `Signal` can indicate `connected=False` (or set read status, write status, etc. appropriately). This callback also gets queued in a background metadata callback thread, meaning that it's entirely possible a client that registers to receive the callback after the message was queued could receive the 'connected' message. Can explain this better in person and with less jet lag on monday.

Perhaps this fix is sufficient for now. Where it goes from there is another question...

hhslepicka:
@klauer that makes total sense. I remember now that we discussed that. Maybe it will be time to tackle that soon. Let's chat Monday and put this on hold for now. It would help also if I explain why I need that and it will for sure be easier in person. Thank you for reviewing it.

hhslepicka:
Tests now failing on some documentation build issue. Not related to PR.

klauer:
Let's try and get a proper fix in, even if not for 1.4.0. This definitely needs more thought.

Issue #794, opened 2020-02-07
MoveStatus watch method error


AbbyGi:
In the MoveStatus _notify_watchers method, if the fraction calculation causes an exception, fraction is set to None and the next if statement causes an error if None is passed into np.isnan.
```py
import numpy as np
a = None
print(np.isnan(a))
```
```py
ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

tacaswell:
Closed by https://github.com/bluesky/ophyd/pull/795

PR #795, opened 2020-02-07
TST: add a test to demonstrate MoveStatus._notify_watchers failure

mrakitin:
See #794 for the context.

### DIFF
--- a/doc/source/architecture.rst
+++ b/doc/source/architecture.rst
@@ -56,15 +56,15 @@ associated with it:
     makes sense to the people directly working with the ophyd instances.
     They need be unique within a `~ophyd.Device` and hence Python ensures that
     the fully qualified name will be unique within a namespace.
-3. *The ``obj.name`` attribute*.  This name is the one that will be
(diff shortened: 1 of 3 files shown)


mrakitin:
The problem is reproduced via the test in this PR: https://travis-ci.com/mrakitin/ophyd/jobs/285017761#L3707.

Should it be merged first, and then a separate fix would solve it? Or is it better to do in one shot?

prjemian:
One Clean Shot

mrakitin:
Yeah, thanks for the suggestion, @prjemian. I thought it would make sense too, but having a future reference to the problem itself would be good too. Anyway, it's documented via #794.

klauer:
👍 after fixing the `watch` callback signature issue.

mrakitin:
41bba72 fixes the issue with formatting introduced in https://github.com/bluesky/ophyd/pull/790.

mrakitin:
Thanks, @klauer!

PR #797, opened 2020-02-08
dotted_name property should include the `obj`

### LINKED ISSUE #796: dotted_name() should include name of the obj
Something's missing in the output of the `obj.dotted_name` property -- the name of the `obj` itself.  For example, given an instance of EpicsMotor:

```
In [11]: import ophyd                                                                                                                       

In [12]: ophyd.__version__                                                                                                                  
Out[12]: '1.4.0rc4'

In [13]: m1.low_limit_switch.dotted_name                                                                                                    
Out[13]: 'low_limit_switch'
```

Expected to receive `'m1.low_limit_switch'`

prjemian:
fixes #796

Here's what the fix looks like in practice:

```
In [14]: dotted_name??                                                                                                                      
Signature: dotted_name(obj)
Source:   
def dotted_name(obj):
    """Return the dotted name"""
    names = []
    while obj.parent is not None:
        names.append(obj.attr_name)
        obj = obj.parent
    names.append(obj.name)
    return '.'.join(names[::-1])
File:      ~/.ipython/user/mining.py
Type:      function

In [15]: dotted_name(m1.low_limit_switch)                                                                                                   
Out[15]: 'm1.low_limit_switch'
```

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -304,6 +304,7 @@ class OphydObject:
         while obj.parent is not None:
             names.append(obj.attr_name)
             obj = obj.parent
+        names.append(obj.name)
         return '.'.join(names[::-1])
 
     @property


tacaswell:
We intentionally did not include the name of the root object in the dotted name.  

If you do, then things like `getattr(obj.parent, obj.dotted_name) is obj` don't round trip.  

Another issue is that the rest of the names are the chain are `attr_names` which are extracted from the classes and python ensures are unique and immutable for us (should check that, but they _should_ be thought of as immutable).  On the other hand the `parent.name` attribute is mutable by the user at run time and there is no guarantee that the name reported by `obj.name` is the name the object has in the enclosing namespace.

While the default `name` of a given sub-object is `self.parent.name + '_' + self.attr_name`, there should be a clear separation between the "names python has a say over because the language needs to work" and the "names the scientists get a say over because the ophyd Device structure is an implementation detail of the collection system that should not leak into analysis".  It makes like mildly easier for those of us in the middle if those two sets of names _roughly_ match, but we should not bake that convenience into the code anyplace.

PR #803, opened 2020-02-18
Refactor debug monitor to use logging not print.

danielballan:


danielballan:
Developed on a phone call with @prjemian, @klauer, @hhslepicka, @jklynch in December. Just remembered to revive and push this branch.

danielballan:
Power-cycled for CI

PR #804, opened 2020-02-20
Support a Signal write_timeout pass to put.

danielballan:
This makes the timeout for a *write* configurable on a per-put and
per-Signal basis. It uses a similar pattern to what we did for read and
connection timeout in #788.

At SST, we were hitting a hard-coded timeout of 30 seconds (somewhere
in caproto or ophyd, I now don't recall...) and the motion of their mono
can sometimes take longer. With these changes, we could set
``write_timeout=180.`` in anticipation of lengthy motions.

### DIFF
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -72,9 +72,12 @@ class PyepicsShimPV(epics.PV):
                                     run_now=run_now,
                                     with_ctrlvars=with_ctrlvars, **kw)
 
-    def put(self, value, wait=False, timeout=30.0, use_complete=False,
+    def put(self, value, wait=False, timeout=None, use_complete=False,
             callback=None, callback_data=None):
         callback = wrap_callback(_dispatcher, 'get_put', callback)
+        # pyepics does not accept an indefinite timeout
+        if timeout is None:
+            timeout = 315569520  # ten years
         return super().put(value, wait=wait, timeout=timeout,
                            use_complete=use_complete, callback=callback,
                            callback_data=callback_data)
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -596,12 +596,47 @@ class EpicsSignalBase(Signal):
     ----------
     read_pv : str
         The PV to read from
+    string : bool, optional
+        Attempt to cast the EPICS PV value to a string by default
     auto_monitor : bool, optional
         Use automonitor with epics.PV
     name : str, optional
         Name of signal.  If not given defaults to read_pv
-    string : bool, optional
-        Attempt to cast the EPICS PV value to a string by default
+    metadata : dict
+        Merged with metadata received from EPICS
+    all_pvs : set
+        Set of PVs to watch for connection and access rights callbacks.
+        Defaults to ``{read_pvs}``.
+    timeout : float or None, optional
+        The timeout for serving a read request on a connected channel. This is
+        only applied if the PV is connected within connection_timeout (below).
+
+        The default value DEFAULT_TIMEOUT means, "Fall back to class-wide
+        default." See EpicsSignalBase.set_default_timeout to configure class
+        defaults.
+
+        Explicitly passing None means, "Wait forever."
+    write_timeout : float or None, optional
+        The timeout for a reply when put completion is used. This is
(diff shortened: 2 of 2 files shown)


ZLLentz:
I'm confused about the use case here. Shouldn't all writes happen quickly in absence of network issues? I expect that PVs related to slow moving motors would successfully write quickly and then a different PV would be used to verify completion.

ophyd/signal.py
```diff
@@ -648,6 +650,16 @@ def __init__(self, read_pv, *, string=False, auto_monitor=False, name=None,
...
             timeout = self.__default_timeout
         self._timeout = timeout
+        if write_timeout is DEFAULT_WRITE_TIMEOUT:
+            # This is very different than the connection and read timeouts
+            # above. It relates to how long an action takes to complete. Any
+            # default value we choose here is likely to cause problems---either
+            # by being too short and giving up too early on a lengthy action or
+            # being too long and delaying the report of a failure.
+            # The important thing it is that it is configurable at per-Signal
+            # level via the write_timeout parameter.
```
ZLLentz: Should note here or in a docstring that this write timeout is only invoked if the pv is already connected. A disconnected PV will be subject to the connection timeout instead. Even with the lengthy comment I don't think this is clear, before looking under the hood I was confused.

tacaswell:
I think this only matters for PVs that block on put-complete significantly.

The other place I bet we would need this is on newer versions of AD when we use "wait_for_plugins" mode.

ophyd/signal.py
danielballan: Good note. The other timeouts needed detailed docstrings as well. Pushed.

tacaswell:
To be fair, we have found this being a problem in only 1 place (and the fix to PVPositioner in #805 would have avoided this all together).

danielballan:
Power-cycled for CI.

danielballan:
> CI is failing miserably - but is it related to the PR?

There was an unrelated issue with a third-party dependency that broke all of our repos, and that's now been fixed. But there was a real issue here as well: pyepics does not accept `timeout=None` for `put`.

```py
In [9]: pv.put('000000', wait=True, use_complete=True, timeout=None)                                                                                                                          
TypeError                                 Traceback (most recent call last)
<ipython-input-9-238d6ef3486e> in <module>
----> 1 pv.put('000000', wait=True, use_complete=True, timeout=None)

<snipped>

~/venv/bnl/lib/python3.6/site-packages/epics/ca.py in put(chid, value, wait, timeout, callback, callback_data)
   1730     if wait:
   1731         while not (completed['status'] or
-> 1732                    (time.time()-start_time) > timeout):
   1733             poll()
   1734         if not completed['status']:

TypeError: '>' not supported between instances of 'float' and 'NoneType'
```

I have added a commit patching that in `_pyepics_compat.py`, injecting a 10-year timeout in place of `None`. It's the best we can do, in the spirit of

>  Ophyd should take control of timeouts.

ophyd/_pyepics_shim.py
```diff
@@ -72,9 +72,12 @@ def add_callback(self, callback=None, index=None, run_now=False,
                                     run_now=run_now,
                                     with_ctrlvars=with_ctrlvars, **kw)
 
-    def put(self, value, wait=False, timeout=30.0, use_complete=False,
+    def put(self, value, wait=False, timeout=None, use_complete=False,
```
danielballan: This change is not really significant, because higher layers always pass down an explicit value now.

ophyd/signal.py
```diff
@@ -726,6 +773,9 @@ def set_default_timeout(cls, *, timeout=2.0, connection_timeout=1.0):
         cls.__default_connection_timeout = connection_timeout
         cls.__default_timeout = timeout
 
+    # TODO Is there a good reason to prohibit setting these three timeout
```
klauer: Is there a good reason to allow the user to change them on the fly? 🤷‍♂

danielballan: Yeah, it might be convenient to be able to try different write timeouts, but if (per the docstring) the recommended approach is to configure this inside a `set` or `trigger`, where we have the most information about how long an action should take, then this is a moot point.

PR #805, opened 2020-02-20
PVPositioner blocks if misused with a PV that has PC.

danielballan:
This is a fundamental bug that went unnoticed for awhile. We noticed
because we tried to use a `PVPositioner` in parallel motion with another
positioner and found that the motion was serial.

### DIFF
--- a/ophyd/pv_positioner.py
+++ b/ophyd/pv_positioner.py
@@ -149,10 +149,12 @@ class PVPositioner(Device, PositionerBase):
     def _setup_move(self, position):
         '''Move and do not wait until motion is complete (asynchronous)'''
         self.log.debug('%s.setpoint = %s', self.name, position)
-        self.setpoint.put(position, wait=True)
         if self.actuate is not None:
+            self.setpoint.put(position, wait=True)
             self.log.debug('%s.actuate = %s', self.name, self.actuate_value)
             self.actuate.put(self.actuate_value, wait=False)
+        else:
+            self.setpoint.put(position, wait=False)
 
     def move(self, position, wait=True, timeout=None, moved_cb=None):
         '''Move to a specified position, optionally waiting for motion to


tacaswell:
This only shows up if a) the PVPositioner does not have an actuate button b) the setpoint PV implemenets put-completion.

This suggests that in the actual case on the floor we should be using PVPositonerPC ?  The device we found this on seems to support both put completion _and_ has a done signal so either way should work.

tacaswell:
The failure is numpy / intake / imagecodecs related.

klauer:
I think `PVPositonerPC` should have been used in the case that you mentioned? Unless there are new categories for this that we hadn't considered, I thought both classes covered just about everything...

danielballan:
Yes, it sounds like we are in agreement that `PVPositionerPC` should have been used. Nonetheless I think using the wrong class is a fairly easy-to-hit trap. The consequences of an accidentally-blocking call are severe: no Ctrl+C handling for the duration of the move. Thus I think this change is worth merging, even if though it should never come up in correctly-configured devices.

danielballan:
Updated title to reflect this.

ophyd/pv_positioner.py
```diff
@@ -149,10 +149,12 @@ def moving(self):
     def _setup_move(self, position):
         '''Move and do not wait until motion is complete (asynchronous)'''
         self.log.debug('%s.setpoint = %s', self.name, position)
-        self.setpoint.put(position, wait=True)
         if self.actuate is not None:
+            self.setpoint.put(position, wait=True)
```
klauer: What there is an `actuate` and setpoint has put completion? Seems like a blanket `wait=False` is what you're after.

But then it gets back to the original point of "choose the right class, please".  I'm really not sold on the argument that it's our responsibility to patch it here. 

The idea of keeping `setpoint.put(..., wait=True)` in both instances is that it would block while allowing the IOC to go through whatever bizarre process it needs to update the setpoint or re-initialize other status PVs. Setting `wait=False` could potentially get in the way of what may be a race to check motion status.

danielballan: That is convincing. I will close this and see if `PVPositioner` -> `PVPositionerPC` fixes the issue at the beamline. I may reopen discussion if we find that we don't fully understand the issue, but I think that we do.

PR #809, opened 2020-02-29
ENH: add IndexedDevice and IndexedComponent

klauer:
In the same vein as `DynamicDeviceComponent`, but hopefully much easier to use. Here's a simple example:

```python
from ophyd import Device, IndexedDevice, IndexedComponent, EpicsSignal


class MyDev(Device):
    cpt1 = IndexedComponent(EpicsSignal,
                            attr='channel_{:02d}',
                            suffix='{:02d}:Test',
                            ranges=range(10)
                            )
    cpt2 = IndexedComponent(EpicsSignal,
                            attr='channel_{:02d}{}',
                            suffix='{:02d}:{}:Test',
                            ranges=[range(10), 'ABCD']
                            )
    cpt3 = IndexedComponent(EpicsSignal,
                            attr='channel_{}',
                            suffix='{}:Test',
                            ranges='ABCD',
                            )
```

Such a device would then allow for access of its components by their indices through `__getitem__` syntax:

```
dev = MyDev('PREFIX:', name='dev')

print(dev.cpt1[0])
print(dev.cpt2[9]['A'])
print(dev.cpt3['A'])
```

Would output:
```
EpicsSignal(read_pv='PREFIX:00:Test', name='dev_cpt1_channel_00', parent='dev_cpt1', timestamp=1583013244.530336, auto_monitor=False, string=False, write_pv='PREFIX:00:Test', limits=False, put_complete=False)
EpicsSignal(read_pv='PREFIX:09:A:Test', name='dev_cpt2_channel_09A', parent='dev_cpt2', timestamp=1583013244.546321, auto_monitor=False, string=False, write_pv='PREFIX:09:A:Test', limits=False, put_complete=False)
EpicsSignal(read_pv='PREFIX:A:Test', name='dev_cpt3_channel_A', parent='dev_cpt3', timestamp=1583013244.5474591, auto_monitor=False, string=False, write_pv='PREFIX:A:Test', limits=False, put_complete=False)
```

### DIFF
--- a/ophyd/indexed_device.py
+++ b/ophyd/indexed_device.py
@@ -220,6 +220,26 @@ def _verify_with_formatter(formatter, format_str, num_args):
         )
 
 
+def _verify_slice_type_consistency(ranges):
+    'Verifies that at each level of mapping, there is only one type used'
+    types = tuple(
+        tuple(set(type(value) for value in range_))
+        for range_ in ranges
+    )
+
+    for range_, type_ in zip(ranges, types):
+        if len(type_) == 1:
+            continue
+
+        raise ValueError(
+            f'Found multiple types {type_} in a single range for '
+            f'IndexedComponent, with range={range_}. Types must be consistent '
+            f'for each level.'
+        )
+
+    return tuple(type_[0] for type_ in types)
+
+
 class IndexedComponent(Component):
     '''
     A Device component that dynamically creates a sub Device
@@ -252,6 +272,8 @@ class IndexedComponent(Component):
             for attr, suffix in expand_attr_and_suffix(ranges):
                 attr = Component(SignalClass, _suffix, **keyword_arg_dict)
 
+        Types must not be mixed at each level. For example, (0, 1, 'A') would
+        be invalid whereas ('0', '1', 'A') would be valid.
     clsname : str, optional
         The name of the class to be generated
         This defaults to {parent_name}{this_attribute_name.capitalize()}
@@ -293,8 +315,8 @@ class IndexedComponent(Component):
         self.attr_to_suffix = dict(zip(attrs, suffixes))
         self.mapping_dict = _mapping_dict_from_ranges(ranges, attrs)
         self.attr_to_index = dict(zip(attrs, itertools.product(*ranges)))
-        self.slice_types = tuple(set(type(value) for value in range_)
-                                 for range_ in ranges)
+        # Slice types ends up being: ((int, ), (str, ), ...)
+        self.slice_types = _verify_slice_type_consistency(ranges)
 
         self.attr = attr
         self.base_class = base_class or IndexedDevice
--- a/ophyd/indexed_device.py
+++ b/ophyd/indexed_device.py
@@ -7,27 +7,96 @@ from .device import Device, Component, Kind, create_device_from_components
 from .utils import underscores_to_camel_case
 
 
+def _summarize_slice_type(slc):
+    'Summarize + verify that only 1 type is used in a slice()'
+    slice_type = set((type(slc.start), type(slc.stop),
+                      type(slc.step)))
+    if type(None) in slice_type:
+        slice_type.remove(type(None))
+
+    if len(slice_type) > 1:
+        raise ValueError(
+            f'Unexpected types in slice: {slice_type}'
+        )
+
+    if len(slice_type):
+        slice_type, = list(slice_type)
+        return slice_type
+
+
 class _IndexedChildLevel(collections.abc.Mapping):
     'A mapping helper to allow access of IndexedDevice[idx][idx2...]'
-    def __init__(self, device, mapping_dict):
+    def __init__(self, device, mapping_dict, leaf_depth):
         self._device = device
         self._d = mapping_dict
+        self._leaf_depth = leaf_depth
+
+    def _get_slice(self, slc):
+        if not isinstance(slc, slice):
+            return {slc: self._d[slc]}
+
+        # Check the type of the slice arguments
+        slice_type = _summarize_slice_type(slc)
+
+        match_list = list(self._d)
+        if slice_type is None or issubclass(slice_type, int):
+            # slice everything (, :,) or slice indices
+            ...
+        else:
+            # Rewrite the slice to be in terms of indices
+            slc = slice(
+                match_list.index(slc.start) if slc.start else None,
+                match_list.index(slc.stop) if slc.stop else None,
+                match_list.index(slc.step) if slc.step else None,
+            )
+
+        return {key: self._d[key] for key in match_list[slc]}
 
     def __getitem__(self, item):
-        value = self._d[item]
-        if isinstance(value, str):
-            return getattr(self._device, value)
-        return _IndexedChildLevel(device=self._device, mapping_dict=value)
+        if isinstance(item, slice):
+            item = (item, )
+        if not isinstance(item, tuple):
+            value = self._d[item]
+            if self._leaf_depth == 1:
+                return getattr(self._device, value)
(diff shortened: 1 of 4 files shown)


tacaswell:
Interesting....

klauer:
Should I move this to an LCLS-specific repository or does it have a place here @tacaswell?

tacaswell:
That was a "that looks cool but complicated" interesting not a "I'll put this fire over with the other fire" interesting

ophyd/indexed_device.py
```diff
@@ -0,0 +1,273 @@
...
+            [(0, 'A'), (0, 'B'), (1, 'A'), (1, 'B')]
+
+        The definition of all attributes to be created is based on the above.
+        Assuming component_class is Component, the following pseudo-code
+        shows what would happen on the Device::
+
+            for attr, suffix in expand_attr_and_suffix(ranges):
+                attr = Component(SignalClass, _suffix, **keyword_arg_dict)
+
+    allow_tuple_access : bool, optional
```
tacaswell: Why would you want to dis-allow tuple access?

Might also want to support `dev[0, 'A']`?

tacaswell:
The test failure is due to the allowed failures config getting ignored.

klauer:
Good points. I'll remove the tuple flag and see what I can do about `slice` syntax, which would be very nice to have.

ophyd/indexed_device.py
```diff
@@ -0,0 +1,273 @@
...
+
+def _mapping_dict_from_ranges(ranges, attrs, *, tuple_access=True):
+    'Create a dictionary to be used by _IndexedChildLevel'
+    mapping_dict = {}
+
+    for index, attr in zip(itertools.product(*ranges), attrs):
+        d = mapping_dict
+        for index_part in index[:-1]:
+            if index_part not in d:
+                d[index_part] = {}
```
tacaswell: if allowing tuple access also want to stuff the index remaineder in at each level so `obj.inxd[0][('A', 'B')]` works?

Can we do this by making `__getitem__` smarter rather than pre-computing?

ophyd/indexed_device.py
klauer: Supported now with the slicing syntax

ophyd/indexed_device.py
klauer: I think pre-computing could potentially be less expensive as it happens only once on class definition.

Also, `__getitem__` dramatically increased in complexity in supporting the slice syntax, so unless there's much better reason, I'd prefer to keep it as-is.

klauer:
One remaining bug that I know of in the slice syntax - returning a list when requesting only a single item:

```
In [4]: dev.cpt2[0]['A':'C']
Out[4]:
[FakeEpicsSignal(name='dev_cpt2_channel_00A', parent='dev_cpt2', timestamp=1583195674.057778),
 FakeEpicsSignal(name='dev_cpt2_channel_00B', parent='dev_cpt2', timestamp=1583195674.057919)]

In [5]: dev.cpt2[0]['A']
Out[5]: FakeEpicsSignal(name='dev_cpt2_channel_00A', parent='dev_cpt2', timestamp=1583195674.057778)

In [6]: dev.cpt2[0, 'A']
Out[6]: [FakeEpicsSignal(name='dev_cpt2_channel_00A', parent='dev_cpt2', timestamp=1583195674.057778)]
```

tacaswell:
> `dev.cpt2[0]['A':'C']`

The prior art from pandas when using label based slicing is to use inclusive bounds.

ophyd/indexed_device.py
```diff
@@ -0,0 +1,335 @@
...
+        Defaults to IndexedDevice
+    '''
+
+    # Allow for subclasses to override the formatting verifier:
+    _formatter = string.Formatter()
+
+    def __init__(self, cls, *, attr, suffix, ranges, clsname=None,
+                 doc=None, kind=Kind.normal, default_read_attrs=None,
+                 default_configuration_attrs=None, base_class=None,
+                 component_class=Component, allow_tuple_access=False,
```
tacaswell: `allow_tuple_access` is not used anywhere else.

tacaswell:
I assume a version of this has been used in anger at LCLS?

ophyd/indexed_device.py
danielballan: I too prefer fancy class setup to fancy `__getitem__`.

klauer:
I think the ambiguity of slice indices (imagine channels 1-5 being sliced `ch[1:3]`; should that give `channels 1, channel 2` or `channel 2, channel 3`?) vs direct access (I would expect `ch[1] == channel 1`) makes this very confusing. It needs some thought.

tacaswell:
`ch[1:3]` should give ch 1, 2, 3 ?

prjemian:
Expecting that to be only two channels.  Starting with 1 and up to, but not
including, 3

tacaswell:
The suggestion to use a closed rather than half open set is inspired by pandas with when you have non-integer labels, if you have `obj.ch['A': 'C']` with possible channels `{'A', 'B', 'C', 'D'}` you expect to get back 3 channels.  If you don't do that it is impossible to do something like `obj,ch['B': 'D']` and get back the last one (you can do `['B':]` but then you have to know that 'D' is "the last one" which ends up being a pain on the call side).

If you need it to work that way for labels, then it should work the same for integers.

I am not excited that I am arguing this position (I would rather have half-open like everything else in python), but it is something where I unfortunately think the pandas developers are correct.

Oh, I now understand @klauer 's actually point that there is also confused off-by-one indexing issues going on.

klauer:
Yeah, off-by-one or more accurately "off-by-starting-index". If your channels start at `50` for whatever reason...

Should the slicing have options in this case? Or should we say that if the starting index isn't 0 then handle it specially?

The inclusive `slice.end` for integer indices is also weird, for sure - as you can see @prjemian is already confused at the notion.

klauer:
Is there any rectifying the above or should we just close this?

tacaswell:
I think there is still something promising here.

Although it is a bit weird (and in fact the thing that made me rage-quit pandas every time I tried it), treating the integer-labels as ordered labels suggest that we should use inclusive slicing.  So if you have channels starting at 50 then `obj[0]` -> raises, `obj[50]` gets you the first channels and `obj[50:54]` gets you the first 5 channels.

klauer:
I rewrote `int` slicing to be inclusive on `slc.stop` and raise `ValueError` when invalid indices are used.

That said, my code from ~1.5 months ago is almost incomprehensible to me at this point, which doesn't seem like a good indicator that this should be merged...

danielballan:
Well, my immediate reaction to this was a worry that might be too clever, but I also think it's too soon to give up on it. Maybe some documentation-driven development, comparing the various ways we have for managing nested devices like this, would help.

danielballan:
As a point in _favor_ of going forward with this, I recently helped @jklynch look at some Xspress3 ophyd code, and the problem turned out to be that they were doing

```py
xpress3.channel1.roi01.something
```

when the correct spelling was

```py
xspress3.channel1.rois.roi01.something
```

I offer that as a recent illustration that patterns like `rois.roi01` are most nested than one might expect, and that `rois[1]` might be nicer to use. I like that both invocations continue to work; i.e. that you always have the option to spell it out like ` dev.cpt1.channel_01` in the example at the top of this post, but `dev.cpt1[1]` is available too.

jklynch:
I'm not sure that indexing would have helped with my `xspress3.channel1.rois.roi01` problem *BUT* maybe indexing would be helpful in applying some operation to all the `rois`.

klauer:
It's been 6 months and there has been not one request for this functionality.

It's time to cut losses and close this for good.

PR #810, opened 2020-02-29
MNT: remove annoying teardown messages

klauer:
A torn-down logger can cause these messages to blow up as tracebacks at the end of a pytest session


tacaswell:
The allowed failure configuration is not working, not sure why.

Issue #814, opened 2020-03-06
semantics of Signal.value have changed


tacaswell:
In 8acc424b325e3447db2ac4601e7bc0c6d1403532 which came in via https://github.com/bluesky/ophyd/pull/653 the meaning of `sig.value` changed from "go consult the control system" to "here have my cached value".  For `Signal` this isn't a problem (as setting it will update the cache) and for `EpicsSignal` with auto-monitoring this will also be OK, but for `EpicsSignal` it means that it returns stale values (which caused confusion at a beamline which is testing the new software [1]).  A quick search across the profiles at NSLS-II shows something like ~1.2k usages of `.value` across 20 beamlines.

Short term, I think the correct thing to do is remove the cache logic from `signal.value` because this is a silent gotcha for our users.

Long term do we:
 - want leave it as a maybe-slow attribute?
 - make it raise on EpicsSignal is not monitoring?
 - always auto-monitor with EpicsSignal?
 - deprecate 

I can see a case for all of them.

[1] https://github.com/NSLS-II-SIX/profile_collection/issues/29

prjemian:
Big problem for both XPCS & USAXS, this is XPCS:
https://github.com/BCDA-APS/apstools/issues/270

prjemian:
Monitor EpicsSignal is the assumption at APS, stunned to find that had not
been the case

The monitor should be the most efficient use.  It only updates on new
information.  There are a few fields which do not post monitors (consult
EPICS record reference manual) but those fields can be handled.

tacaswell:
Early on we had two concerns with "monitor and cache".  One was overwhelming the client with too many updates (particularly if you subscribe to a couple of noisey and/or big PVs that you don't actually need (like the Array pvs on AD or some of the stats values).  The other, a bit ironically, is that the protocol specifies that monitors can be dropped on the floor so we were worried about getting stale information.

----

@prjemian Is there already an ophyd issue that I missed?  This is the sort of thing that should definitely be reported upstream!

prjemian:
Thanks.  I will next time.

tacaswell:
As a rule of thumb, if something changes in a way that requires client code to change and it is _not_ documented as an intentional API change, then we want to know (as that is a bug).  If it is documented as an API change and it is from an rc or master we also want to know (as that is a chance to talk us off of that ledge ;) )

PR #815, opened 2020-03-07
API: make Signal.value sometimes defer to self.get()

tacaswell:
This restores the behavior for ophyd <= 1.4.

While properties that may take an (arbitrarily) long time are not
great, silently changing the semantics of the property has too high of
a chance of causing hard to detect bugs in beamline code.

xref #814

I can cook up some tests for this if required (basically get a real PV to talk to, make an epics signal, side-band change it, assert `obj.value` changed).

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -369,7 +369,10 @@ class Signal(OphydObject):
             The keys must be strings and the values must be dict-like
             with the ``event_model.event_descriptor.data_key`` schema.
         """
-        val = self.value
+        if self._readback is DEFAULT_EPICSSIGNAL_VALUE:
+            val = self.get()
+        else:
+            val = self._readback
         return {self.name: {'source': 'SIM:{}'.format(self.name),
                             'dtype': data_type(val),
                             'shape': data_shape(val)}}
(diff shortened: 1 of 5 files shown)


klauer:
Access `.value` and other properties will be updated on a non-monitored signal.
As blocking + property value consistency doesn't appear to be a priority, swaths of the `ophyd.signal` refactor could be cleaned/refactored/simplified.

Do what you must 🤷‍♂

klauer:
Tests stalling on ophyd/tests/test_signal.py::test_set_method[pyepics]

tacaswell:
I would like to squash this down before we merge (mostly out of vanity to hide my errors ;) ).

tacaswell:
The first commit makes `.value` warn when used in a way that will either a) force a `.get` because we have never used this PV before or b) may lie to you so a `.get` is forced (as was done in earlier versions of ophyd).  Long term we make be able to remove all of this complexity or at least change to raising.

The next 4 commits make sure that Signal / EpicsSignal internals don't warn on your behalf.

The next commit fixes up the the test suite to not warn (due to this issue)

The rest of the commits are things I'm willing to push to a separate PR, but are changes I ran into while trying to sort this out, but ended up being un-related.

The next commit handles getting a `value` kwarg passed into EpicsSignal a bit better (IMO).  We used to just drop it on the floor, tried raising but that broke a bunch of tests so now it propagates through.  The rest are hopefully non-controversial.

tacaswell:
I see intermittent failures locally.  Re-running the failed tests they pass with no change to the code.  Some of it seems to be connection related, some of it seems to be the sim motor moves to slow and the sleeps in the tests are not long enough.

PR #820, opened 2020-03-16
Add logging documentation

danielballan:
The first commit copied the logging documentation from bluesky as is. The
second commit makes the necessary changes to account for library-specific
details.

### DIFF
--- a/doc/source/debugging.rst
+++ b/doc/source/debugging.rst
@@ -1,0 +1,191 @@
+*********************
+Debugging and Logging
+*********************
+
+.. versionchanged:: 1.6.0
+
+   Bluesky's use of Python's logging framework has been completely reworked to
+   follow Python's documented best practices for libraries.
+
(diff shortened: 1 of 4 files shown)


doc/source/debugging.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Logger Names
+------------
+
+Here are the primary loggers used by ophyd.
+
+* ``'ophyd'`` --- the logger to which all ophyd log records propagate
+* ``'ophyd.control_layer'`` --- logs requests issued to the underlying control
+  layer (e.g. pyepics, caproto)
+* ``'ophyd.event_dispatcher'`` --- issues regular summaries of the backlog of
```
danielballan: @klauer Is this a precise description? Would you add anything?

klauer: Seems about right. Considering we're the target audience of the log output, what you have is more than sufficient in my opinion.

PR #823, opened 2020-03-17
Flake8 compliance

dmgav:
Refactoring of code to satisfy Flake8 requirements. Code changes include:
- deletions of unused imports and unused variables (all files except `__init__.py`);
- replacement of `raise NotImplemented` with `raise NotImplementedError` (one instance);
- minor formatting to satisfy Flake8 requirements.

Flake8 configuration: maximum line length is set to 115 characters.

Flake8 runs in a separate build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,61 +42,43 @@ jobs:
   include:
     - python: 3.7  # Separate build to run flake8 check
       env:
-        - FLAKE_8=true
+        - FLAKE8=true
     - python: 3.6
       env:
         - PUBLISH_DOCS=true
-        - FLAKE_8=
         - TEST_CL=pyepics
     - python: 3.6
       env:
         - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=caproto
     - python: 3.7
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=pyepics
     - python: 3.7
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=caproto
     - python: nightly
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=pyepics
     - python: nightly
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=caproto
   allow_failures:
     - python: 3.6
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=caproto
     - python: 3.7
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=caproto
     - python: nightly
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=pyepics
     - python: nightly
       env:
-        - PUBLISH_DOCS=
-        - FLAKE_8=
         - TEST_CL=caproto
 
 before_install:
-  - if [ "$FLAKE_8" ]; then
+  - if [ "$FLAKE8" ]; then
         pip install flake8;
         flake8 .;
         let res=$?;
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,6 @@ jobs:
         - TEST_CL=pyepics
     - python: 3.6
       env:
-        - PUBLISH_DOCS=
         - TEST_CL=caproto
     - python: 3.7
       env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,11 +96,12 @@ jobs:
         - TEST_CL=caproto
 
 before_install:
-  - if [ ! -z "$FLAKE_8" ]; then
+  - if [ "$FLAKE_8" ]; then
         pip install flake8;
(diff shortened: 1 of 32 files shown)


ophyd/tests/test_ophydobj.py
```diff
@@ -186,7 +186,7 @@ def test_register_instance():
     test2 = OphydObject(name='test2')
     assert weakdict['test2'] == test2
 
-    assert OphydObject._OphydObject__any_instantiated == True
+    assert OphydObject._OphydObject__any_instantiated is True
```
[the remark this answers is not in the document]
klauer: I think the original check (with the `is` fix) has some validity here - as `assert x` will pass through more things than `assert x is True` will.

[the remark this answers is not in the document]
danielballan: If this were public, we would have to decide whether the API contract is, "This is truth-y but may be something other than `True`." This is a test of an implementation detail of an implementation detail, so it really doesn't matter, and you have a point that a pickier test may be a better test in this specific case.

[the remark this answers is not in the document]
dmgav: I also think that in this case it's better to keep more detailed check

ophyd/tests/test_positioner.py
```diff
@@ -28,7 +28,7 @@ def test_positioner_settle():
 def test_soft_positioner():
     p = SoftPositioner(name='test', egu='egu', limits=(-10, 10))
 
-    assert p.connected == True
+    assert p.connected is True
```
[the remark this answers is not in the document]
dmgav: Fixed.

hhslepicka:
Can we hold on into merging this? All other open PRs will have conflicts after that, wouldn't they?

dmgav:
Yes, I think this should be discussed. I believe the idea was to enable automatic `flake8` checks. I tried to do only minor necessary changes to the code. No code was added, rearranged or moved. So if there are conflicts, they should be minor.

danielballan:
Should we put this on the agenda for Friday’s call? Can you make it, @hhslepicka? It would be nice to see you, in these socially distant times.

hhslepicka:
I'll be there, @danielballan

danielballan:
On the call, we agreed that this is worth doing and that now is a good time to do it. @tacaswell volunteered to help with the rebase workload that this merge will prompt.

ophyd/areadetector/filestore_mixins.py
```diff
@@ -518,7 +518,7 @@ def stage(self):
 
 
 class FileStoreTIFFSquashing(FileStorePluginBase):
-    '''Write out 'squashed' tiffs
+    r'''Write out 'squashed' tiffs
```
MikeHart85: [PEP 257](https://www.python.org/dev/peps/pep-0257/) specifies:

> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""".

But it looks like flake8 doesn't mind triple single quotes (only correctly added the `r`). Not sure if we should care about this? I've seen both ways used throughout various bluesky projects.

dmgav: I did a search for `'''`: there is some number of occurrences of triple single-quotes. I think that they either should be all changed to double quotes (easy) or left alone. What is your opinion?

MikeHart85: I agree it should be all or nothing, but I'll defer to @danielballan on which it should be.

dmgav: I am not sure, but there may be a reason why `'''` is used in some docstrings in this project.

danielballan: I think that reflects a personal preference of the primary ophyd contributor. I think it would be best to leave them as is in this PR. I wasn't aware that PEP 257 weighed in on this; that's good to know, and it also matches _my_ personal preference. :-)

ophyd/utils/__init__.py
```diff
@@ -1,11 +1,10 @@
 # vi: ts=4 sw=4 sts=4 expandtab
-import logging
+import logging  # noqa: F401
```
mrakitin: Strange flake8 complained about it was imported, but unused. I am seeing it's used in line [9](https://github.com/bluesky/ophyd/pull/823/files#diff-f48b3a9fdaf7b4912ee2661cccd8805eR9) below.

.travis.yml
```diff
@@ -37,52 +37,73 @@ cache:
...
+        - FLAKE_8=
         - TEST_CL=caproto
 
 before_install:
+  - if [ "$FLAKE_8" ]; then
+        pip install flake8;
+        flake8 .;
+        let res=$?;
+        echo "The project code was verified by 'flake8'. Exit code ($res).";
+        exit $res;
```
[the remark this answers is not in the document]
dmgav: The problem that we are trying to solve here:

ophyd/utils/__init__.py
dmgav: I guess in this case `noqa` is unnecessary.

PR #827, opened 2020-03-27
ENH #802 connect EPICS soft motor limits

### LINKED ISSUE #802: motor limits updated outside of bluesky
Sometimes, we change our motor limits from other software while a bluesky session is running.  In such cases, ophyd does not learn that these limits have changed.  (See https://github.com/APS-USAXS/ipython-usaxs/issues/331 for example.)  We can also have procedures that will change the limits from a bluesky plan.  In those cases where we know a motor's limits will be adjusted programmatically, we subclass `EpicsMotor` and connect with the motor's `.HLM` and `.LLM` fields.  Same problem applies, ophyd does not learn of the new limits.

Is it true that ophyd's `EpicsSignalBase` only learns the limit values from the PyEpics PV metadata at the initial connection?  Is there any code that can be called to learn of updates?

prjemian:
Fixes #802 by connecting to EPICS motor record soft limits (`.HLM` and `.LLM`) and keeping ophyd `EpicsMotor` object synchronized with EPICS.

Also supplies two methods for call compatibility with SPEC.

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -344,7 +344,6 @@ class EpicsMotor(Device, PositionerBase):
             # update EPICS
             lo = min(low, high)
             hi = max(low, high)
-            print(lo, self.position, hi)
             if lo <= self.position <= hi:
                 self.high_limit_travel.put(lo)
                 self.low_limit_travel.put(hi)
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -90,10 +90,10 @@ class EpicsMotor(Device, PositionerBase):
             update EpicsSignal object when a limit CA monitor received
             """
             if (
-                self.connected
-                and old_value is not None
-                and value != old_value
-            ):
+                    self.connected
+                    and old_value is not None
+                    and value != old_value
+                ):
                 self.user_setpoint._metadata_changed(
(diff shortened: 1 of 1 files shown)


klauer:
Hmm, travis didn't run on those last two commits...

prjemian:
says travis via flake8: `./ophyd/epics_motor.py:96:17: E125 continuation line with same indent as next logical line`.  Context:  https://github.com/bluesky/ophyd/blob/242e847098b879760f560eeb375affcd5c2918f7/ophyd/epics_motor.py#L92-L97

Really?

klauer:
image
The closing parenthesis and the if block shouldn't be on the same column is what it's saying.

```
 if ( 
         self.connected 
         and old_value is not None 
         and value != old_value 
         ): 
     self.user_setpoint._metadata_changed( 
```

ophyd/epics_motor.py
```diff
@@ -280,6 +301,55 @@ def report(self):
...
+        high : float
+           Limit of travel in the positive direction.
+        low : float
+           Limit of travel in the negative direction.
+        '''
+        if not self.moving:
+            # update EPICS
+            lo = min(low, high)
+            hi = max(low, high)
+            print(lo, self.position, hi)
```
klauer, line 347:
```suggestion
```

PR #829, opened 2020-03-31
Use nice error message if Python too low.

danielballan:
Copied from our cookiecutter.

Tested with `python2 setup.py sdist`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,28 @@
 #!/usr/bin/env python
 import os
+import sys
+
 from setuptools import (setup, find_packages)
 import versioneer
+
+# NOTE: This file must remain Python 2 compatible for the foreseeable future,
+# to ensure that we error out properly for people with outdated setuptools
+# and/or pip.
+min_version = (3, 6)
+if sys.version_info < min_version:
+    error = """
+ophyd does not support Python {0}.{1}.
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,16 +1,38 @@
 #!/usr/bin/env python
+import io
```
mrakitin: Where does it come from? I don't think I saw it before, and in https://github.com/bluesky/bluesky/pull/1331...

danielballan: Standard library https://docs.python.org/3/library/io.html

Its implementation of `io.open` is in distinct from `builtins.open` (a.k.a. `open`) in Python 2 and supports the `encoding` argument that we use here. I hope to be forgetting this archaic Python 2 knowledge soon....

MikeHart85: Since you check `sys.version` and exit if it's below 3.6, I don't think using `io.open` is necessary? 

Even if you use the builtin `open`, the interpreter will never get that far, and the friendly error message will be displayed. The file as a whole doesn't need to be python2 compatible; just the part that gets executed when running it with python2.

MikeHart85: It works fine as it is though, so this could safely be merged.

danielballan: This is right. Fixed and force-pushed.

Issue #825, opened 2020-03-27
Review of the Status API


danielballan:
This issue has two goals:

1. Draft developer-oriented documentation about what an ophyd Status object is and the intent behind its design.
2. Propose possible extensions of that API to be tackled in one or more PRs.

As we build out the documentation, it may be useful to revisit these early design choices and understand them better, even if we elect not to change or extend anything. This could be also be useful for future work around handling failure (e.g. alarm states) better.

## Comparison between Status and Future

Ophyd Status objects are like [`concurrent.futures.Future`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future) objects but with some semantics specific to controlling physical hardware. How do they compare?

* Both have methods for adding callbacks that get called once upon completion (success or failure); `Status.add_callback(f)` and `Future.add_done_callback(f)`. A Status objects calls `f()` with no arguments; a Future calls `f(future)` with a reference to itself.
* Futures can be canceled until the executor begins the associated work and marks them as un-cancel-able. Status objects can never be cancelled.
* Many (not all) Status objects have a reference to the `Device` that they are moving. This enables stopping an action when it is already in progress, `status.device.stop()`, which is something a Future cannot do.
* Futures have _results_ which are set by the executor via `Future.set_result(result)` and collected by the caller via `Future.result(timeout=None)`. In the event of failure, an exception is set via `Future.set_exception(exc)`, which can be collected via `Future.exception(timeout=None)`. Calling `Future.result()` on a failed Future raises the exception. Status objects have no result to communicate back; they only have to communicate whether they are in flight, done successfully, or failed. They can log exceptions but do not provide a way to access them programmatically. They provide a boolean `success` property.
* `Status` has a `done` read-only property; `Future` has a `done()` method with the same meaning.
* Both have a concept of `timeout` but with different meanings. The `Status` object takes in `timeout` at `__init__` time and keeps its own clock starting from `__init__`. A `Future` leaves it up to the caller of `result()` or `exception()` to provide an optional `timeout`, and the clock starts when the method is called.
* Status also has a concept of `settle_time`, a constant time added after the underlying task has completed and before it reports being done.
* Status has special methods for reporting incremental progress in order to support progress bars.
* Status supports `__and__`.
* Future takes no arguments at `__init__`. Status takes in `timeout` and `settle_time`, which is needs to do in order to support its different meaning of `timeout`. It also takes in `done` and `success` at `__init__`.

## Possible Changes

[...]

PR #836, opened 2020-04-09
Give callbacks access to status

danielballan:
Addresses Point 1 of #825.

```py
In [1]: def f(): 
   ...:     print("BANG!") 
   ...:                                                                                                                                                                                       

In [2]: def g(status): 
   ...:     print(f"BANG! from {status}") 
   ...:                                                                                                                                                                                       

In [3]: import ophyd                                                                                                                                                                          

In [4]: thing = ophyd.Device(name='thing')                                                                                                                                                    

In [5]: status = ophyd.DeviceStatus(thing)                                                                                                                                                    

In [6]: status.add_callback(f)                                                                                                                                                                
/home/dallan/Repos/bnl/ophyd/ophyd/utils/__init__.py:166: UserWarning: The signature of a Status callback is now expected to be cb(status). The signature cb() is supported, but support will be removed in a future release of ophyd.
  warnings.warn(

In [7]: status.add_callback(g)                                                                                                                                                                

In [8]: status._finished()                                                                                                                                                                    
BANG!
BANG! from DeviceStatus(device=thing, done=True, success=True)
```

The implementation of the backward-compatibility in `adapt_old_callback_signature` is adapted from caproto, where we recently made a similar change.

TO DO:
- [x] Update all internal callbacks to use the new signature.
- [x] Unit tests

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -228,8 +228,13 @@ class StatusBase:
         # Handle func with signature callback() for back-compat.
         callback = adapt_old_callback_signature(callback)
         if self.done:
+            # Call it once and do not hold a reference to it.
             callback(self)
         else:
+            # Hold a strong reference to this. In other contexts we tend to
+            # hold weak references to callbacks, but this is a single-shot
+            # callback, so we will hold a strong reference until we call it,
+            # and then clear this cache to drop the reference(s).
             self._callbacks.append(callback)
 
     @finished_cb.setter
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -8,6 +8,7 @@ from warnings import warn
 import numpy as np
 
 from .log import logger
+from .utils import adapt_old_callback_signature
 
 
 class UseNewProperty(RuntimeError):
@@ -140,7 +141,7 @@ class StatusBase:
             self._settled()
 
             for cb in self._callbacks:
-                cb()
+                cb(self)
             self._callbacks.clear()
 
     def _finished(self, success=True, **kwargs):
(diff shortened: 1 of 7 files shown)


ophyd/status.py
```diff
@@ -203,11 +208,32 @@ def finished_cb(self):
...
+            backward-compatibility but will issue warnings. Support will be
+            removed in a future release of ophyd.
+        """
+        # Handle func with signature callback() for back-compat.
+        callback = adapt_old_callback_signature(callback, self.__wrapper_weakrefs)
         if self.done:
-            cb()
-        else:
-            self._callbacks.append(cb)
+            callback()
```
tacaswell, line 235: sholudn't we be passing the status objec in here?

ophyd/status.py
```diff
@@ -203,11 +208,32 @@ def finished_cb(self):
...
+        self._callbacks.append(callback)
```
tacaswell, line 236: and I think the `else` was an important part of the logic here as we don't need to track callbacks once they have fired.

danielballan: Oh, of course. :sheep:

ophyd/status.py
danielballan: Yep, fixed.

ophyd/status.py
danielballan: The change was in a separate commit. I have force-pushed it out.

ophyd/utils/__init__.py
```diff
@@ -136,3 +138,38 @@ def getattrs(obj, gen):
...
+    sig = inspect.signature(callback)
+    try:
+        # Does this callbacktion accept one positional argument?
+        sig.bind(None)
+    except TypeError:
+        warnings.warn(
+            "The signature of a Status callback is now expected to "
+            "be cb(status). The signature cb() is "
+            "supported, but support will be removed in a future release "
+            "of ophyd.")
```
ZLLentz, line 167:
```suggestion
            "of ophyd.", DeprecationWarning)
```

danielballan:
I can see the argument for passing it by keyword...I think I'm 60/40 in favor of leaving it positional. There is no ambiguity about argument order since there is just one argument, and I can't foresee this growing additional arguments in the future. (Everything we would want to send is encapsulated in `status`.) From a documentation-driven development POV, I like the idea of keeping the differences between `Status` and `concurrent.futures.Future` to minimum.

PR #837, opened 2020-04-10
Refactor Status around a threading.Event and add methods.

danielballan:
This is a fully-backward-compatible implementation of (3) and (4) in #825.

Adds methods:

```py
set_exception(exc)
set_finished()

# Return None (success), *return* exception (failure), or raise WaitTimeoutError (in progress)
exception(timeout=...)

# Return None (success), *raise* exception (failure), or raise WaitTimeoutError (in progress)
wait(timeout=...)
```

The done state is now implemented internally in terms of a `threading.Event` instead of a boolean. The success state is now implemented internally in terms of a stashed exception. A generic exception is used when we are given `Status(sucesss=False, ...)` or `status._finished(success=False)` and the Device has thus not told us the cause of failure.

The polling logic in the timeout thread and in the top-level function `wait(status, timeout, poll_rate)` uses `threading.Event.wait(timeout)` instead.

Example:

```py
In [1]: import ophyd                                                                                                                                                                          

In [2]: thing = ophyd.Device(name='thing')                                                                                                                                                    

In [3]: good_status = ophyd.DeviceStatus(thing)                                                                                                                                               

In [4]: good_status                                                                                                                                                                           
Out[4]: DeviceStatus(device=thing, done=False, success=False)

In [5]: good_status.wait(1)                                                                                                                                                                   
WaitTimeoutError                          Traceback (most recent call last)
<ipython-input-5-f756f3bbec98> in <module>
----> 1 good_status.wait(1)

~/Repos/bnl/ophyd/ophyd/status.py in wait(self, timeout)
    290         """
    291         if not self._event.wait(timeout=timeout):
--> 292             raise WaitTimeoutError("Status has not completed yet.")
    293         if self._exception is not None:
    294             raise self._exception

WaitTimeoutError: Status has not completed yet.

In [6]: good_status.exception(1)                                                                                                                                                              
WaitTimeoutError                          Traceback (most recent call last)
<ipython-input-6-1e2c6444abff> in <module>
----> 1 good_status.exception(1)

~/Repos/bnl/ophyd/ophyd/status.py in exception(self, timeout)
    273         """
    274         if not self._event.wait(timeout=timeout):
--> 275             raise WaitTimeoutError("Status has not completed yet.")
    276         return self._exception
    277 

WaitTimeoutError: Status has not completed yet.

In [7]: good_status.set_finished()                                                                                                                                                            

In [8]: good_status                                                                                                                                                                           
Out[8]: DeviceStatus(device=thing, done=True, success=True)

In [9]: good_status.wait()                                                                                                                                                                    

In [10]: good_status.exception()                                                                                                                                                              

In [11]: class MotorOnFire(Exception): 
    ...:     ... 
    ...:                                                                                                                                                                                      

In [12]: bad_status = ophyd.DeviceStatus(thing)                                                                                                                                               

In [13]: bad_status                                                                                                                                                                           
Out[13]: DeviceStatus(device=thing, done=False, success=False)

In [14]: bad_status.set_exception(MotorOnFire("Oh no!"))                                                                                                                                      

In [15]: bad_status                                                                                                                                                                           
Out[15]: DeviceStatus(device=thing, done=True, success=False)

In [16]: bad_status.exception()                                                                                                                                                               
Out[16]: __main__.MotorOnFire('Oh no!')

In [17]: bad_status.wait()                                                                                                                                                                    
MotorOnFire                               Traceback (most recent call last)
<ipython-input-17-7670c6fbb488> in <module>
----> 1 bad_status.wait()

~/Repos/bnl/ophyd/ophyd/status.py in wait(self, timeout)
    292             raise WaitTimeoutError("Status has not completed yet.")
    293         if self._exception is not None:
--> 294             raise self._exception
    295 
    296     @property

MotorOnFire: Oh no!

In [18]: good_status_old_way = ophyd.DeviceStatus(thing)                                                                                                                                      

In [19]: good_status_old_way._finished()                                                                                                                                                      

In [20]: good_status_old_way                                                                                                                                                                  
Out[20]: DeviceStatus(device=thing, done=True, success=True)

In [21]: bad_status_old_way = ophyd.DeviceStatus(thing)                                                                                                                                       

In [22]: bad_status_old_way._finished(success=False)                                                                                                                                          

In [23]: bad_status_old_way.exception()                                                                                                                                                       
Out[23]: ophyd.utils.errors.UnknownStatusFailure('The status DeviceStatus(device=thing, done=True, success=False) has failed. To obtain more specific, helpful errors in the future, update the Device to use set_exception(...) instead of _finished(success=False).')
```

Needs tests.

The `_finished(...)` method is deprecated but intentionally does not warn because that would be very noisy, and moving everything over to use `set_exception(...)` where appropriate and provide more useful info will take some time. I propose to handle that in separate future PRs.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -46,7 +46,32 @@ class StatusBase:
         The amount of time to wait between the caller specifying that the
         status has completed to running callbacks
     """
-    def __init__(self, *, timeout=None, settle_time=None, done=False,
+    # Theory of operation:
+    #
+    # At __init__ time, a timeout and settle_time are specified  A thread is
+    # started, on which user callbacks, registered after __init__ time via the
+    # method add_callback(callback), will eventually be run.  The thread waits
+    # on self._settled_event to be set or (timeout + settle_time) seconds to
+    # pass, whichever happens first.
+    #
+    # If (timeout + settle_time) expires and the self._settled_event has not
+    # been set, self._exception is set to StatusTimeoutError, and self._event
+    # is set, marking the status as done (and failed). The callbacks are run.
+    #
+    # If self._settled_event is set before (timeout + settle_time) expires,
+    # then self._event is set and the callbacks are run.
+    #
+    # There are two methods that directly set self._settled_event. One,
+    # set_exception(exc), calls it directly after setting self._exception.
+    # The other, set_finished(), starts a threading.Timer that will set
+    # self._settled_event after a delay (the settle_time). One of these methods
+    # may be called, and at most once. If one is called twice or if both are
+    # called, InvalidState is raised. If they are called too late to prevent a
+    # StatusTimeoutError, they are ignored but one call is still allowed.
+    # Thus, an external callback, e.g.  pyepics, may reports success or failure
+    # after the Status object has expired, but to no effect because the
+    # callbacks have already been called and the program has moved on.
+    def __init__(self, *, timeout=None, settle_time=0, done=False,
                  success=False):
         super().__init__()
         self._tname = None
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -742,27 +742,24 @@ def wait(status, timeout=None, *, poll_rate=0.05):
 
     Parameters
     ----------
-    timeout : float, optional
+    status: StatusBase
+        A Status object
+    timeout: Union[Number, None], optional
         Amount of time in seconds to wait. None disables, such that wait() will
         only return when either the status completes or if interrupted by the
         user.
-    poll_rate : float, optional
+    poll_rate: "DEPRECATED"
         DEPRECATED. Has no effect because this does not poll.
 
     Raises
     ------
-    TimeoutError
-        If time waited exceeds specified timeout
-    RuntimeError
-        If the status failed to complete successfully
+    WaitTimeoutError
+        If the status has not completed within ``timeout`` (starting from
+        when this method was called, not from the beginning of the action).
+    Exception
+        This is ``status.exception()``, raised if the status has finished
+        with an error.  This may include ``TimeoutError``, which
+        indicates that the action itself raised ``TimeoutError``, distinct
+        from ``WaitTimeoutError`` above.
     """
-    # It would probably be more useful to just return status.wait(timeout)
-    # directly rather than chaining a RuntimeError on the end here. Is it worth
-    # maintaining back-compat on this?
-    try:
-        return status.wait(timeout)
-    except WaitTimeoutError as exc:
-        raise TimeoutError(*exc.args)
-    except Exception as exc:
-        raise RuntimeError('Operation completed but reported an error: {}'
-                           ''.format(status)) from exc
+    return status.wait(timeout)
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -178,10 +178,6 @@ class StatusBase:
         """
         Sleep for the settle_time, set the Event, run the callbacks.
         """
-        # wait until the settling time is done to mark completion
-        if self.settle_time > 0.0:
-            time.sleep(self.settle_time)
-
         with self._lock:
             if self._event.is_set():
                 # We timed out while waiting for the settle time.
@@ -229,8 +225,9 @@ class StatusBase:
 
         if self.settle_time > 0:
             # delay gratification until the settle time is up
-            self._settle_thread = threading.Thread(
-                target=self._settle_then_run_callbacks, daemon=True,
+            self._settle_thread = threading.Timer(
+                self.settle_time,
+                self._settle_then_run_callbacks,
             )
             self._settle_thread.start()
         else:
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -71,6 +71,11 @@ class StatusBase:
         if timeout is not None:
             self.timeout = float(timeout)
 
+        # We cannot know that we are successful if we are not done.
+        if success and not done:
+            raise ValueError(
+                "Cannot initialize with done=False but success=True.")
+
         self._callback_thread = threading.Thread(
             target=self._run_callbacks, daemon=True, name=self._tname)
         self._callback_thread.start()
@@ -88,10 +93,6 @@ class StatusBase:
                     "set_exception(...) instead of setting success=False "
                     "at __init__ time.")
                 self.set_exception(exc)
-        elif success:
-            # This is a backward-incompatible change.
-            raise ValueError(
-                "Cannot initialize with done=False but success=True.")
 
     @property
     def done(self):
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -8,7 +8,11 @@ from warnings import warn
 import numpy as np
 
 from .log import logger
-from .utils import adapt_old_callback_signature
+from .utils import (
+    adapt_old_callback_signature,
+    UnknownStatusFailure,
+    WaitTimeoutError,
+)
 
 
 class UseNewProperty(RuntimeError):
@@ -45,9 +49,9 @@ class StatusBase:
         super().__init__()
         self._tname = None
         self._lock = threading.RLock()
+        self._event = threading.Event()
         self._callbacks = deque()
-        self._done = done
-        self.success = success
+        self._exception = None
         self.timeout = None
 
         self.log = LoggerAdapter(logger=logger, extra={'status': self})
@@ -60,10 +64,26 @@ class StatusBase:
         if timeout is not None:
             self.timeout = float(timeout)
 
-        if self.done:
-            # in the case of a pre-completed status object,
-            # don't handle timeout
+        # It is a unnecessary that we allow setting 'done' and 'success' at
+        # __init__ time. The concurrent.futures.Future class does not allow
+        # that. But we can continue to support it.
+        if done:
+            if success:
+                self.set_finished()
+            else:
+                exc = UnknownStatusFailure(
+                    f"The status {self!r} has failed. To obtain more specific, "
+                    "helpful errors in the future, update the Device to use "
+                    "set_exception(...) instead of setting success=False "
+                    "at __init__ time.")
+                self.set_exception(exc)
+            # In the case of a pre-completed status object,
+            # don't start the timeout thread.
             return
+        elif success:
+            # This is a backward-incompatible change.
+            raise ValueError(
+                "Cannot initialize with done=False but success=True.")
 
         if self.timeout is not None and self.timeout > 0.0:
             thread = threading.Thread(target=self._wait_and_cleanup,
@@ -79,13 +99,13 @@ class StatusBase:
         This is set to True at __init__ time or by calling `_finished()`. Once
         True, it can never become False.
         """
-        return self._done
+        return self._event.is_set()
 
     @done.setter
     def done(self, value):
         # For now, allow this setter to work only if it has no effect.
         # In a future release, make this property not settable.
-        if bool(self._done) != bool(value):
+        if bool(self._event.is_set()) != bool(value):
             raise RuntimeError(
                 "The done-ness of a status object cannot be changed by "
                 "setting its `done` attribute directly. Call `_finished()`.")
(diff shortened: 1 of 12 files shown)


ophyd/status.py
```diff
@@ -97,15 +117,40 @@ def done(self, value):
...
+    def success(self, value):
+        # For now, allow this setter to work only if it has no effect.
+        # In a future release, make this property not settable.
+        if bool(self.success) != bool(value):
+            raise RuntimeError(
+                "The success state of a status object cannot be changed by "
+                "setting its `success` attribute directly. Call `_finished()`.")
+        warn(
+            "Do not set the `succcess` attribute of a status object directly. "
+            "It should only be set indirectly by calling `_finished()`. "
```
ZLLentz, line 142: This warning and the above error should refer to the newly added `set_finished` method instead of the older `_finished` method

ophyd/status.py
```diff
@@ -161,29 +253,53 @@ def _finished(self, success=True, **kwargs):
...
+        if success:
+            self.set_finished()
         else:
-            self._settle_then_run_callbacks(success=success)
+            # success=False does not give any information about *why* it
+            # failed, so set a generic exception.
+            exc = UnknownStatusFailure(
+                f"The status {self!r} has failed. To obtain more specific, "
+                "helpful errors in the future, update the Device to use "
+                "set_exception(...) instead of _finished(success=False).")
```
ZLLentz, line 264: I'm a big fan of helpful "here's how to make this better" exceptions like this

ophyd/status.py
```diff
@@ -97,15 +117,40 @@ def done(self, value):
...
+    @success.setter
+    def success(self, value):
+        # For now, allow this setter to work only if it has no effect.
+        # In a future release, make this property not settable.
+        if bool(self.success) != bool(value):
+            raise RuntimeError(
+                "The success state of a status object cannot be changed by "
+                "setting its `success` attribute directly. Call `_finished()`.")
+        warn(
+            "Do not set the `succcess` attribute of a status object directly. "
```
klauer, line 141:
```suggestion
            "Do not set the `success` attribute of a status object directly. "
```

ophyd/status.py
```diff
@@ -127,25 +170,74 @@ def _settled(self):
...
+        Status object, but only by the object that created and returned it.
+        """
+        if self.done:
+            # This is fast path. We do a proper check inside a lock in the call
+            # to self._settle_then_run_callbacks.
+            return
+
+        if self.settle_time > 0:
+            # delay gratification until the settle time is up
+            self._settle_thread = threading.Thread(
```
klauer, line 228: Consider replacing with https://docs.python.org/3/library/threading.html#threading.Timer and removing the `time.sleep`

ophyd/status.py
```diff
@@ -127,25 +170,74 @@ def _settled(self):
...
+                target=self._settle_then_run_callbacks, daemon=True,
+            )
+            self._settle_thread.start()
+        else:
+            self._settle_then_run_callbacks()
```
klauer, line 233: I believe this carries forward the old functionality; however, callbacks may either be run in this thread or in another thread depending on `self.settle_time > 0` is rather confusing and may even cause deadlock depending on locks held during the call to `set_finished`.

ophyd/status.py
```diff
@@ -612,19 +737,13 @@ def wait(status, timeout=None, *, poll_rate=0.05):
...
-        if status.success is not None and not status.success:
-            raise RuntimeError('Operation completed but reported an error: {}'
-                               ''.format(status))
-    elif time_exceeded():
-        elapsed = time.time() - t0
-        raise TimeoutError('Operation failed to complete within {} seconds '
-                           '(elapsed {:.2f} sec)'.format(timeout, elapsed))
+    # It would probably be more useful to just return status.wait(timeout)
+    # directly rather than chaining a RuntimeError on the end here. Is it worth
+    # maintaining back-compat on this?
```
klauer, line 742: Shouldn't we then fix the built-in EpicsMotor and such to `set_exception`?

ophyd/status.py
danielballan: That is a good point. Seems reasonable to remove this attempted optimization in the name of uniformity.

ophyd/status.py
danielballan: Yes. I think we can do this incrementally in separate PRs. I expect it will take some thought and significant effort to decide on the right exceptions and error messages in each case. As it stands, anything that uses the old path (`status._finished(success=False)`) will cause an `UnknownStatusFailure` exception to be raised.

The question here is whether it's better to let `UnknownStatusFailure` (and, in the future, more specific and useful exceptions) through here or to chain them behind a general `RuntimeError` in the name of backward-compatibility. We could make `UnknownStatusFailure` a subclass of `RuntimeError` if we like. Either way, I think it's probably useful to let the original exception through, removing

```py
    except WaitTimeoutError as exc:
        raise TimeoutError(*exc.args)
    except Exception as exc:
        raise RuntimeError('Operation completed but reported an error: {}'
                           ''.format(status)) from exc
```
especially since readers of these tracebacks are often not like to read up through a chained exception.

klauer: Makes sense. I'd be in favor of the `UnknownStatusFailure(RuntimeError)` (subclass) option.

danielballan: I tried that, but it turns out the tests still fail with that approach, because `TimeoutError` is caught and retained as `RuntimeError` on master but would just be raised directly with the new approach.

I think it is more useful to raise the exception set by `set_exception` than to chain or wrap it in `RuntimeError`. The error from `wait(status)` should be consistent with the error from `status.wait()`. I now think this is worth an API change. I would guess that raising a different exception from the `wait` would not have much impact on user code. I’m not sure it is used at all.

klauer: I suppose I didn't quite understand the intricacies of it, but I definitely trust your intuition.

klauer:
(Note: tests aren't failing here; coverage thresholds not being met are making it show up as ❌ )

danielballan:
More work needed here, but a stray thought in the mean time. The `wait()` method enables this usage:

```py
motor.set(3).wait(timeout=10)
```

which seems awfully nice for interactive work. I like that it doesn't rely on any special kwargs in `set`.

danielballan:
I think this is ready.

Addressing [@klauer's comment about the pre-existing conditional on which thread the callbacks ran on](https://github.com/bluesky/ophyd/pull/837#discussion_r406974632) led to significant rework of the internals here, so I have requested re-review.

Regarding thread management:
* Formerly, we ran callbacks either on the thread that called `_finished(...)` or on a new thread created by `StatusBase`, conditional on `settle_time != 0`. Now, `StatusBase` always creates a thread, started at `__init__` time, which waits for timeout or `set_finished()`/`set_exception(exc)`, whichever happens first, and runs the callbacks. Thus, the callbacks are consistently always called on this special thread, never on the same thread that called `_finished(...)`/`set_finished()`/`set_exception(exc)`.
* We conditionally create `threading.Timer(...)` thread if `settle_time != 0` but now all this thread does is set an internal `threading.Event`.

See code comments in `__init__` and `_run_callbacks` for more details.

The user API stands as originally proposed. There are two narrow backward-incompatible changes:

* `StatusBase(success=True, done=False)` raises `ValueError` at init time because it does not make sense.
* The top level function `wait(status, ...)` raises whatever `status.wait()` would raise, which may not necessarily be a `RuntimeError` (though it often will be).

danielballan:
It looks like b39218f exposed a race condition (or something). It is easy to reproduce but not 100% of the time.

```
_____________________________ test_pvpos[pyepics] ______________________________
motor = TestEpicsMotor(prefix='XF:31IDA-OP{Tbl-Ax:X1}Mtr', name='epicsmotor', settle_time=0.1, timeout=10.0, read_attrs=['user...trs=['user_offset', 'user_offset_dir', 'velocity', 'acceleration', 'motor_egu', 'high_limit_value', 'low_limit_value'])
    @pytest.mark.motorsim
    def test_pvpos(motor):
        class MyPositioner(PVPositioner):
            '''Setpoint, readback, done, stop. No put completion'''
            setpoint = Cpt(EpicsSignal, '.VAL')
            readback = Cpt(EpicsSignalRO, '.RBV')
            done = Cpt(EpicsSignalRO, '.MOVN')
            stop_signal = Cpt(EpicsSignal, '.STOP')
    
            stop_value = 1
            done_value = 0
    
        m = MyPositioner(motor.prefix, name='pos_no_put_compl')
        m.wait_for_connection()
    
        m.read()
    
        motor.move(0.1, wait=True)
        time.sleep(1)
>       assert m.position == 0.1
E       assert 0.2 == 0.1
E         -0.2
E         +0.1
ophyd/tests/test_pvpositioner.py:67: AssertionError
```

ophyd/status.py
```diff
@@ -127,25 +184,118 @@ def _settled(self):
...
+            raise ValueError(f"Expected an Exception, got {exc!r}")
+
+        # Ban certain Timeout subclasses that have special significance. This
+        # would probably never come up except due to some rare user error, but
+        # if it did it could be very confusing indeed!
+        for exc_class in (StatusTimeoutError, WaitTimeoutError):
+            if (isinstance(exc, exc_class)
+                    or isinstance(exc, type) and issubclass(exc, exc_class)):
+                raise ValueError(
+                    f"{exc_class} as special significance and cannot be set "
```
ZLLentz, line 256:
```suggestion
                    f"{exc_class} has special significance and cannot be set "
```

ZLLentz: This suggestion is rendering incorrectly in my browser, so if it also is rendering incorrectly for you I was trying to suggest:
as -> has

ophyd/status.py
```diff
@@ -40,14 +46,44 @@ class StatusBase:
...
                  success=False):
         super().__init__()
         self._tname = None
         self._lock = threading.RLock()
+        self._event = threading.Event()  # state associated with done-ness
+        self._settled_event = threading.Event()
+        # "Externally initiated" means set_finished() or set_exception(exc) was
+        # called, as opposed to completion via an internal timeout.
+        self._externally_initiated_completion_lock = threading.Lock()
+        self._externally_initiated_completion = threading.Event()
```
klauer, line 84: This is never used as an event and is protected by the respective lock in all instances. 

I think it can be safely made a boolean.

ophyd/status.py
```diff
@@ -40,14 +46,44 @@ class StatusBase:
...
         self._callbacks = deque()
-        self._done = done
-        self.success = success
+        self._exception = None
         self.timeout = None
```
klauer, line 87: `self.timeout` is set-once (i.e., read-only), as we block immediately on the event with the given timeout in `_run_callbacks`

ophyd/status.py
```diff
@@ -127,25 +184,118 @@ def _settled(self):
...
-
+    def _run_callbacks(self):
+        """
+        Set the Event and run the callbacks.
+        """
+        if self.timeout is None:
+            timeout = None
+        else:
+            timeout = self.timeout + self.settle_time
+        if not self._settled_event.wait(timeout):
```
klauer, line 195: I worry a bit that improperly used status objects will leak a thread forever (assuming `timeout=None`), but I'm not sure what could/should be done about that.

This is ostensibly a small issue with no known offenders. That said, bluesky/ophyd processes tend to run for weeks on end for scanning beamlines, where leaking a thread an hour (for example) could add up quickly.

ophyd/status.py
```diff
@@ -40,14 +46,44 @@ class StatusBase:
...
+    # set_exception(exc), calls it directly after setting self._exception.
+    # The other, set_finished(), starts a threading.Timer that will set
+    # self._settled_event after a delay (the settle_time). One of these methods
+    # may be called, and at most once. If one is called twice or if both are
+    # called, InvalidState is raised. If they are called too late to prevent a
+    # StatusTimeoutError, they are ignored but one call is still allowed.
+    # Thus, an external callback, e.g.  pyepics, may reports success or failure
+    # after the Status object has expired, but to no effect because the
+    # callbacks have already been called and the program has moved on.
+    def __init__(self, *, timeout=None, settle_time=0, done=False,
```
klauer, line 74: These were here for subclasses to set as needed, and also for the ability to re-create objects via their repr.

I don't see any major reason that either of the above 2 would need to be respected/carried forward, so feel free to remove them.

danielballan:
New, very reproducible error. No exceptions from threads that I can find.

```
_____________________________ test_pvpos[pyepics] ______________________________
motor = TestEpicsMotor(prefix='XF:31IDA-OP{Tbl-Ax:X1}Mtr', name='epicsmotor', settle_time=0.1, timeout=10.0, read_attrs=['user...trs=['user_offset', 'user_offset_dir', 'velocity', 'acceleration', 'motor_egu', 'high_limit_value', 'low_limit_value'])
    @pytest.mark.motorsim
    def test_pvpos(motor):
        class MyPositioner(PVPositioner):
            '''Setpoint, readback, done, stop. No put completion'''
            setpoint = Cpt(EpicsSignal, '.VAL')
            readback = Cpt(EpicsSignalRO, '.RBV')
            done = Cpt(EpicsSignalRO, '.MOVN')
            stop_signal = Cpt(EpicsSignal, '.STOP')
    
            stop_value = 1
            done_value = 0
    
        m = MyPositioner(motor.prefix, name='pos_no_put_compl')
        m.wait_for_connection()
    
        m.read()
    
        motor.move(0.1, wait=True)
        time.sleep(1)
>       assert m.position == 0.1
E       assert 0.15 == 0.1
E         -0.15
E         +0.1
ophyd/tests/test_pvpositioner.py:67: AssertionError
```

ophyd/status.py
danielballan: Good point!

ophyd/status.py
```diff
@@ -40,14 +46,44 @@ class StatusBase:
         The amount of time to wait between the caller specifying that the
         status has completed to running callbacks
     """
-    def __init__(self, *, timeout=None, settle_time=None, done=False,
+    # Theory of operation:
```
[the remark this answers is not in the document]
danielballan: Fair. I was inspired by trio which had full-on esoteric essays in the module and class comments. I worry that this level of detail might scare away a user would might wonder if this will be on the test. I will give it some thought.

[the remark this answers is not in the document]
danielballan: K, I'm convinced.

danielballan:
The solution of those errors posted about is encouraging! Turns out that Status objects are faster now, so we were "winning" a race in the fixture. The resolution was using the new `wait()` method instead of a polling loop on `DMOV`. Thanks @MikeHart85 for helping me think through the issue.

I addressed all the comments. This should be good to go.

klauer:
Merge at will

PR #720, opened 2019-05-08
REF: pyepics shim

klauer:
Builds on #710 

Closes #715 

TODO
------
* ~Remove temporary commits~
* ~Wait for pyepics version bump~
* ~Increase required version check in pyepics shim~

Issue #715, opened 2019-04-30
New pyepics features call for refactor of pyepics_shim


klauer:
* Better thread safety on `epics.PV` should obviate the need for shim-level threading locks:
https://github.com/NSLS-II/ophyd/blob/cf2b85c2479d645c144289fca79f74b151183417/ophyd/_pyepics_shim.py#L43-L45

* Automatic context switching for `epics.PV` objects should allow for the removal of:
https://github.com/NSLS-II/ophyd/blob/cf2b85c2479d645c144289fca79f74b151183417/ophyd/_pyepics_shim.py#L137

* `auto_monitor` reconfiguration is now possible with an already-created `epics.PV`. Can be refactored:
https://github.com/NSLS-II/ophyd/blob/cf2b85c2479d645c144289fca79f74b151183417/ophyd/_pyepics_shim.py#L68

klauer:
Version requirement will need to be bumped:
https://github.com/NSLS-II/ophyd/blob/cf2b85c2479d645c144289fca79f74b151183417/ophyd/_pyepics_shim.py#L247

PR #842, opened 2020-04-15
Revert to releasing PVs in monitor thread

### LINKED ISSUE #834: Call to `clear_auto_monitor()` in `release_pvs()` results in exception on program exit
As the title, the following script works with ophyd v1.3.3:
```python
from ophyd import AreaDetector, SingleTrigger
class MyDetector(SingleTrigger, AreaDetector): pass
det = MyDetector("13SIM1:", name = "det")
```
However, with v1.4.0, the following exception is thrown on exit:
```
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 346, in get_provider
KeyError: 'epics.clibs'

  File "/usr/local/lib/python3.6/site-packages/epics/ca.py", line 373, in initialize_libca
  File "/usr/local/lib/python3.6/site-packages/epics/ca.py", line 330, in find_libca
  File "/usr/local/lib/python3.6/site-packages/epics/ca.py", line 280, in _find_lib
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1132, in resource_filename
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 348, in get_provider
ImportError: sys.meta_path is None, Python is likely shutting down
```

`git bisect` shows that the bug came into existence at commit a0c602ad, and the problem can be worked around with the following patch:
```diff
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -119,7 +119,6 @@ def release_pvs(*pvs):
         pv._reference_count -= 1
         if pv._reference_count == 0:
             pv.clear_callbacks()
-            pv.clear_auto_monitor()
             if pv.chid is not None:
                 # Clear the channel on the CA-level
                 epics.ca.clear_channel(pv.chid)
```
But I do not know whether this is the correct fix.

MikeHart85:
Closes #834 

Reverts second bullet point of #715, introduced in #720 

Running `release_pvs` in the `monitor` thread seems to avoid the issue described in #834.

### DIFF
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -118,9 +118,12 @@ class PyepicsShimPV(epics.PV):
 
 def release_pvs(*pvs):
     def _release_pvs():
-        for pv in pvs:
-            pv._reference_count -= 1
-            if pv._reference_count == 0:
+        try:
+            for pv in pvs:
+                pv._reference_count -= 1
+                if pv._reference_count != 0:
+                    continue
+
                 pv.clear_callbacks()
                 pv.clear_auto_monitor()
                 if pv.chid is not None:
@@ -132,8 +135,8 @@ def release_pvs(*pvs):
 
                 # Ensure we don't get this same PV back again
                 epics.pv._PVcache_.pop(pv._cache_key, None)
-
-        event.set()
+        finally:
+            event.set()
 
     event = threading.Event()
     _dispatcher.get_thread_context('monitor').run(_release_pvs)
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -1,6 +1,8 @@
 import atexit
 import logging
+import threading
 from distutils.version import LooseVersion
+
 import epics
 from epics import ca, caget, caput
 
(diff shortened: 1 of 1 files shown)


klauer:
Do you think an alternative could be just catching the problematic exception?

While it would not be so much of a "fix" as a patch to silence it, it's something that happens on short-running scripts or potentially on shutdown when we really don't care of PVs have been properly released.

MikeHart85:
The trouble is that the exception occurs outside of our code, so I don't think there's any way we can catch it. The full traceback output looks like this:

```
  File ".../python3.7/site-packages/pkg_resources/__init__.py", line 360, in get_provider
KeyError: 'epics.clibs'

  File ".../python3.7/site-packages/epics/ca.py", line 373, in initialize_libca
  File ".../python3.7/site-packages/epics/ca.py", line 330, in find_libca
  File ".../python3.7/site-packages/epics/ca.py", line 280, in _find_lib
  File ".../python3.7/site-packages/pkg_resources/__init__.py", line 1145, in resource_filename
  File ".../python3.7/site-packages/pkg_resources/__init__.py", line 362, in get_provider
ImportError: sys.meta_path is None, Python is likely shutting down
```

It looks to me like this is triggered by Python internally, as a consequence of `clear_auto_monitor` referencing `ca`. 

The first exception is actually caught, and the second one occurs in the except clause, due to attempting to (re)import `epics.clibs` during teardown:

```python
def get_provider(moduleOrReq):
    """Return an IResourceProvider for the named module or requirement"""
    if isinstance(moduleOrReq, Requirement):
        return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
    try:
        module = sys.modules[moduleOrReq]  # First exception
    except KeyError:
        __import__(moduleOrReq)  # Second exception
        module = sys.modules[moduleOrReq]
    loader = getattr(module, '__loader__', None)
    return _find_adapter(_provider_factories, loader)(module)
```
This is in `pkg_resources`, which is part of setuptools.

MikeHart85:
Do we ever need to care if `clear_auto_monitor`, or even any part of `release_pvs`, has been called during shutdown?

The simplest "solution" would be to just remove the call to `clear_auto_monitor` there. That would avoid needing to reintroduce threading, but I'm not sure if there would be any negative side-effects?

Alternatively, maybe there's a way to detect we're shutting down (as opposed to releasing PVs at runtime), and ignore calls to `release_pvs` in that case?

All of the options seem dirty in some way or another.

klauer:
Thanks, @MikeHart85, for explaining. This should definitely eventually be fixed in pyepics.

It shouldn't matter for ophyd whether that cleanup gets called during shutdown. 

I wonder what would happen if you monkey-patched `epics._find_lib` after a successful startup, ensuring it never tried reloading the library during shutdown? 

For this PR, you might also consider: a `try:/finally:` block such that the event is always set, even if the code fails so it doesn't block the process from exiting.

MikeHart85:
Closing because #845 is a better solution to this.

PR #843, opened 2020-04-16
Deprecate Status(..., done=True)

### LINKED ISSUE #839: Deprecate setting 'done' and 'success' in StatusBase.__init__.
Follow-up to #837 

Should we continue to support making Status object "done" from the very start, as in `StatusBase(success=True, done=True)` or `StatusBase(success=False, done=True)`? I can't think of a reason for needing to set this at `__init__` time. A `concurrent.futures.Future` object doesn't support that, as noted in #825. Instead, it would be just as good to do:

```py
st = StatusBase(); st.set_finished()  # success=True
```

or

```py
st = StatusBase(): st.set_exception(SomeException(...))  # success=False
```

* One can set a specific reason for failure, rather than `success=False`, which sets a blunt `UnknownStatusFailure` exception lacking any information.
* One avoids the possibility of illegal combinations like `StatusBase(success=True, done=False)` ("How can you know you succeeded if you aren't done?").
* The action is clearer (subjectively).

OTOH, if we think this is widely used we can continue to support it indefinitely and just try to advise people not to use it via documentation.

danielballan:
Closes #839

These are harmless, but I think worth deprecating and removing in the name
of simplicity.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -75,7 +75,7 @@ class StatusBase:
 
     """
     def __init__(self, *, timeout=None, settle_time=0,
-                 done=False, success=False):
+                 done=None, success=None):
         super().__init__()
         self._tname = None
         self._lock = threading.RLock()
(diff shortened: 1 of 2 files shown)


ZLLentz:
We could also switch the default values to `None` or some other sentinel value to catch all cases of providing arguments to `done` and `success` for a wider-reaching deprecation warning, just in case someone is calling these with `done=False` for some arcane reason.

danielballan:
Definitely. Fixed and force-pushed.

danielballan:
I'll open up a separate PR retroactively explaining all the changes to Status.

PR #845, opened 2020-04-17
Use finalize instead of __del__ to tear down Signals

### LINKED ISSUE #834: Call to `clear_auto_monitor()` in `release_pvs()` results in exception on program exit

MikeHart85:
Fixes #834

Replaces #842

`__del__` turns out to be unreliable and causes issues when interacting with specific versions of pyepics, epics-base and libca during teardown. This is because the order that things are destroyed in is inconsistent and implementation specific.

`weakref.finalize` is used instead to avoid this problem. Thanks @danielballan for suggesting it!

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -2,6 +2,7 @@
 import time
 import threading
 import warnings
+import weakref
 
 import numpy as np
 
@@ -91,6 +92,8 @@ class Signal(OphydObject):
             timestamp = time.time()
 
         self._destroyed = False
+        self._finalizer = weakref.finalize(self, self.destroy)
+
         self._set_thread = None
         self._tolerance = tolerance
         # self.tolerance is a property
@@ -452,14 +455,6 @@ class Signal(OphydObject):
         '''
         self._destroyed = True
         super().destroy()
-
-    def __del__(self):
-        try:
-            # Attempt to destroy the signal, but ignore any possible exceptions
-            # as Python may have already garbage-collected related objects
-            self.destroy()
-        except Exception:
-            ...
 
     def _run_metadata_callbacks(self):
         'Run SUB_META in the appropriate dispatcher thread'


MikeHart85:
Thanks @klauer, and thanks for all the help and suggestions. Wouldn't have arrived at this without them!

MikeHart85:
Should be ready to merge.

Can't think of a way to meaningfully test for the issue this fixes in a unit test, since it only happens during shutdown and only with specific dependency versions/combinations.

leofang:
Run a subprocess that recreates the shutdown failure and see if an error occurs?

MikeHart85:
That could trigger it, but only with specific combinations of versions of pyepics, epics-base and libca installed on the system the test is running on. It seems more like something to capture in a system test than a unit test, but I don't think we have anything set up for that?

PR #846, opened 2020-04-20
Rewrite Status docs

danielballan:
Includes first draft of release notes for the next release

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -1,6 +1,5 @@
-
-API
-===
+API Reference
+=============
 
 
 
--- a/doc/source/status.rst
+++ b/doc/source/status.rst
@@ -186,11 +186,13 @@ additional options.
 Partial Progress Updates
 ------------------------
 
-Status objects provide an additional method named ``watch``, as in
-:meth:`~ophyd.status.MoveStatus.watch`, which can be used to receive
-incremental progress updates suitable for building progress bars. The method
-accepts a callback which must accept the following parameters as optional
-keyword arguments:
+Some Status objects provide an additional method named ``watch``, as in
+:meth:`~ophyd.status.MoveStatus.watch`, which can be used to subscribe to
+*incremental* progress updates suitable for building progress bars. See
+:doc:`bluesky:progress-bar` for one application of this feature.
+
+The ``watch`` method accepts a callback which must accept the following
+parameters as optional keyword arguments:
 
 * ``name``
 * ``current``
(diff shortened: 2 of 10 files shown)


danielballan:
Part of the thinking here is to imitate the structure of the [trio API docs](https://trio.readthedocs.io/en/stable/reference-core.html), which alternate:

* narrative content and some illustrative examples (very terse, targeted ones, not complete applied examples like a tutorial would have)
* API docs auto-generated from docstrings

Trio puts all of it on one giant page; I'm going for separate pages because I think it's easier to navigate / less overwhelming. I plan to do the same for Signal and Device, Component, etc. in future PRs.

danielballan:
I agree. Fix pushed.

doc/source/api.rst
```diff
@@ -1,39 +0,0 @@
-
-API
-===
-
-
-
-.. autosummary::
```
tacaswell, line 7: Are we sure all of these got re-documented someplace else?

doc/source/status.rst
```diff
@@ -1,36 +1,243 @@
...
+Status API details
+------------------
+
+.. autoclass:: ophyd.status.StatusBase
+   :members:
+
+In addition we provide specialized subclasses that know more about the object
 they are tied to.
 
-.. autosummary::
```
tacaswell, line 21: I like a page-per-class a bit better than having them all inline, but :man_shrugging:

doc/source/status.rst
```diff
@@ -1,36 +1,243 @@
...
+.. code:: python
+
+   from ophyd import Device, Component, SubscriptionStatus
+
+   class MyToyDetector(Device):
+       ...
+       # When set to 1, acquires, and then goes back to 0.
+       acquire = Component(...)
+
+       def trigger(self):
```
tacaswell, line 173: This does not match what the code expects.   It looks like the code expect this callback to return True/False and then calls `self._finished` for you.

doc/source/status.rst
danielballan: The way that h11 and trio flow has converted me to having them inline is helpful when the number of classes is small. I would definitely not change the bluesky plans page to take this approach.

doc/source/status.rst
danielballan: Thanks, fixed.

doc/source/api.rst
danielballan: Reverted to leave this as is.

Issue #841, opened 2020-04-15
Consider making a Signal class backed by Redis


klauer:
Alternative title: consider investigating Signals that can access centralized/beamline-level key-value stores.

This could be used to work around mixing so-called "hard" (backed by IOC) and "soft" (pure Python constructs with no IOC behind them) signals, allowing for consistency between processes and machines.

No modification of IOCs would then be required to support additional ophyd-specific signals and/or metadata, as in the case of the triggering refactor proposed by @tacaswell and @danielballan.

klauer:
To be closed by #904

PR #850, opened 2020-04-26
ENH: first draft of redis backed signals

tacaswell:
Responsive to #841

### DIFF
--- a/ophyd/redis_signal.py
+++ b/ophyd/redis_signal.py
@@ -1,0 +1,117 @@
+import redis
+import json
+from ophyd.ophydobj import OphydObject, Kind
+from ophyd.status import Status
+from ophyd.utils.epics_pvs import data_type, data_shape
+import time
+
+
+class NoKey(KeyError):
+    ...
+
+
+class RedisSignal(OphydObject):
+    def __init__(self, channel, *, r, name=None, **kwargs):
+        if name is None:
+            name = channel
+        super().__init__(name=name, **kwargs)
+        if not isinstance(r, redis.Redis):
+            r = redis.Redis(**r)
+        self._r = r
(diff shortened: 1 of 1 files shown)


tacaswell:
These also compose like you would expect

```python
class RD(Device):
    s = Cpt(RedisSignal, channel='foo', r=r)
    ss = Cpt(StructuredRedisSignal, channel='bar', r=r, schema='abc')
rd = RD(name='bob')
```

klauer:
Neat! Looks like a good start.

mrakitin:
Can we have an example of how to start Redis, e.g. with Docker?

tacaswell:
I did `pacman -S redis` and didn't ask any other questions....

```
docker run  -p 127.0.0.1:6379:6379 --name some-redis -d redis
```

to use the default ports

mrakitin:
Thanks, @tacaswell!

tacaswell:
Thoughts on Actually doing this?

klauer:
> Thoughts on Actually doing this?

What does that mean?

stmudie:
Just noticed this idea. We have a redis based ophyd signal in use on the SAXS beamline at the Australian Synchrotron.  It also supports subscriptions.

tacaswell:
@klauer move this no non-draft and merge it :)

@stmudie Is that signal public or would you be willing to open a PR to ophyd with it?  I was learning how to use redis while writing this so I would be surprised that I got everything right.

stmudie:
@tacaswell The repo is on our internal bitbucket, however, I'm more than happy to work on a PR. I'll integrate some of the ideas from this PR into it and do a bit more testing, and create a PR late next week.

PR #856, opened 2020-04-29
FIX: make Kind.omitted have a bit set

### LINKED ISSUE #855: `Kind.omitted` integer value of `0` results in confusing `__contains__` checks
Referring to this table, where the cell indicates the Python check for `__contains__`,
```python
(row_value) in (col_value)
```

| Row in Col   | Kind.omitted   | Kind.normal   | Kind.config   | Kind.hinted   |
|--------------|----------------|---------------|---------------|---------------|
| Kind.omitted | True           | True          | True          | True          |
| Kind.normal  | False          | True          | False         | True          |
| Kind.config  | False          | False         | True          | False         |
| Kind.hinted  | False          | False         | False         | True          |

`Kind.normal` in `Kind.hinted` makes sense and is done by design.
However `Kind.omitted in Kind.*` are all `True`. This is a problem.

Fix
----
`Kind.omitted` should be set to a unique, non-zero bitmask value.

klauer:
Closes #855 

This results in:
| Row in Col   | Kind.omitted   | Kind.normal   | Kind.config   | Kind.hinted   |
|--------------|----------------|---------------|---------------|---------------|
| Kind.omitted | True           | False         | False         | False         |
| Kind.normal  | False          | True          | False         | True          |
| Kind.config  | False          | False         | True          | False         |
| Kind.hinted  | False          | False         | False         | True          |

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -36,10 +36,10 @@ class Kind(IntFlag):
     traverse it in read(), read_configuration(), or neither. Additionally, if
     decides whether to include its name in `hints['fields']`.
     """
-    omitted = 0b000
-    normal = 0b001
-    config = 0b010
-    hinted = 0b101  # Notice that bool(hinted & normal) is True.
+    omitted = 0b1000
+    normal = 0b0001
+    config = 0b0010
+    hinted = 0b0101  # Notice that bool(hinted & normal) is True.
 
 
 class UnknownSubscription(KeyError):


klauer:
`__contains__` internally of course also uses `&`:

https://github.com/python/cpython/blob/95e208dce505c542b8e4f8f42c57e6d4793b6895/Lib/enum.py#L724-L729

The idea behind`omitted` is that it's saying there are no bits set - which makes sense on its own.
Perhaps it should be removed entirely?

MikeHart85:
Thinking about it some more, `&` doesn't produce correct results either unless omitted has a bit set, so the change makes sense in any case. Still curious to understand why `in` rather than `&`, but shouldn't hold up merging.

klauer:
`in` is just syntactical sugar, of course.

The impetus behind the PR was iterating over signals to show a typhos display - I wanted all normal signals without specifically listing `kind in (Kind.normal, Kind.hinted)` or doing the longer (but certainly equally valid) `(kind & Kind.normal) == Kind.normal`. And `Kind.omitted in Kind.normal` broke that simple check, leading me down this path.

If everyone is still OK with the PR, do we want `omitted` to be first numerically, as @tacaswell mentioned?

MikeHart85:
> `__contains__` internally of course also uses `&`:

It does, but it's not quite the same:

```python
>>> bool(Kind.omitted in Kind.normal)
True
>>> bool(Kind.omitted & Kind.normal)
False
```

Because of the added equality check at the end:

```python
return other._value_ & self._value_ == other._value_
```

MikeHart85:
> I wanted all normal signals without specifically listing `kind in (Kind.normal, Kind.hinted)` or doing the longer (but certainly equally valid) `(kind & Kind.normal) == Kind.normal`

I see, it makes sense now.

I was thinking about checking for individual bits, and Kind.hinted has multiple set... which means you cannot assume Kind.normal is a single bit either.

Maybe *that* is where the design is a bit awkward? Typically every bit in a flag has a defined meaning by itself, even if there are aliases for combinations (or for "no bits set", which is what I took "omitted" to mean).

klauer:
Well, in current master this happens too:
```python
In [1]: bool(Kind.omitted & Kind.omitted)
Out[1]: False
```

Agreed about the awkwardness of `Kind.hinted`. I think this just arose from the reasonable desire to say `kind="hinted"` in components instead of `from ophyd import Kind; Cpt(kind=Kind.normal | Kind.hinted)`

To be clear, I'm still of the opinion that `Kind.omitted` is awkward too.

MikeHart85:
Just brainstorming / fishing for thoughts:

```python
from enum import IntFlag, auto

class Kind(IntFlag):
    # The actual bit flags
    # Not sure about caps, just one way to distinguish visually
    # Arguably _flag or _bit would do the trick just as well
    # auto() assumes the values don't matter, 
    # probably shouldn't use if int value is saved somewhere
    NORM_FLAG = auto()
    CONF_FLAG = auto()
    HINT_FLAG = auto()

    # Aliases for conditions and combinations
    omitted = 0
    hinted = NORM_FLAG | HINT_FLAG

    # These arguably obsolete, could keep for backwards compat
    normal = NORM_FLAG
    config = CONF_FLAG
```

klauer:
All this discussion is great and worthwhile - and vastly superior to "shrug, approve + merge" 😁 

Another idea: what if we left the `Kind` to be all just single-bits, as you've done with `_FLAG` but made our component kind kwarg convert the string `'hinted'` into `Kind.normal|Kind.hinted`? Perhaps a 'hinted config' flag might mean something in the future... ?

MikeHart85:
I think I like it better in the Kind class, because that way all the options are in one place and the implementation of interpreting `kind='textname'` remains simply: `kind = Kind['textname']`

I was a bit worried about whether `Flag`/`IntFlag` was meant to be used this way, but the official docs have examples like...

```python
>>> class Color(Flag):
     RED = auto()
     GREEN = auto()
     BLUE = auto()
     MAGENTA = RED | BLUE
     YELLOW = RED | GREEN
     CYAN = GREEN | BLUE
```
... and the behavior or `repr` also tracks this correctly:

```python
>>> c = Color.RED | Color.BLUE
>>> c
<Color.MAGENTA: 5>
```

Completely tangential thought:

Maybe a proactive `hidden` would be a useful addition (or replacement) for the passive `omitted`?

```python
class Kind:
    hidden = 0b1000
    omitted = 0b0000
```

The distinction being:

If something was hidden, this suggests it was not be shown regardless of any other flags.
If something was omitted, this suggests it was not shown because it didn't qualify for any criteria.

tacaswell:
If I recall correctly, the reason "hinted" is a 2 bit enum is so that `knd.Normal & kind.Hinted` is True and the repr was nice.

klauer:
While it's clear in this scenario, `bool(a & b)` isn't a good general check as it only tells if one or more bits overlap - not which ones.

@MikeHart85 I like the idea of having individual bits in kind and then or'ing them together for ease-of-use. I can't think of any naming scheme for those individual bits that wouldn't be confusing (`hint`) or awkward (`HINT_FLAG`). We may need the naming expert @danielballan to get us out of this.

danielballan:
Heh. Flattered. I will think on it....

klauer:
Waiting on new decisions to be made here. Closing this in favor of a new PR. Suggest bringing discussion back to the issue #855.

PR #859, opened 2020-05-06
Log status callback exceptions

danielballan:


ophyd/status.py
```diff
@@ -257,7 +257,12 @@ def _run_callbacks(self):
         # The callbacks have access to self, from which they can distinguish
         # success or failure.
         for cb in self._callbacks:
-            cb(self)
+            try:
+                cb(self)
+            except Exception:
+                self.log.exception(
```
klauer, line 263: Looks good, but I noticed above some things use `logger` and not `self.log` - is that intentional?

danielballan: Good catch. Looks like a failure to update things after we started using `LoggerAdapter` here. Fixed in c0fc1e7, such that the only remaining usage of `logger` is to wrap it in a `LoggerAdapter`.

danielballan:
Travis is green but the UI hasn't updated: https://travis-ci.org/github/bluesky/ophyd/builds/697618302

PR #863, opened 2020-05-08
Let wait_for_connection() return self instead of None.

danielballan:
Now that we find ourselves adding ``wait_for_connection()`` in more places, it
would be convenient to be able to do it in one statement:

```py
x = SomeSignalClass(...).wait_for_connection()
```

### DIFF
--- a/doc/source/release_notes.rst
+++ b/doc/source/release_notes.rst
@@ -2,6 +2,26 @@
  Release History
 =================
 
+Unreleased
+==========
+
+Added
+------
+
+* The method ``wait_for_connection()`` on various Signal classes now returns
+  the Signal instance. The enables shortening
+
+  .. code:: python
+
+     x = Signal(...)
+     x.wait_for_connection()
+
+  to
+
+  .. code:: python
+
+     x = Signal(...).wait_for_connection()
+
 1.5.0 (2020-05-01)
 ==================
 
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -141,8 +141,14 @@ class Signal(OphydObject):
         return d
 
     def wait_for_connection(self, timeout=0.0):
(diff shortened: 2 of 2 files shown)


prjemian:
Not so fast, there.  Convenient for one-off cases but adds delays when many PVs will be connected since all the waits will happen in series.  While this is a choice and it is not necessary to call the wait right away, I don't believe this is a significant improvement and I believe newcomers will see this as *the right thing to do all the time* when they see it in the documentation.  And then they will ask why does the connection process take so long?

Where are the *more places* that `wait_for_connection()` is being added?

tacaswell:
I have seen this is in beamline profiles.  As we have gotten better about letting the connection process run in parallel in the background, we are finding all of the places that instantiated objects are being used in the profiles.   These things used to "work" because either the PVs of interest were actually waited on or some other PV was and we won the race most of the time :shrug: .

While I agree this is not something we want to use everywhere, I do think this is a nice quality of life improvement for us.

Can me make the same change to `Device`?

prjemian:
I agree with the code changes but we should avoid making this seem as if it the recommended way.  In the instrument profile, define as many EpicsSignal objects as the code block intends, then apply the waits.  The block of waits will finish quickly after the slowest one.  While this looks like more lines of code (it is), it is optimized to make all the connections together rather than in sequence.

klauer:
I share @prjemian 's concerns but also like the shorthand.

danielballan:
I agree with @prjemian's concern. Let's sit on this for a bit and consider what alternatives we might have.

danielballan:
On balance, I think the risk of accidental misuse isn’t worthy the brevity, so I’ll close this proposal. Thanks for spotting the problem and pumping the brakes here, @prjemian .

PR #864, opened 2020-05-11
MNT/API: remove the asyncio integration in ophyd.sim

tacaswell:
It was not quite correct and is causing dead-locks in jupyter notebooks.  Rather than attempt to discover and attach to an asyncio loop, always use threads to implement delays.

This should un break the tutorials notebooks.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,3 +8,4 @@ exclude =
     doc/source,
     ophyd/areadetector/docs.py
 max-line-length = 115
+ignore: W504,W503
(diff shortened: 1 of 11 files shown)


tacaswell:
I tossed a bunch of un-related commits on this to defeat flake8. I can leave them here or move them to their own PR as people wish.

tacaswell:
So apparently a bunch of these code paths are not tested which is not great...

danielballan:
Well, they are soon to be covered by the _de facto_ integration testing of bluesky/tutorials' CI.

PR #865, opened 2020-05-12
Avoid closing over self in the finalizers.

danielballan:
This does not actually fix anything, but it seems a step closer to "good practice" so it may be worth merging for its own sake.

Notice that the original in `master` makes reference to the instance state when the finalizer is _run_ whereas this grabs the attributes it needs when the finalizer is _defined_ and then makes no further reference to `self`.

My reading of this line in the weakref docs

> A more robust alternative can be to define a finalizer which only references the specific functions and objects that it needs, rather than having access to the full state of the object:

is that to make full use of the benefits of `weakref.finalize` over `__del__`, the finalizer should avoid closing over `self`.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -102,7 +102,6 @@ class Signal(OphydObject):
             timestamp = time.time()
 
         self._destroyed = False
-        self._finalizer = weakref.finalize(self, self.destroy)
 
         self._set_thread = None
         self._tolerance = tolerance
@@ -786,6 +785,15 @@ class EpicsSignalBase(Signal):
                            "name={self.name}, id={id(self)}")
             EpicsSignalBase._mark_as_instantiated()
 
+        def finalize(read_pv, cl):
+            cl.release_pvs(read_pv)
+
+        self._read_pv_finalizer = weakref.finalize(self, finalize, self._read_pv, self.cl)
+
+    def destroy(self):
+        super().destroy()
+        self._read_pv_finalizer()
+
     @classmethod
     def _mark_as_instantiated(cls):
         "Update state indicated that this class has been instantiated."
@@ -1165,13 +1173,6 @@ class EpicsSignalBase(Signal):
 
         return {self.name: desc}
 
-    def destroy(self):
-        '''Disconnect the EpicsSignal from the underlying PV instance'''
-        super().destroy()
-        if self._read_pv is not None:
-            self.cl.release_pvs(self._read_pv)
-            self._read_pv = None
-
 
 class EpicsSignalRO(EpicsSignalBase):
     '''A read-only EpicsSignal -- that is, one with no `write_pv`
@@ -1225,6 +1226,16 @@ class EpicsSignalRO(EpicsSignalBase):
         super().__init__(read_pv, string=string, auto_monitor=auto_monitor,
                          name=name, **kwargs)
         self._metadata['write_access'] = False
+
+        def finalize(write_pv, cl):
+            cl.release_pvs(write_pv)
+
+        self._write_pv_finalizer = weakref.finalize(self, finalize, self._write_pv, self.cl)
+
+    def destroy(self):
+        super().destroy()
+        self._write_pv_finalizer()
+
 
     def put(self, *args, **kwargs):
         'Disabled for a read-only signal'
@@ -1710,13 +1721,6 @@ class EpicsSignal(EpicsSignalBase):
     @use_limits.setter
     def use_limits(self, value):
         self._use_limits = bool(value)
-
-    def destroy(self):
-        '''Destroy the EpicsSignal from the underlying PV instance'''
-        super().destroy()
-        if self._write_pv is not None:
-            self.cl.release_pvs(self._write_pv)
-            self._write_pv = None
 
 
 class AttributeSignal(Signal):
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1227,16 +1227,6 @@ class EpicsSignalRO(EpicsSignalBase):
                          name=name, **kwargs)
         self._metadata['write_access'] = False
 
-        def finalize(write_pv, cl):
-            cl.release_pvs(write_pv)
-
-        self._write_pv_finalizer = weakref.finalize(self, finalize, self._write_pv, self.cl)
-
-    def destroy(self):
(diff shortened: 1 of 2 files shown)


danielballan:
Any reason you can think of not to do this, @MikeHart85?

ophyd/signal.py
```diff
@@ -102,7 +102,6 @@ def __init__(self, *, name, value=0., timestamp=None, parent=None,
             timestamp = time.time()
 
         self._destroyed = False
-        self._finalizer = weakref.finalize(self, self.destroy)
```
MikeHart85, line 105: If I'm reading these changes correctly:

With this removed, there is no longer a mechanism by which `Signal.destroy` will be called automatically when a `Signal` (or any derived class) is torn down. It will now only happen if called manually.

Previous behavior was that `destroy` was effectively the destructor for `Signal` and derived classes, and called automatically... initially by `Signal.__del__` and then by this finalizer. If a derived class implemented its own `destroy`, it would get called automatically instead (and was responsible for propagating upward via `super`).

I just want to make sure I'm reading this right and that this is intended, since it seems like a pretty big departure from how Signal teardown used to work.

danielballan: Good point. Added release notes entry (which I should have done in any case) explaining the implementations of the change. Please review.

ophyd/signal.py
```diff
@@ -1165,13 +1173,6 @@ def describe(self):
 
         return {self.name: desc}
 
-    def destroy(self):
-        '''Disconnect the EpicsSignal from the underlying PV instance'''
-        super().destroy()
-        if self._read_pv is not None:
-            self.cl.release_pvs(self._read_pv)
-            self._read_pv = None
```
MikeHart85, line 1173: Short version: Do you think there's any value in keeping the `self._read_pv = None` in the new `finalize` for potential side-effects? (and same for the write version of course)

Long version:

This follows suit with the example in the weakref docs, removing the `not None` check and `= None` on `_read_pv`. Same with the write. The idea being that finalizers can only be invoked once, so there's no need to guard from repeated calls.

But the weakref example is very simple; their `self.name` is just a string and setting it to `None` served no other purpose than that condition. Here, `self._read_pv` is used in many other places and is a PV object. Setting it to `None` (or not) may have other consequences (including when the PV object is garbage collected).

That said, none of the tests are failing and I can't think of any real scenario where this would make a tangible difference. I think it would only come into play if `destroy` was called manually, and the reference to the `Signal` instance kept around (or even attempted to be used again, which you shouldn't do anyway). So feel free to handwave this as a non-issue.

MikeHart85:
For what it's worth, I wrote up some quick sample code to better understand the impact of this:

```python
import gc
import weakref
from time import sleep

class A:
    def __init__(self, name):
        self._name = name

        def finalize(name):
            print('[%s] finalized!' % name)
        self._finalizer = weakref.finalize(self, finalize, name)

    def __del__(self):
        print('[%s] garbage collected!' % self._name)


class B:
    def __init__(self, name):
        self._name = name
        self._finalizer = weakref.finalize(self, self.finalize, name)

    def finalize(self, name):
        print('[%s] finalized!' % name)

    def __del__(self):
        print('[%s] garbage collected!' % self._name)



foo = A('foo')  # Runtime delete, no self reference
bar = A('bar')  # Shutdown delete
baz = B('baz')  # Runtime delete, self reference

print('--- tick ---')
sleep(1)

foo = None
# del foo
baz = None
# del baz

gc.collect()  # Force garbage collection

print('--- tock ---')
sleep(1)
```

Output:

```
$ python weakref_manual.py 
--- tick ---
[foo] garbage collected!
[foo] finalized!
--- tock ---
[baz] finalized!
[baz] garbage collected!
[bar] finalized!
[bar] garbage collected!
```

Note A (foo) and B (baz) behave very differently, even though both are "freed" after the `tick`. The only difference is avoiding the self reference, as you did here. Not even in terms of what's actually referenced **inside** the finalizer, but just having the self reference passed in by virtue of being a method is enough to prevent garbage collection at the expected time. This could lead to memory leaks!

A curious difference is also that foo is `__del__`eted first, and then finalized. Whereas the other two are finalized first. This was consistent no matter how many times I ran this. The deciding factor seems to be manually deleting or setting to `None`.

Setting to `None` or using `del` made no difference.

Anyway, this is just to confirm that this isn't just a matter of best practices, but has real impact. Finalizers should not reference self, or even be methods of the object being finalized.

danielballan:
That is a very good to know!

> Not even in terms of what's actually referenced inside the finalizer, but just having the self reference passed in by virtue of being a method is enough to prevent garbage collection at the expected time.

This matches what I think I understand about methods, in that they are created from the instance each time they are called, so they definitely need `self`.

> This could lead to memory leaks!

It sure looks like that way. Indeed, this has a bigger impact than I realized. Glad we caught it.

MikeHart85:
Yeah, in retrospect it makes perfect sense. A method reference to `self.finalize` must of course contain a reference to `self`... which means now weakref maintains a reference to the instance we're looking to finalize, so the refcount never drops to 0, so it never gets gc'ed until Python shuts down.

I guess I just naively assumed that, since it says "weakref" on the tin, there must be some magic in there to ensure the reference count is adjusted to prevent that. But it's just not that magical. :)

Ready to merge or are you still thinking about the `self._read_pv = None` above?

danielballan:
I got absorbed in your other comment and missed that one. Looks like it should be addressed before merging.

ophyd/signal.py
danielballan: I'm of two minds about this.

A "dead" object is more useful for post-mortem debugging than `None`. If the state is a file, one often keeps a closed file handle, which can still be inspected to show its name as so on. Here we might keep a released PV for the same reason.

I usually err on the side of cleaning up state rather than keeping stuff around "in case you need it" for debugging, particularly in the case of large data structures that take up significant memory, but if the released PV is no longer using network resources of significant memory, I'm inclined to keep it for convenience if we cannot think of any bad implications.

MikeHart85: Sounds reasonable to me, and I can't think of any downside in practice.

danielballan:
For anyone following alone, the decision in a discussion inline above was to leave it as is. Ready to merge.

PR #866, opened 2020-05-12
CA library not safe to use when shutting down

MikeHart85:
Following up on the recurring unexpected exceptions/segfaults during teardown:

@danielballan and @dmgav and I chatted about this a bit this morning and in a nutshell concluded:

1. Solving this in pyepics, along the lines of what @klauer you proposed [here](https://github.com/pyepics/pyepics/compare/master...klauer:fix_shutdown_perhaps), is ultimately the correct way to tackle this. It will need some fleshing out though, to address the additional exceptions I mentioned here https://github.com/bluesky/ophyd/issues/834#issuecomment-615480304. 
2. Alternatively, or additionally, to make ophyd play nice with current versions of pyepics, it may be a good idea to have ophyd avoid making any calls that involve libCA once it detects that Python is shutting down, since it is no longer safe to do so at that point.

The change in this PR is an idea on how to achieve point 2.

It appears to work, in that the caproto tests run without problems with this in place.

It feels a bit dirty though, and makes me question my sanity, so I welcome any and all critical comments and thoughts on why this is a horrible idea.

### DIFF
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -1,20 +1,11 @@
 import atexit
 import logging
-from distutils.version import LooseVersion
+
 import epics
 from epics import ca, caget, caput
 
+from distutils.version import LooseVersion
 from ._dispatch import _CallbackThread, EventDispatcher, wrap_callback
-
-import atexit
-
-
-def invalidate_ca():
-    global ca
-    ca = None
-
-
-atexit.register(invalidate_ca)
 
 _min_pyepics = '3.4.0'
 
@@ -31,6 +22,15 @@ else:
     thread_class = ca.CAThread
 
 
+def invalidate_ca():
+    global _ca_valid
+    _ca_valid = False
+
+
+# LibCA is not safe to use once Python begins shutting down
+_ca_valid = True
+atexit.register(invalidate_ca)
+
 module_logger = logging.getLogger(__name__)
 name = 'pyepics'
 _dispatcher = None
@@ -38,7 +38,7 @@ get_pv = epics.get_pv
 
 
 def get_dispatcher():
-    'The event dispatcher for the pyepics control layer'
+    """The event dispatcher for the pyepics control layer"""
     return _dispatcher
 
 
@@ -125,7 +125,7 @@ class PyepicsShimPV(epics.PV):
 
 
 def release_pvs(*pvs):
-    if ca is None:
+    if not _ca_valid:
         return
 
     for pv in pvs:
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -7,7 +7,14 @@ from epics import ca, caget, caput
 from ._dispatch import _CallbackThread, EventDispatcher, wrap_callback
 
 import atexit
-atexit.register(lambda: ca = None)
+
+
+def invalidate_ca():
+    global ca
+    ca = None
+
+
+atexit.register(invalidate_ca)
 
 _min_pyepics = '3.4.0'
 
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -6,6 +6,9 @@ from epics import ca, caget, caput
 
 from ._dispatch import _CallbackThread, EventDispatcher, wrap_callback
 
+import atexit
+atexit.register(lambda: ca = None)
+
 _min_pyepics = '3.4.0'
 
 if LooseVersion(epics.__version__) < LooseVersion(_min_pyepics):
@@ -115,6 +118,9 @@ class PyepicsShimPV(epics.PV):
 
 
 def release_pvs(*pvs):
+    if ca is None:
+        return
+
     for pv in pvs:
         pv._reference_count -= 1
         if pv._reference_count == 0:


ophyd/_pyepics_shim.py
```diff
@@ -6,6 +6,16 @@
 
 from ._dispatch import _CallbackThread, EventDispatcher, wrap_callback
 
+import atexit
+
+
+def invalidate_ca():
+    global ca
+    ca = None
```
dmgav, line 14: May be it's a good idea to make a comment in the code on why exactly we want to invalidate `ca` (accidental attempt to use `ca` during interpreter shutdown will raise an exception, which is better than segfault, and we also check if `ca` is `None` when needed).

MikeHart85: Agreed, this definitely needs some comments to explain what's going on if we decide to go ahead with this.

As you correctly picked up, the basic idea is to make attempts to use `ca` "fail in a better way" so to speak. And to force the failure, instead of depending on seemingly random circumstances, like calling `pytest` with exactly the right arguments, etc.

One problem with this that we also import `epics`, `caget` and `caput` above. Even if we `None` all of them here, there are still references to `pv`s floating around and, for example, calls to `pv.clear_auto_monitor()` will trigger the problem.

So vigilance is still required and arguably picking on `ca` in particular is a bit arbitrary here.

tacaswell:
I was about to make the exact same comment as @danielballan (but with a global bool flag rather than a list).

MikeHart85:
I was going to ask: Is there an advantage to using a list over a bool?

dmgav:
Surprisingly, I thought about the separate bool flag too, but then I noticed that `ca` is used a few more times throughout the module, so ideally we should check if the flag is set in every function that is using `ca`.

MikeHart85:
I was hoping to get away with only checking in release_pvs (for now?), since I think that's the only one that could realistically be called during shutdown... is it safe to tentatively assume that, until proven otherwise?

leofang:
> I was going to ask: Is there an advantage to using a list over a bool?

I have no idea, but I know a few projects use list. I'm interested too.

danielballan:
The advantage is that if you do:

```py
# modue.py
global_state = True

def change_state():
    global_state = False
```

and some other package in the module does

```py
from .module import global_state
```

it will not get up to date information, where if you mutate `global_state` itself the reference remains useful. But that is not applicable here and I think the current implementation is good.

danielballan:
Power-cycled to re-run CI now that #864 is merged, which fixed some flake8 issues that appeared on `master` due to flake8 becoming stricter.

leofang:
> ```
> from .module import global_state
> ```
> it will not get up to date information

Yeah I think some style guides forbid importing module variables for exactly this reason.

MikeHart85:
Thanks for the explanation @danielballan!

I modified it to work the way you suggested. Seems safer / a good habit to encourage, even if not strictly needed the way it is used here.

danielballan:
Will merge/release by end of week if there are no objections.

klauer:
This still isn't quite right. Downstream library typhos test suite is hanging due to ophyd cleanup even with this PR:

```
Thread 0x00000001133fedc0 (most recent call first):
  File "/Users/klauer/docs/Repos/pyepics/epics/ca.py", line 882 in pend_io
  File "/Users/klauer/docs/Repos/pyepics/epics/ca.py", line 544 in wrapper
  File "/Users/klauer/docs/Repos/pyepics/epics/ca.py", line 905 in poll
  File "/Users/klauer/docs/Repos/pyepics/epics/ca.py", line 544 in wrapper
  File "/Users/klauer/docs/Repos/pyepics/epics/ca.py", line 1053 in connect_channel
  File "/Users/klauer/docs/Repos/pyepics/epics/ca.py", line 570 in wrapper
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 437 in connect
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 48 in wrapped
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 422 in wait_for_connection
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 48 in wrapped
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 729 in get_ctrlvars
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 48 in wrapped
  File "/Users/klauer/docs/Repos/ophyd/ophyd/_pyepics_shim.py", line 103 in _getarg
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 963 in count
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 357 in _check_auto_monitor
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 48 in wrapped
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 409 in auto_monitor
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 48 in wrapped
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 444 in clear_auto_monitor
  File "/Users/klauer/docs/Repos/pyepics/epics/pv.py", line 48 in wrapped
  File "/Users/klauer/docs/Repos/ophyd/ophyd/_pyepics_shim.py", line 136 in release_pvs
  File "/Users/klauer/docs/Repos/ophyd/ophyd/signal.py", line 1708 in destroy
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/weakref.py", line 548 in __call__
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/weakref.py", line 624 in _exitfunc
```

Note that release_pvs clears auto monitor, which tries to connect because `count` isn't available. I had tried to fix this on pyepics's side, but apparently it was a huge can of worms (https://github.com/pyepics/pyepics/pull/203)

danielballan:
Thanks, @klauer. Will hold off in that case.

klauer:
@hhslepicka with the new pyepics master, is your test suite still failing?

hhslepicka:
Not failing anymore

MikeHart85:
Just checking that I understood correctly: Typhos tests pass okay with both the latest pyepics master and this branch of ophyd, correct?

To be sure we're not ending up with redundant fixes I tried various on/off combinations of:

1. this branch
2. https://github.com/pyepics/pyepics/pull/204
3. https://github.com/pyepics/pyepics/commit/1bb4a68c81d5fe128cfdd43aa9cebf5d4bdeb9a2
4. https://github.com/bluesky/ophyd/pull/866

Caproto's tests pass with both 1+4 and 1+2+4. They fail with 2+4, 2+3, and 2+3+4.

If I understood correctly, Typhos tests pass with 1+2+4, but fail with just 1+4.

In summary: This PR should still be merged, but some downstream projects may experience failures unless they also have pyepics at the current master or later. 2 and 4 are needed and already merged, so no action required. 3 is not merged and doesn't need to be, so no action required.

Is this consistent with everyone else's understanding?

klauer:
I think that's probably about right. Though any project with many disconnected PVs will greatly increase the possibility of the teardown hanging without (2) and (3). It certainly may happen with the ophyd test suite. Try even the simple:

```python
import ophyd

sigs = {
    i: ophyd.EpicsSignal(f'sim:mtr{i}', name=f'sig{i}')
    for i in range(10000)
}
```
Hangs with pyepics/pyepics@9371da8 (on clear_auto_monitor)
Succeeds with pyepics/pyepics@master

This, in my opinion, indicates that there should be a new push for a pyepics tag and a bump in ophyd's requirement of it.

hhslepicka:
Typhos pass with:
- pyepics - fixed from 204
- ophyd - master

MikeHart85:
Okay, so we probably want to merge this and maybe beta tag it so we can fix caproto tests, but hold off on a full ophyd release until (3) is merged and pyepics has a new tag.

It would be great to confirm this branch doesn't break Typhos tests as well.

MikeHart85:
And a heads up on the pyepics side of things, with the current master the following snippet still triggers an uncatchable exception on teardown:

```python
import epics

class Broken:
    def __init__(self, pv):
        self.pv = epics.get_pv(pv)
    def __del__(self):
        self.pv.clear_auto_monitor()

mypv = Broken('sim:mtr1')
```

Adding (3) replaces the exception with the RuntimeError that is introduced there. But that is a big improvement, because this one can be caught in __del__.

Using weakref.finalize in a simple example like this sidesteps the issue entirely, but as we've seen that's not enough by itself in more complex situations.

MikeHart85:
I also tried replacing, in epics/ca.py:

```python
    if AUTO_CLEANUP:
        atexit.register(finalize_libca)
```
... with ...
```python
    if AUTO_CLEANUP:
        weakref.finalize(sys.modules[__name__], finalize_libca)
```
... hoping that this would delay finalization enough to avoid the problem altogether, because the module shouldn't go out of scope until nothing needs it, ergo after __del__ and any other finalizers referencing things in `ca` ran. 

Unfortunately it made no noticeable difference. :(

hhslepicka:
@MikeHart85 why not pursue the fix for that at pyepics instead of patching here?
Seems like the best way despite not being the easier way.

MikeHart85:
That would be ideal, agreed. Initially it just wasn't really clear to me whether the error was on our end (calling functions we shouldn't be during teardown) or in pyepics (failing on function calls that should succeed or fail more gracefully). 

Now I think it's fair to say it's the latter, but we already have these workarounds, and not yet any complete solution in pyepics. If we can find one, I'd certainly prefer that to this.

MikeHart85:
The pyepics PR referenced above would make this PR obsolete, but let's see if it is accepted before closing this.

MikeHart85:
With the pyepics PR merged, this PR is likely obsolete and can be closed without merging.

I'm not seeing Caproto test failure anymore locally, when using Ophyd 1.5.0 and the latest pyepics master. Would be great to confirm that this is also the case in CI and other projects that had similar issues.

PR #867, opened 2020-05-12
Release notes for 1.5.1

danielballan:
Dated for tomorrow, anticipating a release. May need updating depending on
whether https://github.com/bluesky/ophyd/pull/866 is merged or not.

PR #872, opened 2020-06-11
Release notes for version 1.5.1

MikeHart85:
Started this branch before I realized #867 existed, but I've copied over the first point from there. The second point I think was for #866, but that got scrapped in favor of the upstream fix.


MikeHart85:
Should merge #871 first. It's already referenced here.

Issue #876, opened 2020-06-17
Do not start threads on import.


danielballan:
This is less than ideal:

```py
>>> threading.active_count()
1
>>> import ophyd.sim
>>> threading.active_count()
11
```

I guess this is due to the various simulated signals that update on a background thread. It is worth considering an API breaking change to move these out of global scope (only create them when `hw()` is called) and/or add a special method to explicitly start their background thread.

dmgav:
At the first glance I have 17 new threads started once I run `import ophyd.sim`: 15 of those threads are started by https://github.com/bluesky/ophyd/blob/b572a4caa2d0be29cf74f60218e684c62b0dac04/ophyd/__init__.py#L72
and the other 2 by https://github.com/bluesky/ophyd/blob/b572a4caa2d0be29cf74f60218e684c62b0dac04/ophyd/sim.py#L1434

danielballan:
The two threads from `ophyd.sim` itself generate INFO-level log messages periodically, such that `logging.basicConfig()` gets very noisy. (A downstream user reported this. We tend not to use `basicConfig()` because it's a blunt instrument, but nonetheless I don't think it would be reasonable to say, "Just don't use `logging.basicConfig()`.") The original motivation for this issue was to resolve that. But now I'm also curious why we see different numbers of threads.

What output do you get for this, @dmgav?

```py
>>> import threading
>>> [t.name for t in threading.enumerate()]
['MainThread']
>>> import ophyd
^[[A[t.name for t in threading.enumerate()]
['MainThread', 'metadata', 'monitor', 'get_put', 'util0', 'util1', 'util2', 'util3', 'debug_monitor']
>>> import ophyd.sim
>>> [t.name for t in threading.enumerate()]
['MainThread', 'metadata', 'monitor', 'get_put', 'util0', 'util1', 'util2', 'util3', 'debug_monitor', 'Thread-9', 'Thread-10']

```

dmgav:
```
In [1]: import threading

In [2]: [t.name for t in threading.enumerate()]
Out[2]: ['MainThread', 'IPythonHistorySavingThread']

In [3]: import ophyd

In [4]: [t.name for t in threading.enumerate()]
Out[4]: 
['MainThread',
 'IPythonHistorySavingThread',
 'selector',
 'command',
 'check_for_unresponsive_servers',
 'retry',
 'search',
 'activate_subscriptions',
 'selector',
 'metadata',
 'monitor',
 'get_put',
 'util0',
[...]

PR #877, opened 2020-06-22
'ophyd.sim': prevent threads for unused objects ('rand', 'rand2') from starting automatically

dmgav:
This PR addresses the issue https://github.com/bluesky/ophyd/issues/876

The objects `rand` and `rand2`, which are instances of the class `SynPeriodicSignal`, are created at the time of importing of `ophyd.sim`. The constructor for the class `SynPeriodicSignal` is automatically starting a background thread for each object even if the object is not used in simulation. In this PR a new method `start_simulation` is added to `SynPeriodicSignal` to start the background thread before the object is used. The constructor now is not creating the thread. Instead, the method has to be called manually in order to start periodic updates of the PV. This breaks the existing API and may prevent existing simulation code from running correctly.

In order to provide backward compatibility, a number of methods of `Signal` and `SynSignal` classes are overridden in `SynPeriodicSignal`, including `trigger`, `get`, `put`, `set`, `read` and `subscribe` methods. In addition to their original functionality, those methods also verify if the background thread is started for the object and start it if necessary. In addition, those methods are printing a deprecation warning with informative message on how to fix the issue. The message is printed only once (before the thread is created), so it is not expected to be annoying.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -297,7 +297,7 @@ class SynPeriodicSignal(SynSignal):
 
     def subscribe(self, *args, **kwargs):
         self._start_simulation_deprecated()
-        return super().subscribe()
+        return super().subscribe(*args, **kwargs)
 
 
 class _ReadbackSignal(Signal):
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -226,6 +226,10 @@ class SynPeriodicSignal(SynSignal):
                  **kwargs):
         if func is None:
             func = np.random.rand
+
+        self._period = period
+        self._period_jitter = period_jitter
+
         super().__init__(name=name, func=func,
                          exposure_time=exposure_time,
                          parent=parent, labels=labels, kind=kind,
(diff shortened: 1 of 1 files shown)


ophyd/sim.py
```diff
@@ -226,30 +226,78 @@ def __init__(self, func=None, *,
...
+
+    def _start_simulation_deprecated(self):
+        """Call `start_simulation` and print deprecation warning"""
+        if self.__thread is None:
+            msg = ("Deprecated API: Objects of SynPeriodicSignal must be initialized before simulation\n"
+                   "by calling 'start_simulation()' method. Two such objects ('rand' and 'rand2') are\n"
+                   "created by 'ophyd.sim' module. Call\n"
+                   "    rand.start_simulation() or rand2.start_simulation()\n"
+                   "before the object is used.")
+            logger.warning(msg)
```
danielballan, line 275: Nit pick: I think using `self.log`, defines by the base class `OphydObj`, can be used here to provide additional info—namely, which instance this warning is being reported from.

dmgav: Yes, I think it's a good idea to replace the logger. The only thing is that I thought that the deprecation warning should be displayed to the user no matter what and logger must be explicitly activated, which may not be done for simulated runs (examples). Or the logs may go into a file. If we want users to update the examples, then may be it's a good idea to print the message as well?

danielballan: >  logger must be explicitly activated

Python's default configuration includes [a "handler of last resort"](https://docs.python.org/3/library/logging.html#logging.lastResort) which writes messages of level WARNING or higher to the stderr. Thus, the warnings here _will_ be shown unless the message is processed by an explicitly registered handler that directs it elsewhere. In the case our tutorials/examples, there are no other handlers registered. In the case of beamlines, I believe that WARNING and higher are shown in the stderr and INFO and higher are logged to a file.

We generally avoid printing because there is no way for downstream libraries to silence prints.

danielballan:
P.S. For the record, I confirmed locally that this reduces the number of threads from 11 (as documented in #876) to 9:

```py
>>> import threading
>>> threading.active_count()
1
>>> import ophyd
>>> threading.active_count()
9
>>> import ophyd.sim
>>> threading.active_count()
9
```

PR #879, opened 2020-06-23
Do not set the default logging handler ('NullHandler') at import

dmgav:
This PR follows the discussion with @danielballan regarding the default behavior of logging: the default behavior of the logger should be to print all logged messages with levels `warning` and above. Currently, the logger is set to use `NullHandler` at the time of importing Ophyd packages. This results in no log messages printed unless other handers are added, which needs to be done explicitly in the code. 
Changes: the line that adds `NullHandler` is deleted, therefore no handlers are added to the Ophyd logger during import.

### DIFF
--- a/ophyd/__init__.py
+++ b/ophyd/__init__.py
@@ -4,7 +4,6 @@ import os
 from ophyd.log import set_handler  # noqa: F401
 
 logger = logging.getLogger(__name__)
-logger.addHandler(logging.NullHandler())
 
 cl = None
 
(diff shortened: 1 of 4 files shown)


klauer:
There are others:
```
$ git grep NullHandler
ophyd/__init__.py:logger.addHandler(logging.NullHandler())    # <-- this PR
ophyd/areadetector/__init__.py:logger.addHandler(logging.NullHandler())
ophyd/utils/__init__.py:logger.addHandler(logging.NullHandler())
ophyd/utils/startup.py:logger.addHandler(logging.NullHandler())
```

dmgav:
@klauer Good point. I removed those entries as well.

danielballan:
Will merge Friday if there are no objections.

PR #881, opened 2020-06-29
FIX: mixin classes that have components must inherit from Device

tacaswell:
This is likely a side-effect of changing from using a full meta-class
to `__init_subclass__`.

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1205,7 +1205,14 @@ class Device(BlueskyInterface, OphydObject):
             # Initial access of signal
             cpt = self._sig_attrs[attr]
         except KeyError:
-            raise AttributeError(f'Component does not exist: {attr}') from None
+            raise RuntimeError(
+                f'Component {attr!r} exists at the Python level and '
+                'the has triggered the `_instantiate_component` '
+                'code path on Device, but has not been registered with '
+                'the Component management machinery.  This maybe due to '
+                'using multiple inheritance with a mix-in class that defines '
+                'a Component but does not inherent from Device.'
+            ) from None
 
         try:
             self._signals[attr] = cpt.create_component(self)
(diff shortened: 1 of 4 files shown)


JunAishima:
can you remove the device.BlueskyInterface import? This is causing the test to fail.

ophyd/device.py
```diff
@@ -1205,7 +1205,14 @@ def _instantiate_component(self, attr):
             # Initial access of signal
             cpt = self._sig_attrs[attr]
         except KeyError:
-            raise AttributeError(f'Component does not exist: {attr}') from None
+            raise RuntimeError(
+                f'Component {attr!r} exists at the Python level and '
+                'the has triggered the `_instantiate_component` '
```
danielballan, line 1210: Missing a word here, can't quite tell which one.

JunAishima: there should be no first "the" in the second line?

prjemian: Not that, still wouldn't make sense.  Perhaps the missing word is _Device_? As in 

> Component {attr!r} exists at the Python level and the Device has triggered the `_instantiate_component`

prjemian: or 

> ... the caller has triggered ...

JunAishima: @tacaswell would you like to fix up the second half-sentence? thanks, will pass the review immediately once it is the way you want it (and looks like a complete sentence)

tacaswell: Y'all knew what I meant...

ophyd/device.py
```diff
@@ -1206,10 +1206,10 @@ def _instantiate_component(self, attr):
             cpt = self._sig_attrs[attr]
         except KeyError:
             raise RuntimeError(
-                f'Component {attr!r} exists at the Python level and '
-                'the has triggered the `_instantiate_component` '
+                f'The Component {attr!r} exists at the Python level and '
+                'has triggered the `_instantiate_component` '
                 'code path on Device, but has not been registered with '
-                'the Component management machinery.  This maybe due to '
+                'the Component management machinery in Device.  This maybe due to '
```
danielballan, line 1212:
```suggestion
                'the Component management machinery in Device.  This may be due to '
```

PR #887, opened 2020-07-21
ScalerCH.select_channels default parameter

### LINKED ISSUE #869: BUG: ScalerCH.select_channels() selects channel with no name text
In https://github.com/BCDA-APS/use_bluesky/issues/56, the name field of the first channel of the scaler was left blank in EPICS.  The `select_channels()` code failed to handle this properly.  Even when that name was set after the bluesky session started, `select_channels()` code still failed to handle this properly.  If the channel had a name before bluesky connected to the scaler, then `select_channels()` code works properly.

Also, now is a good time to make the `chan_names` argument into a kwarg with default of `None` or `[]`.  Do that enhancement in a separate issue (#870).

### LINKED ISSUE #870: ENH: ScalerCH.select_channels() arg should be optional
In the `ScalerCH.select_channels()` method, it would be more intuitive if the `chan_names` argument were optional.  Make this a kwarg with default value of `None` (to match current handling of `chan_names`).

prjemian:
fixes #869 (chan01 was always selected even if empty name) and fixes #870 (makes `None` a default parameter: change the arg into a kwarg)

### DIFF
--- a/ophyd/scaler.py
+++ b/ophyd/scaler.py
@@ -142,7 +142,7 @@ class ScalerCH(Device):
         if chan_names is None:
             chan_names = name_map.keys()
 
-        read_attrs = ['chan01']  # always include time
+        read_attrs = []
         for ch in chan_names:
             try:
                 read_attrs.append(name_map[ch])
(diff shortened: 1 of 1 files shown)


prjemian:
Thanks for the review!

PR #892, opened 2020-08-20
One particular log line in ophyd is too verbose.

danielballan:
This line dominates `bluesky.log` and generated 9 MB of log data in
about a minute at SMI. This should be DEBUG level.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -210,7 +210,7 @@ class Signal(OphydObject):
             Check the value prior to setting it, defaults to False
 
         '''
-        self.log.info(
+        self.log.debug(
             'put(value=%s, timestamp=%s, force=%s, metadata=%s)',
             value, timestamp, force, metadata
         )


ZLLentz:
Thank you for changing this, this log message was very loud at other facilities as well. We had ended up filtering it out.

danielballan:
Good to know. Is there anything else you are filtering out that we should consider changing?

We are _finally_ rolling out ophyd widely at NSLS-II after years stuck on 1.3.x, so we're getting back to "eat our own dogfood" here.

ZLLentz:
We've also completely filtered out the logs from `bluesky`, but I can't find an issue or commit message with the full details as to why, nor can I remember. The level changed here in `ophyd` was by far the noisiest message.

danielballan:
It was briefly the case that bluesky and ophyd registered handlers for themselves at import time, but we quickly thought better of that choice. As of bluesky v1.6.0 if you don't want bluesky logs you should be able to just "do nothing" and none will be handled.

ZLLentz:
Yes, that is the case now. The issue comes when someone (like me) adds a handler to the root logger, expecting things to be just ok (spoilers, this sets you up for guaranteed spam, no matter how many libs you blacklist). We'll probably be switching to a more sustainable model where we add a logging handler to each specific repo that we want the user to see messages from and/or log debug information to files.

danielballan:
That makes sense. Yeah, I think we can agree that both things are true: attaching to the root logger is guaranteed spam, and also ophyd should not be generating ~10 MB/s of log messages at INFO level. :-D

PR #893, opened 2020-08-20
Capture num_images as config.

danielballan:
We should have been capturing `num_images`! I notice that we do capture
`num_exposures`, which is a counter of how many images have been acquired so
far, whereas `num_images` is the total that *will* be acquired in this batch.

We might consider removing `num_exposures` to avoid confusion---I don't think it
very useful to capture it.

### DIFF
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -42,6 +42,7 @@ class CamBase(ADBase):
     _default_configuration_attrs = (ADBase._default_configuration_attrs +
                                     ('acquire_time', 'acquire_period',
                                      'model', 'num_exposures', 'image_mode',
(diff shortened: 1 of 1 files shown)


mrakitin:
Discovered while deploying at SMI, have a workaround in the IPython startup file for now: https://github.com/NSLS-II-SMI/profile_collection/blob/57aa16f293566c6b211d6b56cda721c1d0ec7af6/startup/21-pilatus.py#L138.

danielballan:
I should mention on a public channel that @ronpandolfi independently hit this same problem earlier this week, correctly diagnosed and explained it, which made it easy for me to identify and fix the issue when it arose at NSLS-II today.

PR #897, opened 2020-08-23
DOC fix typo

prjemian:


danielballan:
Good catch. I see the problem, where the descriptions of relative and absolute are flipped. But in fixing that this introduces a new problem because it lists atol and rtol in a different order than they are given in the function signature. Can you flip the descriptions without flipping the order of the parameters?

prjemian:
Aha!  Order of parameter documentation is important.  Got it.

danielballan:
Yep! It is not technically enforced in any way, but it helps clarity for the reader if the order in the Parameter docs matches the order in the signature. 👍

PR #903, opened 2020-09-04
provide float format specifier where previously implied (as integer)

### LINKED ISSUE #902: incorrect format specifier in ophyd.signal._get_with_timeout()
Just got [bitten](https://github.com/APS-USAXS/ipython-usaxs/issues/253#issuecomment-687164423) by a `ValueError: Precision not allowed in integer format specifier` coming from this code:  https://github.com/bluesky/ophyd/blob/6a6ebeaa46afa439d82e1c4cd476fc5255acffa6/ophyd/signal.py#L1057-L1060

Is it missing a float specifier?  I believe so.  Easy fix for two places.

prjemian:
fixes #902

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1044,7 +1044,7 @@ class EpicsSignalBase(Signal):
         except TimeoutError as err:
             raise ConnectionTimeoutError(
                 f"Failed to connect to {pv.pvname} "
-                f"within {connection_timeout:.2} sec") from err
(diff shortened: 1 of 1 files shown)


tacaswell (approved):

tacaswell:
Anyone (including @prjemian ) can merge this when CI is green.  I don't think we need to force a test of the integer code path.

mrakitin (commented):
Good catch! Thanks for the fix.

PR #904, opened 2020-09-07
Redis Signal with Subscriptions.

stmudie:
A Redis Signal based on #850 and the redis signal in production on the SAXS beamline at the Australian Synchrotron. It supports redis subscriptions if the redis server has notify-keyspace-events set to $K or AKE. There is also a Factory Class for returning keys for a specific Redis instance.

My first PR to Ophyd, definitely keen to get some feedback @tacaswell, @danielballan.


danielballan:
Thanks for pushing on this, @stmudie! Will take look in detail later this week.

ophyd/redis_signal.py
```diff
@@ -0,0 +1,292 @@
...
+            callback
+        """
+        events = self._r.config_get()["notify-keyspace-events"]
+        if not search("^(?=.*(A|\$))(?=.*K)", events):
+            raise NoEventNotifications
+
+        if self._pubsub is None:
+            self._pubsub = self._r.pubsub(ignore_subscribe_messages=True)
+
+        self._pubsub.subscribe(**{f"__keyspace@0__:{self._key}": self._callback})
```
tacaswell, line 191: Ah ha!  This is what I missed in my attempt at doing this :+1: .

ophyd/redis_signal.py
```diff
@@ -0,0 +1,292 @@
...
+            If notify-keyspace-events is not set to AK or $K on the redis server.
+
+        Returns
+        -------
+        cid : int
+            id of callback, can be passed to `unsubscribe` to remove the
+            callback
+        """
+        events = self._r.config_get()["notify-keyspace-events"]
+        if not search("^(?=.*(A|\$))(?=.*K)", events):
```
tacaswell, line 185:
```suggestion
        if not search(r"^(?=.*(A|\$))(?=.*K)", events):
```

Python is getting stricter about invalid escape sequences, I assume that '\' needs to be ignored by Python but be seen by the regular expresssion?

ophyd/tests/test_redis_signal.py
```diff
@@ -0,0 +1,42 @@
+import time
+import fakeredis
```
tacaswell, line 2: If you add `fakeredis` to https://github.com/bluesky/ophyd/blob/master/test-requirements.txt it will get automatically installed on travis.

ophyd/tests/test_redis_signal.py
```diff
@@ -0,0 +1,42 @@
...
+    assert signal.get() == value
+    assert signal.timestamp == start_t
+
+    # test put method
+    signal.put(new_value)
+    assert signal.get() == new_value
+    # reset value
+    signal.put(value)
+    assert signal.read()[key]["value"] == value
+    # signal.describe()
```
danielballan, line 29: Might as well un-comment these to exercise the codepaths at least.

ophyd/redis_signal.py
```diff
@@ -0,0 +1,292 @@
...
+            raise NoKey
+
+        self._last_read = self._deserializer(v)
+
+        return {
+            self.name: self._last_read,
+        }
+
+    def describe(self):
+        val = self.read()
```
danielballan, line 143: Should this use `self._last_read` (assume it's been read once) or force a read every time?

ophyd/redis_signal.py
```diff
@@ -0,0 +1,292 @@
...
+        self._run_subs(
+            sub_type=self.SUB_VALUE,
+            old_value=self._last_read["value"],
+            value=read["value"],
+            timestamp=read["timestamp"],
+        )
+
+
+class RedisSignalFactory:
+    """Factory to return RedisSignals"""
```
danielballan, line 230: Can elaborate more on the use case for this?

mrakitin:
@stmudie, @tacaswell, can we consider this PR overriding https://github.com/bluesky/ophyd/pull/850?

mrakitin:
Do you mind rebasing, or let me rebase it on `master` so that it could pick up the GHA-based CI testing? Thanks!

PR #907, opened 2020-09-12
Correct documentaiton about SubscriptionStatus callback signature.

danielballan:
When the subscriptions are run, `value` and `old_value` are always passed in by keyword, not positionally. This happens in the ophyd layer and is not dependent on the control layers (pyepics, caproto) so we can be sure of it. Thus, their order does not matter.  This created a recent concern when it was discovered that some callbacks had signature `f(value, old_value, **kwargs)` while other had signature `f(old_value, value, **kwargs)`. The documentation should be clear on this point, and might as well require `f(*, value, old_value, **kwargs)` to be future-proof against some caller than attempts to pass in arguments positionally.

### DIFF
--- a/doc/source/status.rst
+++ b/doc/source/status.rst
@@ -148,7 +148,7 @@ is marked finished based on some ophyd event. It reduces this:
        acquire = Component(...)
 
        def trigger(self):
-           def check_value(old_value, value, **kwargs):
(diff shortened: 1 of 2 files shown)


danielballan:
Thanks to @jklynch and @johnsinsheimer for flagging this problem.

jklynch (approved):

PR #912, opened 2020-09-25
Add missing return  value in test IOC.

danielballan:
This bug was introduced recently in #749. It was my oversight.


mrakitin (approved):
It should not affect the production code, so no immediate upgrade is needed. But it is good to have the tests fixed.

danielballan:
Right. Also, the bug hasn’t been in a release yet.

Issue #911, opened 2020-09-25
NDFile PVs should be implemented in FileBase instead of FilePlugin


jwlodek:
I am working on adding `ADEiger` support into ophyd - currently it seems that each beamline have their own implementation in their `profile_collection` repository. While working on this, I think I noticed a minor limitation in the current inheritance structure for file plugins. Essentially, a file plugin consists of `NDPluginBase` PVs, and `NDFile` PVs. The `NDFile` template is not only used for file plugins however, and some high speed detectors (like `ADEiger`) support API operations for file saving aside from the areaDetector process (See the link below)

https://github.com/areaDetector/ADEiger/blob/01bb57cc2d1e3d78fbf56f88c38406be3e61574e/eigerApp/Db/eiger.template#L6

This should really be supported by ophyd, since in some cases these detectors can only run in their highest speed modes when using the internal API writing, since that handles asynchronous I/O so the detector can continue to collect data without waiting for image saving to complete - as opposed to the plugins which can only do this up to a max queue size that is usually limited to ~20  frames. Basically, I'm proposing to create a new class `FileBase`, and then have `FilePlugin` extend `PluginBase` and `FileBase`, rather than define the PV signals directly in the `FilePlugin` class. Thoughts?

jwlodek:
Closed via https://github.com/bluesky/ophyd/pull/913

PR #913, opened 2020-09-28
WIP - Introduce new FileBase class for use with FilePlugin + detectors that have direct disk write access

jwlodek:
This is a work in progress PR that tries to address the issue I described here: https://github.com/bluesky/ophyd/issues/911

Essentially, certain detectors use the typical `NDFile` PVs to perform direct I/O via the vendor API rather than the areaDetector plugins - this is often the case for max-speed modes on high speed detectors (ex. Eiger). As a result, I broke off most of the signals defined by `FilePlugin` into a new `FileBase` class, and had `FilePlugin` extend from it. Then, if detectors need, they could extend the `FileBase` class. 

I am marking this as a WIP, since I would like some feedback, and to perform some real world tests.

### DIFF
--- a/ophyd/areadetector/plugins.py
+++ b/ophyd/areadetector/plugins.py
@@ -61,6 +61,7 @@ __all__ = [
     'Overlay',
     'OverlayPlugin',
     'PluginBase',
+    'FileBase',
     'PosPlugin',
     'ProcessPlugin',
     'PvaPlugin',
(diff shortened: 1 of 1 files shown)


tacaswell (approved):

ophyd/areadetector/plugins.py
```diff
@@ -776,10 +777,8 @@ class TransformPlugin(PluginBase, version=(1, 9, 1), version_type='ADCore'):
     )
 
 
-class FilePlugin(PluginBase, GenerateDatumInterface, version=(1, 9, 1), version_type='ADCore'):
+class FileBase:
```
tacaswell, line 780: This class needs to inherit from `Device` so that the `Cpt` objects are registered (which is done via the subclass init hook).

jwlodek:
Seems Travis is happier with the extra change

danielballan (approved):
Self-merge at will if you feel this is ready, @jwlodek.

jwlodek:
It seems I don't have permissions to self-merge, at least not that I can see

Issue #889, opened 2020-08-13
Add convenience functions for generating boilerplate area detector classes.


danielballan:
Currently, even the simplest area detector configurations require defining custom classes in beamlines' IPython profiles. There is no full-featured AD class built in to ophyd.

This makes setting up an AD a fairly intensive process that requires knowledge of details of both object-oriented Python and ophyd beyond what's required to set up an `ophyd.EpicsMotor`. Notice the use of multiple inheritance to configure triggering and the need for a custom file plugin class to provide integration with "assets" (Resource and Datum documents).

**Some of this is irreducible complexity --- area detectors are more complex than motors! --- but not _all_ of it.**

(Source for this example is https://github.com/NSLS-II-CHX/profile_collection/blob/master/startup/20-area-detectors.py. Go there for full details.)

```py
# imports omitted for brevity

class TIFFPluginWithFileStore(TIFFPlugin, FileStoreTIFFIterativeWrite):
    """Add this as a component to detectors that write TIFFs."""
    pass

class StandardProsilicaV33(SingleTriggerV33, ProsilicaDetector):
    cam = Cpt(ProsilicaDetectorCamV33, 'cam1:')
    image = Cpt(ImagePlugin, 'image1:')
    stats1 = Cpt(StatsPluginV33, 'Stats1:')
    stats2 = Cpt(StatsPluginV33, 'Stats2:')
    stats3 = Cpt(StatsPluginV33, 'Stats3:')
    stats4 = Cpt(StatsPluginV33, 'Stats4:')
    stats5 = Cpt(StatsPluginV33, 'Stats5:')
    trans1 = Cpt(TransformPlugin, 'Trans1:')
    roi1 = Cpt(ROIPlugin, 'ROI1:')
    roi2 = Cpt(ROIPlugin, 'ROI2:')
    roi3 = Cpt(ROIPlugin, 'ROI3:')
    roi4 = Cpt(ROIPlugin, 'ROI4:')
    proc1 = Cpt(ProcessPlugin, 'Proc1:')
    over1 = Cpt(OverlayPlugin, 'Over1:')

class StandardProsilicaWithTIFFV33(StandardProsilicaV33):
    tiff = Cpt(TIFFPluginWithFileStore,
               suffix='TIFF1:',
               write_path_template='/XF11ID/data/%Y/%m/%d/',
               root='/XF11ID/data')
```

The result is a lot of cargo-culted copy/pasted code across profiles. To improve the situation should be:

[...]

Issue #888, opened 2020-08-12
Refactor triggering in area detector to simplify it


danielballan:
The implementation of area detector triggering is one of the oldest parts of ophyd. It (re-)developed when we had the least experience with this newly-conceived hardware abstraction. It allows much more flexibility that we actually have needed in practice. Knowing that, we can massively simplify it.

## Current implementation

WARNING: Thar be :dragon: s!

When an area detector is triggered, we dynamically add items to the dictionary that will be returned by `describe()` and `read()`. In practice, we add one item, typically named something like "fccd_image" or "eiger1m_image". We also provide a timestamp at this point. The line in this snippet is `self._dispatch(...)`.

https://github.com/bluesky/ophyd/blob/22089ca1795dd41b214c3589fb25af84f644356d/ophyd/areadetector/trigger_mixins.py#L129-L138

We also allow one `trigger()` to add _multiple_ items to `describe()` and `read()`, with unique names of course. This was designed to support a range of readings in one Event with different gains, or with light and dark frames. In practice, we long ago found better, simpler ways address these use cases.

https://github.com/bluesky/ophyd/blob/22089ca1795dd41b214c3589fb25af84f644356d/ophyd/areadetector/trigger_mixins.py#L221-L263

All currently-enabled file plugins are told to generate a datum and add it to an internal cache that will be merged with the return of the items returned by `read()`. That's right, we support the possibility that the detector is running with multiple file plugins, like HDF5 _and_ TIFF. This has never been needed in practice.

https://github.com/bluesky/ophyd/blob/bc9770f19c16bda79df12049242db5c2583690df/ophyd/areadetector/detectors.py#L49-L70

https://github.com/bluesky/ophyd/blob/22089ca1795dd41b214c3589fb25af84f644356d/ophyd/areadetector/filestore_mixins.py#L349-L396

Notice the lock `self._locked_key_list` that ensures that `describe()` and `read()` always return the same keys _within a given stage/unstage cycle_. It would be better if they returned the same keys _always_. This is the only place is ophyd where a Device dynamically adds keys to the dictionaries returned by `describe()` and `read()` or where they contain values for which there is no corresponding `Component`. It would be great to remove this exception from the codebase.

## Proposed implementation

Rather than dynamically adding keys to the `describe()` and `read()` dict, add soft-signal (`ophyd.Signal`) `Component` to the Device when can be used to stash the `datum_id` that goes with the current reading.

[...]

PR #919, opened 2020-10-06
Add classes for Eiger detector and cam

jwlodek:


ophyd/areadetector/detectors.py
```diff
@@ -3,7 +3,7 @@
 
 `areaDetector`_ detector abstractions
 
-.. _areaDetector: http://cars.uchicago.edu/software/epics/areaDetector.html
+.. _areaDetector: https://areadetector.github.io/master/index.html
```
danielballan, line 6: :+1:

ophyd/areadetector/cam.py
```diff
@@ -299,6 +303,80 @@ class EmergentVisionDetectorCam(CamBase):
     auto_gain = ADCpt(SignalWithRBV, 'EVTAutoGain')
 
 
+class EigerDetectorCam(CamBase, FileBase):
```
danielballan, line 306: Thanks!

Have you cross-referenced this with the implementation hard-coded into CHX's profile? I would bet that the implementation in this PR is more thoughtfully composed, but it would be good to know if it is a strict superset of the Components included there. Ideally we would move CHX to rely on this one instead.

https://github.com/NSLS-II-CHX/profile_collection/blob/7ef674a1b399f26304ce5342896d23680acbf878/startup/20-area-detectors.py#L160-L192

Also, that Eiger has a special file-writing functionality that is unlike the normal Area Detector file plugins. We wrote a one-off ophyd class to shoe-horn it into looking like ophyd's concept of a file plugin. 

https://github.com/NSLS-II-CHX/profile_collection/blob/7ef674a1b399f26304ce5342896d23680acbf878/startup/20-area-detectors.py#L117-L157

Would that be useful upstream? Do _all_ Eigers have this custom file-writing functionality? I have never been clear on that; we just dashed this CHX implementation out under pressure in very early days and haven't circled back to carefully consider it.

jwlodek: I started with the CHX profile collection implementation as a base (the variable names are the same for the Signals shared with that implementation), so it should be a superset of that class. I just added signals for all the remaining PVs defined by the Eiger driver.

The file saving in Eiger that doesn't use the default file plugins is common to all Eiger detectors. Basically, since the detector is too fast to follow the normal plugin chain, the vendor API writes to a super-fast cache first, and then this built in file plugin in the eiger driver asynchronously copies the saved hdf5 files from the cache to the target location specified by the filepath PV.

This actually isn't limited to just the Eiger driver - there are several detectors that use File PVs to control built in vendor API file saving operations, skipping the typical areaDetector plugin chain. The giveaway is basically if the `NDFile.template` file is included in the driver database definition:

https://github.com/areaDetector/ADEiger/blob/01bb57cc2d1e3d78fbf56f88c38406be3e61574e/eigerApp/Db/eiger.template#L6

Basically to account for this, in a previous PR I had split the definition of the PVs defined in `NDFile.template` out from the `FilePlugin` class into the new `FileBase` class, and these signals should cover you for the one-off file plugin wrapper signals - they will now just be under the `EigerDetectorCam` umbrella, so that one-off class should theoretically be no longer needed. Instead, the stage/unstage code should probably be moved into a subclass of `EigerDetector`.

I think that overall what would be the most beneficial here in upstream is code to avoid the boilerplating required to create a usable ophyd object for an areaDetector, I believe described in an issue here: https://github.com/bluesky/ophyd/issues/889 and https://github.com/bluesky/ophyd/issues/888

Generally all areaDetectors use the same plugin format, so maybe helper functions that automatically add all the plugins loaded by default by [`commonPlugins.cmd`](https://github.com/areaDetector/ADCore/blob/master/iocBoot/EXAMPLE_commonPlugins.cmd) would be a nice start?

danielballan: Great. Thanks for explaining; I hadn't yet learned some of that context.

I suspected that your `FileBase` PR was laying track for this exact scenario but didn't want to assume.

I agree that would be a nice start. I am hopeful that DAMA can prioritize #888 and #889 soon, but if you have motivation and time to take it on yourself before then, please don't hesitate to claim one or both.

ophyd/areadetector/docs.py
```diff
@@ -2680,7 +2680,11 @@
           Toggles AutoGain mode for EVT camera. This will have the camera automatically adjust the gain rather than using the 
           user adjustable Gain value.
         ''',
-    ),  # end of EVTDoc.html
+        ),  # end of EVTDoc.html
+
+    'EigerDoc.html' : dict(
```
tacaswell, line 2685: @klauer Was there a script to scrape the contents of this file out of the upstream docs?

tacaswell (approved):

tacaswell:
@jwlodek It looks like this needs a rebase due to conflicts in the release notes.

danielballan:
Rebased, and addressed whitespace changes. Anyone should merge at will when CI passes. Targeting 1.6.0 release this week.

danielballan:
Thanks agian, @jwlodek.

PR #920, opened 2020-10-08
add Xspress3Detector and Xspress3DetectorCam

jklynch:
This PR adds classes for the xspress3 using areaDetector version 3+.

### DIFF
--- a/ophyd/areadetector/cam.py
+++ b/ophyd/areadetector/cam.py
@@ -1,4 +1,3 @@
-from enum import IntEnum, unique
 import logging
 
 from ..utils import enum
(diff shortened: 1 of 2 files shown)


mrakitin (approved):
Looks good to me, thanks @jklynch.

For the context, here is @newville's slides presented at NSLS-II - https://millenia.cars.aps.anl.gov/software/xspress3/Xspress3_Oct2020.pdf, and this class is for the corresponding AD.

newville:
@jklynch you might find the code at https://github.com/pyepics/epicsscan/blob/master/lib/detectors/xspress3.py useful for how we use the xspress3 with our data collection system. This is probably somewhat different from ophyd's model (fwiw, this code uses `pyepics.Device` and pre-dates `ophyd` - I have no plans on ).  

Heres, each Detector contains a Trigger and a list of Counters.  In addition, each Detector has a number of modes:

 -  ContinuousMode, for running the detector in "free run" mode
 -  ScalerMode, for step-count-step-count scans.
 -  NDArrayMode, or mapping mode: external triggers advance the detector frame, with AD FileSaver expected to save files.
 -  ROIMode, for triggered scanning where you want to save only a handful of ROIs, as with ROIStats.  Data has to be read out from the Counter PVs for saving.

Each detector has a `pre_scan()` method to prepare for a scan. This would set the mode, set which Counters are used (say, which ROIs and SCAs to save), paths for the file-saver, etc, and a `post_scan()` method which will generally do something like put the detector into ContinuousMode.

Each detector also has `arm()`, `start()`, and `stop()` method.   A scan will basically go like
   for thing in detectors+positioners:
        thing.pre_scan(**scan_info)
   scan(**scan_info)
   gather_and_write_data(**scan_info)
   for thing in detectors+positioners:
        thing.post_scan(**scan_info)

The `scan()` method there would be different for the different modes so there can be a difference in how often the detector has to be `arm()`ed before `start()`.

Anyway, that's mostly to just explain why the code is arranged the way it is.  I don't know anything about how one specifies trajectory or fly scans with ophyd/bluesky.

danielballan:
What is the relationship between this and https://github.com/NSLS-II/nslsii/blob/master/nslsii/detectors/xspress3.py?

jklynch:
The intention with the new Xspress3Detector is to work with an up-to-date version of AreaDetector and to look more like the other AreaDetector classes. I am not familiar with the nslsii Xspress3Detector (my bad). It looks like it may not line up perfectly with the new Xspress3 IOC but I may be wrong about that.

Would it make sense to keep the nslsii code as-is or try to update it to work with the current AreaDetector? Maybe it is best to make the AreaDetector version explicit somehow?

danielballan:
Yes, I think so. Maybe @mrakitin or @tacaswell could comment on why that lives in `nslsii` as opposed to ophyd. Does NSLS-II run a particular facility-specific variant of the IOC? If not, I imagine we should move that upstream into ophyd and continue developing and updating it here.

It looks like that implementation provides a lot of functionality beyond your basic "Add a couple device-specific PVs."

jklynch:
In partial response to @danielballan it may be that some code, in particular the `channel` code, can work with both the "old" and "new" xspress3.

ophyd/areadetector/cam.py
```diff
@@ -765,3 +767,43 @@ class URLDetectorCam(CamBase):
...
+        TTL_BOTH = 4
+        LVDS_VETO_ONLY = 5
+        LVDS_BOTH = 6
+
+    """
+    The ERASE PV takes two integer values:
+        0 - Done
+        1 - Erase
+    """
+    erase = ADCpt(EpicsSignal, "ERASE")
```
tacaswell, line 798:
```suggestion
    erase = ADCpt(EpicsSignal, "ERASE", kind='omitted')
```

tacaswell:
The version in nslsii (which was extracted from hxntools) has a bunch of additional PVs https://github.com/NSLS-II/nslsii/blob/60a8e75410faa40573491774c4f92e506178ad93/nslsii/detectors/xspress3.py#L246-L286 who's names suggest they are relevant.  Did those PVs fail to make the migration to the new IOC or are we not worried about them?

jklynch:
I have added the PVs from the nslsii version with the exception of `ctrl_mca_roi` because I could not find it in https://github.com/epics-modules/xspress3/blob/master/xspress3App/Db/xspress3.template.

The `ctrl_mca_roi` PV shows up in the documentation but nowhere else (?) https://github.com/epics-modules/xspress3/blob/62ee9bb58997cfb7b9d870fe178bc2ffab8390f3/etc/makeDocumentation/manual.src#L97

PR #922, opened 2020-10-16
Enh use monitor

tacaswell:
This is code notes from a discussion with @prjemian and @danielballan which does to seperate things:

 - Tell EpicsMotor to automonitor more things (which should drastically speed up `mtr.read_configuration()`
 - re-threads `EpicsSignal.get(..., use_monitor=True)` back through to the control layer layer + sets a per-Signal default (which tracks auto-monitor).

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1166,7 +1166,8 @@ class EpicsSignalBase(Signal):
         '''
         if kwargs:
             warnings.warn('Signal.get no longer takes keyword arguments; '
-                          'These are ignored and will be deprecated.')
+                          'These are ignored and will be deprecated.',
+                          DeprecationWarning)
         if as_string is None:
             as_string = self._string
 
@@ -1592,7 +1593,8 @@ class EpicsSignal(EpicsSignalBase):
         '''
         if kwargs:
             warnings.warn('Signal.get_setpoint no longer takes keyword arguments; '
-                          'These are ignored and will be deprecated.')
+                          'These are ignored and will be deprecated.',
+                          DeprecationWarning)
         if as_string is None:
             as_string = self._string
 
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -737,10 +737,12 @@ class EpicsSignalBase(Signal):
                        )
                       )
 
-    def __init__(self, read_pv, *, string=False,
+    def __init__(self, read_pv, *,
+                 string=False,
                  auto_monitor=DEFAULT_AUTO_MONITOR,
                  name=None,
-                 metadata=None, all_pvs=None,
+                 metadata=None,
+                 all_pvs=None,
                  timeout=DEFAULT_TIMEOUT,
                  write_timeout=DEFAULT_WRITE_TIMEOUT,
                  connection_timeout=DEFAULT_CONNECTION_TIMEOUT,
@@ -1092,7 +1094,7 @@ class EpicsSignalBase(Signal):
         return (self._metadata['lower_ctrl_limit'],
                 self._metadata['upper_ctrl_limit'])
 
-    def _get_with_timeout(self, pv, timeout, connection_timeout, as_string, form):
+    def _get_with_timeout(self, pv, timeout, connection_timeout, as_string, form, use_monitor):
         """
         Utility method implementing a retry loop for get and get_setpoint
 
@@ -1119,7 +1121,9 @@ class EpicsSignalBase(Signal):
             'pv[%s].get_with_metadata(as_string=%s, form=%s, timeout=%s)',
             pv.pvname, as_string, form, timeout
         )
-        info = pv.get_with_metadata(as_string=as_string, form=form, timeout=timeout)
+        info = pv.get_with_metadata(
+            as_string=as_string, form=form, timeout=timeout, use_monitor=use_monitor
+        )
         self.control_layer_log.debug('pv[%s].get_with_metadata(...) returned', pv.pvname)
 
         if info is None:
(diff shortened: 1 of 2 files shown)


danielballan (commented):
Seems right, as long as CI is happy. Do the new warnings get tripped by our test suite? I worry a little about generating a slew of user-facing warnings that it will take us some time to get around to fixing. Perhaps we could start by using a `DeprecationWarning`, not shown by default in user space.

prjemian:
Simple test code with EPICS motor record instance (running in a docker container on the same workstation, BTW): `sky:m1`

```python
#!/usr/bin/env python3

from ophyd import EpicsMotor
import time

m1 = EpicsMotor("sky:m1", name="m1")
m1.wait_for_connection()
m1.read_configuration()

def test(count=1):
    for _i in range(count):
        m1.read_configuration()

n = 10000
t0 = time.time()
test(n)
dt = (time.time() - t0) / n
print(f"{n} times, {dt*1000:.3f}ms per time")
```

### ophyd origin master

```
(bluesky_2020_9) prjemian@poof ~/.../Bluesky/ophyd $ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
(bluesky_2020_9) prjemian@poof ~/.../Bluesky/ophyd $ /tmp/test_automonitor.py 
10000 times, 1.611ms per time
```

### ophyd  tacaswell:enh_use_monitor

```
(bluesky_2020_9) prjemian@poof ~/.../Bluesky/ophyd $ git checkout enh_use_monitor 
Switched to branch 'enh_use_monitor'
Your branch is up-to-date with 'tom/enh_use_monitor'.
(bluesky_2020_9) prjemian@poof ~/.../Bluesky/ophyd $ /tmp/test_automonitor.py 
10000 times, 0.210ms per time
```

Convincing!
Is this demo sufficiently robust?  Needs to test at a beam line that experiences a read timeout error (APS 8-ID-I or 9-ID-C).

prjemian:
The speedup is an order of magnitude faster using values cached from CA monitors.

prjemian (commented):
Works on simple test as shown in comments.  Needs tests at beam lines that report the original problem.

prjemian:
For comparison, test the `ophyd.sim.motor` simulated motor:

```
In [1]: from ophyd.sim import motor as m1                                                                                                          

In [2]: import time                                                                                                                                

In [3]: m1.wait_for_connection() 
   ...: m1.read_configuration()                                                                                                                    
Out[3]: 
OrderedDict([('motor_velocity', {'value': 1, 'timestamp': 1603820951.6882622}),
             ('motor_acceleration',
              {'value': 1, 'timestamp': 1603820951.6883285})])

In [4]: def test(count=1): 
   ...:     for _i in range(count): 
   ...:         m1.read_configuration() 
   ...:  
   ...: n = 10000 
   ...: t0 = time.time() 
   ...: test(n) 
   ...: dt = (time.time() - t0) / n 
   ...: print(f"{n} times, {dt*1000:.3f}ms per time")                                                                                              
10000 times, 0.018ms per time

In [5]: m1                                                                                                                                         
Out[5]: SynAxis(prefix='', name='motor', read_attrs=['readback', 'setpoint'], configuration_attrs=['velocity', 'acceleration'])

In [6]:                                                                                                                                            
```

This is an order of magnitude faster than the CA monitor case.  Why?  `SynAxis.get()` has a simpler call stack than `EpicsMotor.get()`?

tacaswell:
> The speedup is an order of magnitude faster using values cached from CA monitors.

An I suspect you would get an extra order of magnitude if you were talking to an IOC on an actual remote host due to network round-trip times (we have seen 10-20ms `read_configuration` times on the floor!).

> This is an order of magnitude faster than the CA monitor case. Why?

It `%prun` in IPython should get you a nice line profile, my knee-jerk guess is that even with this change we are still descending into the pyepics call stack where as with `SynAxis` everything in "soft" signals and does not call deeper than the Ophyd methods.

prjemian:
I can test run with an IOC on a remote host.  Am running a long test on an APS host now.  Next will be to pick either USAXS (more likely) or XPCS (instrument in use by SPEC users now).

prjemian:
@tacaswell The PV read timeout problem happens with EpicsSignal and EpicsSignalRO objects, as well.  Can we make a `@classmethod` to change the default `auto_monitor=True` for `EpicsSignalBase`?

ophyd/signal.py
```diff
@@ -1123,12 +1130,20 @@ def get(self, *, as_string=None,
             for the connection to complete.
         form : {'time', 'ctrl'}
             PV form to request
+
         '''
+        if kwargs:
+            warnings.warn('Signal.get no longer takes keyword arguments; '
+                          'These are ignored and will be deprecated.')
```
danielballan, line 1137:
```suggestion
                          'These are ignored and will be deprecated.',
                          DeprecationWarning)
```

ophyd/signal.py
```diff
@@ -1542,11 +1560,18 @@ def get_setpoint(self, *, as_string=None,
         form : {'time', 'ctrl'}
             PV form to request
         '''
+        if kwargs:
+            warnings.warn('Signal.get_setpoint no longer takes keyword arguments; '
+                          'These are ignored and will be deprecated.')
```
danielballan, line 1565:
```suggestion
                          'These are ignored and will be deprecated.',
                          DeprecationWarning)
```

danielballan (commented):
Let's do at least a cycle of `DeprecationWarning` to give us a chance to catch up before we start spamming user-space.

tacaswell:
rebased and ready for review.

ophyd/signal.py
```diff
@@ -737,10 +737,12 @@ class EpicsSignalBase(Signal):
                        )
                       )
 
-    def __init__(self, read_pv, *, string=False,
+    def __init__(self, read_pv, *,
+                 string=False,
```
prjemian, line 741: Why this addition? (`string=False`)

prjemian: Oh, not addition, just reformatting.

prjemian (approved):
LGTM

danielballan (approved):

prjemian:
@danielballan : In time for the 1.6.0 release?

danielballan:
Yes. I hope we’ll tag 1.6.0 Monday. One issue left, plus updating the release notes PR.

PR #926, opened 2020-10-27
EpicsSignal.set() timeout should default to EpicsSignal.timeout

### LINKED ISSUE #925: Why default `timeout` kwarg in EpicsSignal `.set()` not using signal's `.timeout`?
Why the `timeout` kwarg in EpicsSignal `.set()` method does not default to the `.timeout` value?  Looking here:  https://github.com/bluesky/ophyd/blob/3dddf111946c1eb7200361d2158445db2acfb76d/ophyd/signal.py#L1719

NOTE: The PV in use here is the `.STUP` (status update) field of an EPICS motor record.  Set it to 1, "ON" and the record support will set it back to 0 "OFF" as the motor status is updated.  Thus, the ophyd `.set` operation will not succeed since the PV will never reach the "ON" value.

This `.set()` will default the `timeout` kwarg to `None`.  Since I've set the default timeout for the signal to be 10 s, I was expecting `s.timeout` to be 10 (and it is).  

```
(bluesky_2020_9) prjemian@poof /tmp $ ipython
Python 3.8.2 (default, Mar 26 2020, 15:53:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from ophyd.signal import EpicsSignalBase                                                                                  

In [2]: EpicsSignalBase.set_default_timeout(timeout=10, connection_timeout=5)                                                     

In [3]: from ophyd import EpicsSignal                                                                                             

In [4]: s = EpicsSignal("sky:m1.STUP", name='s', string=True)                                                                     

In [5]: s.timeout                                                                                                                 
Out[5]: 10
```

Now, try `st = s.set("ON")` and wait 15 seconds (5 seconds beyond timeout) before printing the status.  Expecting it be done.

```
import time
st = s.set("ON")
time.sleep(15)
print(st)
```

Expecting `done`, but no.

```
In [6]: import time 
   ...: st = s.set("ON") 
   ...: time.sleep(15) 
   ...: print(st) 
   ...:                                                                                                                           
Status(obj=EpicsSignal(read_pv='sky:m1.STUP', name='s', value='OFF', timestamp=1603771932.302131, auto_monitor=False, string=True, write_pv='sky:m1.STUP', limits=False, put_complete=False), done=False, success=False)
```

---

Repeat but use a `timeout=10` kwarg with `.set()` in a new ipython session:

```
In [1]: from ophyd import EpicsSignal 
   ...: import time 
   ...: s = EpicsSignal("sky:m1.STUP", name='s', string=True) 
   ...: st = s.set("ON", timeout=10) 
   ...: time.sleep(15) 
   ...: print(st) 
   ...:                                                                                                                           
set_and_wait(value=ON, timeout=10, atol=None, rtol=None)
Status(obj=EpicsSignal(read_pv='sky:m1.STUP', name='s', value='OFF', timestamp=1603772109.341146, auto_monitor=False, string=True, write_pv='sky:m1.STUP', limits=False, put_complete=False), done=True, success=False)

```

That shows the expected sequence.  Signal is set, status timeout reached and reported, final print shows status is done but not successful.

_Originally posted by @prjemian in https://github.com/bluesky/ophyd/pull/901#issuecomment-716968738_

prjemian:
Changes to fix #925.  Expected the `EpicsSignal.set()` method to use the signal's `timeout` attribute.  This PR will make that happen.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -714,7 +714,8 @@ class EpicsSignalBase(Signal):
 
     # See set_default_timeout for more on these.
     __default_connection_timeout = 1.0
-    __default_timeout = 2.0
+    __default_timeout = 2.0         # *read* timeout
+    __default_write_timeout = None  # Wait forever.
 
     _read_pv_metadata_key_map = dict(
         status=('status', AlarmStatus),
@@ -755,14 +756,7 @@ class EpicsSignalBase(Signal):
             timeout = self.__default_timeout
         self._timeout = timeout
         if write_timeout is DEFAULT_WRITE_TIMEOUT:
-            # This is very different than the connection and read timeouts
-            # above. It relates to how long an action takes to complete. Any
-            # default value we choose here is likely to cause problems---either
-            # by being too short and giving up too early on a lengthy action or
-            # being too long and delaying the report of a failure.
-            # The important thing it is that it is configurable at per-Signal
-            # level via the write_timeout parameter.
-            write_timeout = None  # Wait forever.
+            write_timeout = self.__default_write_timeout
         self._write_timeout = write_timeout
 
         if name is None:
@@ -821,7 +815,8 @@ class EpicsSignalBase(Signal):
         cls.__any_instantiated = True
 
     @classmethod
-    def set_default_timeout(cls, *, timeout=2.0, connection_timeout=1.0):
+    def set_default_timeout(cls, *, timeout=2.0, connection_timeout=1.0,
+                            write_timeout=None):):
         """
         Set the class-wide defaults for timeouts
 
@@ -849,6 +844,8 @@ class EpicsSignalBase(Signal):
         connection_timeout: float, optional
             Time (seconds) allocated for establishing a connection with the
             IOC.
+        write_timeout: float, optional
+            Time (seconds) allocated for writing, not including connection time.
 
         Raises
         ------
@@ -864,6 +861,12 @@ class EpicsSignalBase(Signal):
                 "with the same default retry setting in place.")
         cls.__default_connection_timeout = connection_timeout
         cls.__default_timeout = timeout
+        # The write_timeout is very different than the connection and read timeouts
+        # above. It relates to how long an action takes to complete. Any
+        # default value we choose here is likely to cause problems---either
+        # by being too short and giving up too early on a lengthy action or
+        # being too long and delaying the report of a failure.
+        cls.__default_write_timeout = write_timeout
 
     # TODO Is there a good reason to prohibit setting these three timeout
     # properties?
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -888,8 +888,12 @@ class EpicsSignalBase(Signal):
         # properties?
 
     @classmethod
-    def set_default_timeout(cls, *, **kwargs):
-        warnings.warn('Use set_defaults() instead.  Command ignored.')
+    def set_default_timeout(cls, **kwargs):
+        warnings.warn(
+            "set_default_timeout() will be removed "
+            "in a future release. Use set_defaults() instead."
+        )
+        cls.set_defaults(**kwargs)
 
     @property
     def connection_timeout(self):
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -287,19 +287,19 @@ class Signal(OphydObject):
             except TimeoutError:
                 success = False
                 self.log.warning(
-                    'set_and_wait(name=%s, value=%s, timeout=%s, atol=%s, rtol=%s)',
+                    '%s: set_and_wait(value=%s, timeout=%s, atol=%s, rtol=%s)',
                     self.name, value, timeout, self.tolerance, self.rtolerance
                 )
             except Exception:
                 success = False
                 self.log.exception(
-                    'set_and_wait(name=%s, value=%s, timeout=%s, atol=%s, rtol=%s)',
+                    '%s: set_and_wait(value=%s, timeout=%s, atol=%s, rtol=%s)',
                     self.name, value, timeout, self.tolerance, self.rtolerance
                 )
             else:
                 success = True
                 self.log.info(
-                    'set_and_wait(name=%s, value=%s, timeout=%s, atol=%s, rtol=%s) succeeded => %s',
+                    '%s: set_and_wait(value=%s, timeout=%s, atol=%s, rtol=%s) succeeded => %s',
                     self.name, value, timeout, self.tolerance, self.rtolerance, self._readback)
 
                 if settle_time is not None:
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -27,7 +27,8 @@ if __name__ != 'ophyd.signal':
     )
 
 
-# Sentinels used for default values; see set_default_timeout below for details.
+# Sentinels used for default values; see set_defaults() below for details.
+DEFAULT_AUTO_MONITOR = object()
 DEFAULT_CONNECTION_TIMEOUT = object()
 DEFAULT_TIMEOUT = object()
 DEFAULT_WRITE_TIMEOUT = object()
@@ -681,7 +682,7 @@ class EpicsSignalBase(Signal):
         only applied if the PV is connected within connection_timeout (below).
 
         The default value DEFAULT_TIMEOUT means, "Fall back to class-wide
-        default." See EpicsSignalBase.set_default_timeout to configure class
+        default." See EpicsSignalBase.set_defaults to configure class
         defaults.
 
         Explicitly passing None means, "Wait forever."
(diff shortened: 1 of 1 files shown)


prjemian:
See #925 for details.

prjemian:
Also, it would aid the diagnostics **immensely** if these reports included the signal's name: https://github.com/bluesky/ophyd/blob/3dddf111946c1eb7200361d2158445db2acfb76d/ophyd/signal.py#L284-L303
Since they are inside the `.set()` method, there is easy access to that information.

ophyd/signal.py
```diff
@@ -1711,6 +1711,9 @@ def set(self, value, *, timeout=None, settle_time=None):
         --------
         Signal.set
         '''
+        if timeout is None:
+            timeout = self.timeout
```
tacaswell, line 1715: In `put` we use `self.write_timeout` instead of `self.timeout`, we should probably be consistent with that here.

The other "fix" may be to change the default value of timeout to be `timeout=DEFAULT_WRITE_TIMEOUT` like put (as I assume it eventually falls through to that).

ophyd/signal.py
```diff
@@ -287,20 +287,20 @@ def set_thread():
             except TimeoutError:
                 success = False
                 self.log.warning(
-                    'set_and_wait(value=%s, timeout=%s, atol=%s, rtol=%s)',
-                    value, timeout, self.tolerance, self.rtolerance
+                    'set_and_wait(name=%s, value=%s, timeout=%s, atol=%s, rtol=%s)',
```
tacaswell, line 290: :+1:

tacaswell: one concern is that I don't think we can actually pass `name` to `set_and_wait`, but I like the idea.

ophyd/signal.py
```diff
@@ -1711,6 +1711,9 @@ def set(self, value, *, timeout=None, settle_time=None):
         --------
         Signal.set
         '''
+        if timeout is None:
```
tacaswell, line 1714: This sentenel should now check for `DEFAULT_WRITE_TIMEOUT`?

ophyd/signal.py
prjemian: In _this particular_ case, `self.name` is accessible since the log comes from an inner (?what's the right name for this type of thingy?) function and has access to all its parent's attributes.

ophyd/signal.py
prjemian: Ok, now I understand this usage pattern.  `DEFAULT_WRITE_TIMEOUT` is a defined object, different from `None` but meant to show this value is not defined yet, right?

ophyd/signal.py
prjemian: When I check my test case, `self.write_timeout` is `None` while `self.timeout` is as defined (by me) via `EpicsSignalBase.set_default_timeout(timeout=10, connection_timeout=5)`.

So we need a second test for `None`ness.

prjemian:
```
In [6]: s.write_timeout is None                                                                                                                    
Out[6]: True

In [7]: s.timeout                                                                                                                                  
Out[7]: 10
```

ophyd/signal.py
tacaswell: Yeah, we need to sort out "did the user not pass _anything_ in which case we do our version of the default, otherwise just pass through to the underlying level" so we can thread `timeout=None` through to mean "no timeout".

prjemian: see c94f3ae 

This preserves the intent if user passed `timeout=None` meaning _do not timeout_.  It tries to use `self.write_timeout` but if _that_ was `None`, uses `self.timeout`.  Oy!

ophyd/signal.py
tacaswell: Sorry, I meant that `set_and_wait(name='xyz', ..)` won't work (but it does look nice in the log!)

Would `'%s: set_and_wait(...)'` be better?

ophyd/signal.py
```diff
@@ -1711,6 +1711,11 @@ def set(self, value, *, timeout=None, settle_time=None):
         --------
         Signal.set
         '''
+        if timeout is DEFAULT_WRITE_TIMEOUT:
+            timeout = self.write_timeout
+            if timeout is None:
```
tacaswell, line 1716: I do not think we want to fallback to the read time out here given that they are independently configurable https://github.com/bluesky/ophyd/blob/33bfc595ac2862aede96f023296f564832144551/ophyd/signal.py#L679-L699

and in the init we have a long comment differentiating the two https://github.com/bluesky/ophyd/blob/33bfc595ac2862aede96f023296f564832144551/ophyd/signal.py#L757-L765

tacaswell: We also don't do this fallback in `put` https://github.com/bluesky/ophyd/blob/33bfc595ac2862aede96f023296f564832144551/ophyd/signal.py#L1661-L1665

prjemian: My original version (setting `timeout` from `self.timeout`) was my best understanding of what to do,  Then I learned about `DEFAULT_WRITE_TIMEOUT`.  Now I'm puzzled what to do.  I thought the default timeout that was set in `EpicsSignalBase.set_default_timeout(timeout=10, connection_timeout=5)` would apply in this situation.

Can you suggest the code to use?

prjemian: Perhaps this is the one to use as a fallback? https://github.com/bluesky/ophyd/blob/33bfc595ac2862aede96f023296f564832144551/ophyd/signal.py#L866

prjemian: The reason I'm puzzled is that `self.write_timeout` is `None` in my test case and has not been affected by the `set_default_timeout()` method.  I assumed it would be.  Perhaps fallback (when `write_timeout` is unconfigured) to `__default_timeout ` fits best what I assumed would happen.

tacaswell:
I think this history here is:

 - we did the big "add global (read) timeout settings" change (via https://github.com/bluesky/ophyd/pull/788)
 - some time later when we tried to use this in anger on the floor we discovered we had only done the read (not write side) and added config for that too (via https://github.com/bluesky/ophyd/pull/804)
 - as part of #804 we forgot to thread the configuration of the default through the class level `set_default_timeout` classmethod

The naming on this is also confusing because in the read methods everything is called `timeout`, but in the write methods the local variables and function parameters are still called `timeout`, but the internal state is `write_timeout`.  In the init we have both `timeout` and `write_timeout` as kwargs.

prjemian:
Following #804, is it reasonable to support configuration of the default write timeout via `set_default_timeout()`?

tacaswell:
Yes

prjemian:
Good, I'll focus on that next, then we can circle back to this code and _make it right_, too.

prjemian:
Alternatively, the existing` set_default_timeout()` method and a proposed `set_default_auto_monitor()` method (1) could be combined into a single `set_defaults()` method and thus remove duplicated code and comments.

1. #927, from https://github.com/BCDA-APS/apstools/issues/450#issuecomment-719640338

ophyd/signal.py
```diff
@@ -858,15 +869,27 @@ def set_default_timeout(cls, *, timeout=2.0, connection_timeout=1.0):
...
+        cls.__default_write_timeout = write_timeout
 
-    # TODO Is there a good reason to prohibit setting these three timeout
-    # properties?
+        # TODO Is there a good reason to prohibit setting these three timeout
+        # properties?
+
+    @classmethod
+    def set_default_timeout(cls, *, **kwargs):
+        warnings.warn('Use set_defaults() instead.  Command ignored.')
```
danielballan, line 892: I would prefer to pass through to `cls.set_defaults(**kwargs)` here so that this change is backward compatible. We can warn about _future_ removal.

danielballan (commented):
This looks ready to me, minus one point about backward-compatibility marked inline.

ophyd/signal.py
```diff
@@ -858,15 +869,27 @@ def set_default_timeout(cls, *, timeout=2.0, connection_timeout=1.0):
...
+        # being too long and delaying the report of a failure.
+        cls.__default_write_timeout = write_timeout
 
-    # TODO Is there a good reason to prohibit setting these three timeout
-    # properties?
+        # TODO Is there a good reason to prohibit setting these three timeout
+        # properties?
+
+    @classmethod
+    def set_default_timeout(cls, *, **kwargs):
```
danielballan, line 891: This is a syntax error. You can either do `(cls, *args, **kwargs)` or `(cls, **kwargs)`. The latter means, "Only kwargs are accepted," which seems like your intent and the correct thing to do here, since `set_default_timeout` only ever accepted keyword arguments.

```suggestion
    def set_default_timeout(cls, **kwargs):
```

prjemian: Good catch.  One fine day, I will learn this pattern.

ophyd/signal.py
prjemian: Kinder, gentler, done

danielballan (approved):

PR #928, opened 2020-11-02
Adjust signal logging

danielballan:
Ophyd is very noisy at the INFO logging level, including at import time. In the last week, both Michael Sutherland (a user) and @tacaswell have bumped into this and been thrown by it. Quick comments:

* At beamlines we don't generally put INFO-level logging in the stderr, but it's apparently not uncommon for users or beamline staff to do this, so INFO level (or higher) should be used with discretion.
* The intent of the `control_layer_log` is to do the logging that we _wish_ pyepics did. (Eventually we'll add that pyepics, perhaps.) It's in a separate logger so that it can be separately filtered. In this PR, I move something from `log` to `control_layer_log` and turn the level of all `control_layer_log` calls to DEBUG.
* For some reason `SynSignal` logged aggressively --- including messages to note that `__init__` or `trigger` had been called. I don't think that are very helpful for debugging, so I have removed them. I left logging calls in the callbacks, which are more useful because those are not called directly by the user.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -138,7 +138,6 @@ class SynSignal(Signal):
             connected=True,
         )
 
-        self.log.info(self)
 
     def describe(self):
         res = super().describe()
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -147,12 +147,9 @@ class SynSignal(Signal):
         return res
 
     def trigger(self):
-        self.log.info('trigger %s', self)
-
         st = DeviceStatus(device=self)
         delay_time = self.exposure_time
         if delay_time:
-            self.log.info('%s delay_time is %d', self, delay_time)
 
             def sleep_and_finish():
                 self.log.info('sleep_and_finish %s', self)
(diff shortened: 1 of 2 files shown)


ophyd/sim.py
```diff
@@ -138,7 +138,6 @@ def __init__(self, func=None, *,
             connected=True,
         )
 
-        self.log.info(self)
```
ZLLentz, line 141: This one explains a lot of the startup logging spam

ZLLentz (approved):
I agree with all of these changes. We've seen a lot of logging noise on the INFO channel at SLAC as well (though we simply reconfigured our loggers appropriately in most cases). I like the use of the control layer log and appreciate keeping that info around in the debug stream, as I suspect it is still useful when things go wrong.

ophyd/sim.py
danielballan: Yep! I have a bet with the author of this line that if it ever comes up that this would have bailed us out of a hard-to-debug situation I owe a nice gift of some kind. We'll see!

PR #929, opened 2020-11-02
Do not log where EventDispatch queues are empty.

danielballan:
This removes a log message that I added and have since regretted.

The use of this message is to detect if event-proessing queues
are getting backed up, more or less akin to when caproto logs "high load"
messages. It is not necessary to log "low load"; omission is sufficient.

I added this in the first place because I wanted to ensure that the loop
was _running_ but that kind of debug activity is better handled by prints
as needed and should not have been merged into the main branch.


ZLLentz (approved):
Your logic for removing this log message seems sound to me

PR #935, opened 2020-11-09
Interpret num_capture=0 correctly.

### LINKED ISSUE #934: Is get_frames_per_point always giving 0?
When I run [this example script](https://gist.github.com/danielballan/3419df60330fb1543ac6306462a2e6c0#file-interactive_test-py-L31-L33) and monitor `13SIM1:HDF1:NumCapture_RBV`, it never changes from its initial value of `0`. This remains true whether the `13SIM1:HDF1:FileWriteMode_RBV` is set to `Capture` or `Stream` (as specified in `det.hdf1.stage_sigs` and verified by monitoring).

Thus, I have to override `get_frames_per_point()` to rely on `det.cam.num_images` (the _desired_ number of images) rather than `det.hdf1.num_capture`, as shown in the highlighted lines of the example script.

I notice that we override `get_frames_per_point` in the same fashion [in CSX's profile](https://github.com/NSLS-II-CSX/xf23id1_profiles/blob/4ef6a3e30b4d10c1ed33e91deb9fa36b99a60187/profile_collection/startup/csx1/devices/areadetector.py#L74-L75) and [in FXI's profile](https://github.com/NSLS-II-FXI/profile_collection/blob/5eea8486b35f45f150de501cc02d9a4690b0e626/startup/10-area-detector.py#L68-L69)

Should we make the same change upstream? Am I misunderstanding or misremembering how the original was supposed to work?

danielballan:
As explained in the comment, the `FileStoreHDF5` class has been recording its `num_capture` signal as the number of frames that will be captures for a given acquisition. The _correct_ behavior is to interpret `num_capture=0` as a special case because that means "Capture as many as you receive." In that case, our best guess (really an upper bound) is `num_images` from the detector's camera.

Closes #934

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -456,7 +456,15 @@ class FileStoreHDF5(FileStorePluginBase):
                                 ])
 
     def get_frames_per_point(self):
-        return self.num_capture.get()
+        num_capture = self.num_capture.get()
+        # If num_capture is 0, then the plugin will capture however many frames
+        # it is sent. We can get how frames it will be sent (unless
+        # interrupted) by consulting num_images on the detector's camera.
+        if num_capture == 0:
+            return self.parent.cam.num_images.get()
+        # Otherwise, a nonzero num_capture will cut off capturing at the
+        # specified number.
+        return num_capture
 
     def stage(self):
         super().stage()


danielballan:
Integration test:

```py
In [3]: det.cam.num_images.get()
Out[3]: 100

In [4]: det.hdf1.stage_sigs['num_capture']  # factory default sets num_capture to 0 when staged
Out[4]: 0

In [5]: RE(count([det]))
Out[5]: ('2089838d-6e31-4802-bf05-e1b9f599ca57',)

In [6]: catalog[-1].primary.read()['det_image'].shape
Out[6]: (1, 100, 1024, 1024)

In [7]: det.hdf1.stage_sigs['num_capture'] = 5  # This will cut off capturing at 5, even though num_images is higher.

In [8]: RE(count([det]))
Out[8]: ('3884e59a-53ce-48e5-9de5-ae2dac963639',)

In [9]: catalog[-1].primary.read()['det_image'].shape
Out[9]: (1, 5, 1024, 1024)
```

tacaswell (approved):

tacaswell:
Makes sense given the number of places we sub-class this and do the same logic currently on the floor...

prjemian (approved):

PR #936, opened 2020-11-09
Configure GH Actions for CI

danielballan:
We have run out of Travis-CI free credits. The plan is to move to GH Actions
for the short term and then reassess.


tacaswell (approved):
Passes when run on Dan's branch.

PR #943, opened 2020-11-18
ensure ScalerCH .S has default auto_monitor=False

prjemian:
Found & solved this problem with the [APS USAXS instrument](https://github.com/APS-USAXS/ipython-usaxs/issues/426).

### DIFF
--- a/ophyd/scaler.py
+++ b/ophyd/scaler.py
@@ -66,7 +66,7 @@ class ScalerChannel(Device):
     chname = FCpt(EpicsSignal, '{self.prefix}.NM{self._ch_num}',
                   kind=Kind.config)
     s = FCpt(EpicsSignalRO, '{self.prefix}.S{self._ch_num}',
-             kind=Kind.hinted)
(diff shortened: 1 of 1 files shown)


prjemian:
@danielballan Can this PR make it into the pending release?

jklynch (approved):

ophyd/scaler.py
```diff
@@ -66,7 +66,7 @@ class ScalerChannel(Device):
     chname = FCpt(EpicsSignal, '{self.prefix}.NM{self._ch_num}',
                   kind=Kind.config)
     s = FCpt(EpicsSignalRO, '{self.prefix}.S{self._ch_num}',
-             kind=Kind.hinted)
+             kind=Kind.hinted, auto_monitor=False)
```
mrakitin, line 69: Is it only needed for the `s` component, but not the others? I assume the default is `True` now, right?

MikeHart85 (approved):

prjemian:
The default is controlled by the end user.  Only the s component needed
this change in our use.

PR #947, opened 2020-11-24
MNT: motor record limits switch states are read-only

ZLLentz:
Very small thing I noticed today: `HLS` and `LLS` use `EpicsSignal` when they should be using `EpicsSignalRO`. This was found when `typhos` confusingly decided to make writable widgets for limit hit state.

Ref: https://epics.anl.gov/bcda/synApps/motor/motorRecord.html#Fields_limit


klauer (approved):

prjemian (approved):
Good catch!

mrakitin (approved):
Good catch :)

mrakitin:
The tests for 3.7 are failing on one test with the `TimeoutError`, which seems unrelated to the proposed changes.

PR #949, opened 2020-12-02
Deploy documentation.

danielballan:
Because this action deploys documentation to another repo (`bluesky.github.io`) it has to use a deploy key instead of `GITHUB_TOKEN` (which is scoped only the repo running the action). I have added a deploy key with write access to `bluesky.github.io` and made the private key securely available to all repositories in this organization as `ACTIONS_DOCUMENTATION_DEPLOY_KEY`.

The action seems to do everything we need: https://github.com/marketplace/actions/github-pages-action#%EF%B8%8F-create-ssh-deploy-key

I have taken security considerations into account: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions


danielballan:
**DO NOT MERGE.** Let's test this action on a sandbox org before we deploy.

danielballan:
I pushed a copy of the repo `bluesky/bluesky.github.io` to `bluesky-sandbox/bluesky-sandbox.github.io` and a copy of `bluesky/ophyd` to `bluesky-sandbox/ophyd`. I merged this branch into `bluesky-sandbox/ophyd` with only this change:

```diff
-external_repository: bluesky/bluesky.github.io
+external_repository: bluesky-sandbox/bluesky-sandbox.github.io
```

On merging, the workflow completed successfully: https://github.com/bluesky-sandbox/ophyd/runs/1493650066?check_suite_focus=true

and the artifacts were uploaded as expected: https://github.com/bluesky-sandbox/bluesky-sandbox.github.io/tree/master/ophyd

I believe this is ready to merge.

dmgav:
Looks good to me. My only concern is that the script will also be executed when changes are pushed to 'master' branch in a fork (I tend to enable CI in forks). I am not sure how the script will behave but it is likely to fail, which will make CI fail. I typically try to disable doc deployment scripts in forks: https://github.com/bluesky/bluesky-queueserver/blob/7ec1e38a7f79e1f6bb37459110dd31839acfddb1/.github/workflows/docs_publish.yml#L11
This involves hard coding of the repository name, but repository name is rarely changed. The script for building docs should still be executed in the fork.

danielballan:
Thanks, @dmgav. Good catch. Fix pushed, as you suggested.

dmgav (approved):

PR #950, opened 2020-12-03
Fix copy/paste typo in workflow condition.

danielballan:
Follow-up to #949


dmgav (approved):
I think `repository_owner` is more suitable than simply `repository` as it will stay the same for all Bluesky projects.

PR #951, opened 2020-12-08
Update MockFlyer in sim.py to return self.name in describe_collect

lsammut:
Return `self.name`(instead of 'stream_name')  in MockFlyer `describe_collect` method. This will show the flyer name and prevent ambiguation when running multiple flyers.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -765,7 +765,7 @@ class MockFlyer:
         dd = dict()
         dd.update(self._mot.describe())
         dd.update(self._detector.describe())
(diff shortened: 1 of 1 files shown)


mrakitin (approved):
Thanks for the correction, @lsammut! That looks correct to me.

danielballan:
Thanks!

The unit tests segfaulted. I'm sure that's unrelated to this PR, but in case something else has gone wrong upstream I've rerun them to be safe. If they pass, merge at will whoever sees this first. :-)

PR #954, opened 2020-12-11
GitHub Actions follow-up

danielballan:
- [x] Remove Travis-CI and encoded tokens for doctr
- [x] Factor installation of system and Python packages into a standalone bash script that can be reused between workflows.
- [x] Make `__version__` set correctly so that the version shown in the documentation is correct. (The issue was that we have to fetch the tags and the most recently tagged commit so that `git describe` works.)
- [x] Extract the name of the repo (`'ophyd'` in this case) from the environment to determine the `destination_dir` for published docs. By avoiding hard-coding the repo name, we make `publish-docs.yml` portable without changes. This is of particular interest because we learned how easy it is to copy `publish-docs.yml` to another repo, not notice that the `destination_dir` needs to be updated, and then overwrite a different project's documentation.
- [x] Run tests on a daily schedule (as well as on `push` and `pull_request`). This is like the Travis-CI CRON we used to run.


.github/workflows/docs.yml
```diff
@@ -16,16 +16,13 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Install
+      shell: bash -l {0}
+      run: source continuous_integration/scripts/install.sh
+    - name: Install documentation-building requirements
```
prjemian, line 22: Can you add a blank line before each new group.  Readability.  It's a preference of mine.

prjemian: Also, just learned the hard way to add a `shell` specifier before `run`, to control which shell for the commands to follow.  With no spec, it's `sh -e {0}`.  If conda is used (as hklpy needs it), then `shell: bash -l {0}` should be specified.

My advice on shells comes from the README of this action: https://github.com/conda-incubator/setup-miniconda/tree/8246c3d5d22bc5ca3416ddf4f74b77c2fb1d33d0#IMPORTANT

.github/workflows/flake8.yml
```diff
@@ -8,10 +8,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - name: Install Dependencies
+      - name: Install flake8
         run: |
```
prjemian, line 12: shell: . . .

.github/workflows/publish-docs.yml
```diff
@@ -15,24 +15,26 @@ jobs:
...
-        pip install .
-        pip install -r requirements-docs.txt
-        pip list
+        python -m pip install -r requirements-docs.txt
+        python -m pip list
     - name: Build Docs
       run: |
-        git fetch --tags
         make -C docs/ html
+    - name: Set env.REPOSITORY_NAME  # just the repo, as opposed to org/repo
```
prjemian, line 34: Why not do this very early on?  It becomes available earlier.  Some other repos may want this for an action.

.github/workflows/testing.yml
```diff
@@ -1,6 +1,10 @@
 name: Unit Tests
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+      - cron: '00 4 * * *'  # daily at 4AM
```
prjemian, line 7: 👍

.github/workflows/publish-docs.yml
```diff
@@ -15,24 +15,26 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1000  # should be enough to reach the most recent tag
```
prjemian, line 19: 🥇

continuous_integration/scripts/install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -vxeuo pipefail
+
+sudo apt-get install graphviz procserv libhdf5-dev
+# These packages are installed in the base environment but may be older
+# versions. Explicitly upgrade them because they often create
+# installation problems if out of date.
+python -m pip install --upgrade pip setuptools wheel numpy
+# Verioneer uses the most recent git tag to generate __version__, which appears
```
prjemian, line 9: spelling: Versioneer

prjemian (changes requested):
Changes are minor or style. YMMV

danielballan:
Thanks for the close read, @prjemian. All suggestions addressed.

prjemian (approved):

dmgav (approved):
I see no issues.

dmgav:
I noticed that by default GH Actions are will cancel all jobs if one of the jobs fails ('fail-fast' mode is ON), which I find very annoying. If this behavior is not wanted, 'fail-fast' mode can be disabled as this: https://github.com/bluesky/bluesky-queueserver/blob/0a5f34a7246cca29babb0c9064d237595f27eba2/.github/workflows/testing.yml#L12

danielballan:
That makes sense to me. Updated. Thanks, @dmgav.

mrakitin:
For future reference: https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows.html. That's where we originally found it for the queueserver.

.github/workflows/flake8.yml
```diff
@@ -6,12 +6,19 @@ jobs:
...
-          pip install flake8
+          python -m pip install flake8
+          python -m pip list
+
       - name: Run flake8
+        shell: bash -l {0}
         run: |
-          flake8
\ No newline at end of file
+          flake8
```
mrakitin, line 24: `fail-fast: false` for this file too?

mrakitin (approved):
Looks good!

I think `fail-fast: false` can be used in every workflow, to override the default behavior.

PR #955, opened 2020-12-15
Update CI shield in README.

danielballan:


danielballan:
Looks like:
image

prjemian:
`ophyd - EPICS Client Abstractions`

Could this change to `ophyd - hardware client abstractions` (so not to exclude the non-EPICS world that must exist somewhere)

danielballan:
Ah, I meant to do that while I was here. Thanks for the bump. Done.

danielballan:
I'll that thumbs up from @prjemian as an approval and self-merge. Pushing to get all our major repos in good standing today....

PR #963, opened 2021-02-04
Doc multi collect flyer

tacaswell:
Migrated from a slack conversation.

### DIFF
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -190,7 +190,7 @@ The expected sequencing of the commands is
 
 - ``kickoff`` and wait
 - ``complete`` and wait
-- a to ``collect``
+- ``collect``
 
 Optionally, devices my implement "partial collection" so that they can
 be incrementally collected during acquisition.  While this may not be
@@ -201,7 +201,7 @@ point of view of the data consumers.
 - ``kickoff`` and wait
 - 0 or more calls to ``collect``
 - ``complete`` and wait
-- a to ``collect``
+- ``collect``
 
 
 ::
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -186,6 +186,76 @@ The flyable interface provides four methods
    ~flyers.FlyerInterface.describe_collect
    ~flyers.FlyerInterface.collect
 
+The expected sequencing of the commands is
+
+- ``kickoff`` and wait
+- ``complete`` and wait
+- a to ``collect``
+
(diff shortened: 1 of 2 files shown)


docs/source/architecture.rst
```diff
@@ -186,6 +186,76 @@ The flyable interface provides four methods
    ~flyers.FlyerInterface.describe_collect
    ~flyers.FlyerInterface.collect
 
+The expected sequencing of the commands is
+
+- ``kickoff`` and wait
+- ``complete`` and wait
+- a to ``collect``
```
mrakitin, line 193:
```suggestion
- ``collect``
```

docs/source/architecture.rst
```diff
@@ -186,6 +186,76 @@ The flyable interface provides four methods
...
+Optionally, devices my implement "partial collection" so that they can
+be incrementally collected during acquisition.  While this may not be
+technically possible in every situation, it can be used to get partial
+results and allow the fly scan to look more like a step scan from the
+point of view of the data consumers.
+
+- ``kickoff`` and wait
+- 0 or more calls to ``collect``
+- ``complete`` and wait
+- a to ``collect``
```
mrakitin, line 204:
```suggestion
- ``collect``
```

ophyd/sim.py
```diff
@@ -784,36 +786,37 @@ def flyer_worker():
             st.set_finished()
 
         threading.Thread(target=flyer_worker, daemon=True).start()
-
-        return st
+        kickoff_st = DeviceStatus(device=self)
+        kickoff_st.set_finished()
```
mrakitin, line 790: Does it mean it will return the `done` status immediately? Is it what we need?

mrakitin (approved):
Looks good to me. I have suggested a couple of copy-edits.

ophyd/sim.py
tacaswell: Yes.  I guess we could be very fancy and do this in at the top of the thread, but there is nothing we are waiting on here for the simulated device to report "I have started flying!".

The goal is to avoid the following:

1. kick off flyer
2. start moving motor
3. move through data range
4. flyer finishing arming itself and starts collecting data

so kickoff is an async thing so we can do

1. kick off flyer1
2. kickoff flyer2
3. wait for both to report they are read to take data
4. start moving motor
5. ????
6. profit!

mrakitin: That makes sense to me! Thanks for the details.

PR #968, opened 2021-02-26
Update release notes for v1.6.1

MikeHart85:


docs/source/release_notes.rst
```diff
@@ -2,6 +2,16 @@
...
 
+1.6.1 (2021-02-26)
+==================
+
+Fixes
+-----
+
+* Added a missing name argument to AreaDetector documentation
+* Correctly propagate changes to the default ``connection_timeout`` setting
+  via ``set_defaults``, instead of always defaulting to 1.0 seconds.
```
mrakitin, line 13:
```suggestion
  via ``set_defaults``, instead of always defaulting to 1.0 second.
```

mrakitin (approved):
Looks good. Minor suggestion.

docs/source/release_notes.rst
MikeHart85: It's a bit counter-intuitive, but I think the way I have it is grammatically correct. 

Only an integer 1 can be singular, while all floating point numbers are plural. So it would have to either be "1 second" or "1.0 seconds"... though I have no strong preference to which of the two to use. I picked 1.0 because that is exactly as it is in the code.

mrakitin: That's a good point. I think the release notes are for humans, not computers, maybe having "1 second" makes sense here. I'll update my suggestion.

mrakitin (approved):

mrakitin:
@MikeHart85, do you want to incorporate #963 into the release notes too?

MikeHart85:
Yes, adding that as well. 

I think we should also merge in #969 if possible. The change looks good to me, given the discussion in the issue, but another pair of eyes would be great. Could someone else also have a look and hit the merge button?

danielballan:
Good point. Done.

MikeHart85:
I think this is ready to merge, unless there are any other PRs we want to include?

If everything looks ready to go, I'll proceed with the release as soon as someone merges these notes.

danielballan (approved):

PR #972, opened 2021-03-01
FIX: protect against a failed reading in set_and_wait

tacaswell:
If `signal.get` returns None (which pyepics will sometimes do) and tolerances
are in use, passing `None` through to numpy will cause numpy to choke.

We were already implicitly re-trying in the non-tolerance case.

Encountered this in the wild at PDF over the weekend.


mrakitin (approved):
Looks good to me. I hope this change will make ophyd more reliable in production.

PR #977, opened 2021-03-17
FIX: re-enable setting a Signal to None

tacaswell:
The change in #972 (to make sure we did not accidentally feed None to numpy)
also made it impossible to set a (soft)Signal to None.

While there may or may not be a good case for doing this, it is implicitly
tested in bluesky so restore the functionality.


dmgav:
Tested with Bluesky CI. Seem to fix the issue.

dmgav (approved):

mrakitin (approved):

PR #985, opened 2021-05-04
Fix enum py310

tacaswell:
We may want to take this as an opportunity to re-think the Kind enum? 

Python 3.10 makes a number of changes to Enum and its sub-classes.  Under the new default behavior of IntFlag the fact that we have `0b1b1` as a flag, but not `0b100` is considered a class definition time error (preventing import of ophyd).

This opts back into allowing our (unorthodox) definition to import again and hides the new features behind a version gate.

python/cpython#24215

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -32,7 +32,17 @@ def select_version(cls, version):
     return all_versions[matched_version]
 
 
-class Kind(IntFlag):
+if sys.version_info < (3, 10, 0, 'alpha', 5):
+    IFBase = IntFlag
+else:
+    from enum import KEEP
+
+
+    class IFBase(IntFlag, boundary=KEEP):
+        ...
+
+
+class Kind(IFBase):
     """
     This is used in the .kind attribute of all OphydObj (Signals, Devices).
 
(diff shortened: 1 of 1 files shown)


klauer:
Some rethinking of the enum will indeed be necessary (see previous PR/issue attempting to change the values), but _agreement_ on what should be done will be the tough part.

tacaswell:
> ...but agreement on what should be done will be the tough part.

Is that not always the case ?

prjemian (commented):
Anything that makes `Kind` easier to understand will be welcome.  I am not committed to anything other than the single name use cases (`omitted`, `config`, `normal`, `hinted`, and not the multiple cases such as this contrived example: `omitted | config`).

tacaswell:
In practice we seem to have a simple ordering of `OMITTED < CONFIG < NORMAL < HINTED`, maybe we can switch from intflag to a normal enum?

tacaswell:
It looks like upstream is going to back off of this change (by default) https://bugs.python.org/issue44242#msg394593 so this should be put on hold.

tacaswell:
This did get reverted upstream, closing.

Issue #986, opened 2021-05-04
DOC publishing fails due to Sphinx error


prjemian:
The publishing build failed here:

```python
make: Entering directory '/home/runner/work/ophyd/ophyd/docs'
sphinx-build -b html -d build/doctrees  -W source build/html
Running Sphinx v1.6.7
making output directory...

Exception occurred:
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/matplotlib/sphinxext/plot_directive.py", line 280, in setup
    app.add_css_file('plot_directive.css')
AttributeError: 'Sphinx' object has no attribute 'add_css_file'
The full traceback has been saved in /tmp/sphinx-err-hze40yz5.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:61: html] Error 1
make: Leaving directory '/home/runner/work/ophyd/ophyd/docs'
Error: Process completed with exit code 2.
```

The Sphinx version is exceptionally old (most recent release is 4.0.0b or some such).  The change this PR contributed cannot be responsible for that problem.  This is a new bug then.

_Originally posted by @prjemian in https://github.com/bluesky/ophyd/issues/984#issuecomment-831647029_

prjemian:
https://github.com/bluesky/ophyd/runs/2493817187?check_suite_focus=true

mrakitin:
That's what I found regarding the `AttributeError: 'Sphinx' object has no attribute 'add_css_file'` error: https://github.com/sphinx-doc/sphinx/issues/7747#issuecomment-636023612.

The sphinx package in ophyd's doc requirements has that hard pin:
https://github.com/bluesky/ophyd/blob/82053af306815a4980aa4689a31d717f3234bd43/requirements-docs.txt#L6

Matplotlib extension seems to require the latest API, so we should upgrade the sphinx version.

prjemian:
I'm using Sphinx 3.? or higher on my side. If the server builds the docs, I'm OK with closing this issue.

mrakitin:
Closed via #991.

PR #991, opened 2021-05-17
Fix #990: Import 'SignalRO' with 'ophyd'

### LINKED ISSUE #990: SignalRO not imported by ophyd/__init__.py
To use `SignalRO`, it must be imported from its submodule, rather than the main `ophyd` module.

```python
from ophyd.signal import SignalRO  # This works
from ophyd import SignalRO  # This does NOT work
```

Could `SignalRO` be imported in '[ophyd/__init__py](https://github.com/bluesky/ophyd/blob/master/ophyd/__init__.py#L78)' just like (`Signal`, `EpicsSignal`, `EpicsSignalRO`, `DerivedSignal`) and many other ophyd classes?  Or is `SignalRO` obscured, and intended only for edge cases?

'[ophyd/__init__py](https://github.com/bluesky/ophyd/blob/master/ophyd/__init__.py#L78)':
```python
[line 78] from .signal import (Signal, EpicsSignal, EpicsSignalRO, DerivedSignal)  # Replace this line...
[line 78] from .signal import (Signal, SignalRO, EpicsSignal, EpicsSignalRO, DerivedSignal)  # ...with this line
```

pshafer-als:
`SignalRO` is now imported by 'ophyd.__init__.py' 
so that it is available when importing `ophyd`

```python
from ophyd import SignalRO  # This now works
from ophyd import *  # ...so does this
```

### DIFF
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -592,3 +592,21 @@ def test_posix_paths(paths, path_signal):
 def test_path_semantics_exception():
     with pytest.raises(ValueError):
         EpicsPathSignal('TEST', path_semantics='not_a_thing')
+
+        
+@pytest.fixture(scope="module")
+def imported_ro_signal_class_from_pkg():
+    from ophyd import SignalRO
+    return SignalRO
+
+
+@pytest.fixture(scope="module")
+def imported_ro_signal_class_from_module():
+    from ophyd.signal import SignalRO
+    return SignalRO
+
+
+def test_import_ro_signal_class(
+        imported_ro_signal_class_from_pkg, 
+        imported_ro_signal_class_from_module):
+    assert imported_ro_signal_class_from_pkg == imported_ro_signal_class_from_module
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -593,9 +593,9 @@ def test_path_semantics_exception():
     with pytest.raises(ValueError):
         EpicsPathSignal('TEST', path_semantics='not_a_thing')
 
-        
+
 def test_import_ro_signal_class():
     from ophyd import SignalRO as SignalRoFromPkg
     from ophyd.signal import SignalRO as SignalRoFromModule
-    
+
     assert SignalRoFromPkg is SignalRoFromModule
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -594,19 +594,8 @@ def test_path_semantics_exception():
         EpicsPathSignal('TEST', path_semantics='not_a_thing')
 
         
-@pytest.fixture(scope="module")
-def imported_ro_signal_class_from_pkg():
-    from ophyd import SignalRO
-    return SignalRO
-
-
-@pytest.fixture(scope="module")
-def imported_ro_signal_class_from_module():
-    from ophyd.signal import SignalRO
-    return SignalRO
-
-
-def test_import_ro_signal_class(
-        imported_ro_signal_class_from_pkg, 
-        imported_ro_signal_class_from_module):
-    assert imported_ro_signal_class_from_pkg == imported_ro_signal_class_from_module
+def test_import_ro_signal_class():
+    from ophyd import SignalRO as SignalRoFromPkg
+    from ophyd.signal import SignalRO as SignalRoFromModule
+    
+    assert SignalRoFromPkg is SignalRoFromModule
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -3,5 +3,5 @@ ipython
 ipywidgets
 matplotlib
 numpydoc
-sphinx==1.6.7
+sphinx>=1.8.1,!=2.0.0
 sphinx_rtd_theme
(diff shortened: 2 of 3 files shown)


prjemian (changes requested):
This should be paired with a unit test.  Perhaps in `ophyd/tests/test_signal.py`

pshafer-als:
Sounds good.  Will give it a try.

ophyd/tests/test_signal.py
```diff
@@ -592,3 +592,21 @@ def test_posix_paths(paths, path_signal):
...
+@pytest.fixture(scope="module")
+def imported_ro_signal_class_from_module():
+    from ophyd.signal import SignalRO
+    return SignalRO
+
+
+def test_import_ro_signal_class(
+        imported_ro_signal_class_from_pkg, 
+        imported_ro_signal_class_from_module):
+    assert imported_ro_signal_class_from_pkg == imported_ro_signal_class_from_module
```
mrakitin, line 612: Maybe it can be done simpler without extra fixtures, such as:
```py
In [5]: from ophyd import SignalRO as SignalROFromPkg

In [6]: from ophyd.signal import SignalRO as SignalROFromModule

In [7]: SignalRO is SignalROFromModule
Out[7]: True
```

In my example I used the identity check (with `is`). I think your test can use that too.

pshafer-als: Sure, that works for me. Thanks for the guidance.  I'll make these changes.

For the glob import test, the following statement needs to be done at module level: `from ophyd import *`.  Is it ok for me to add a new file to the 'tests' folder for this purpose?  Or is there a better way to do that?

I don't want to pollute the namespace of the test_module.

pshafer-als: Actually, I'll try this with `from ophyd.__init__ import SignalRO`, so won't need a separate file.

`ophyd.__all__` is not defined; otherwise that should work also.

mrakitin: Thanks! Regarding the star import, it's generally discouraged, and, as you rightly mentioned, it can pollute the namespace and can lead to confusion. I think the explicit imports would be sufficient for tests.

mrakitin (approved):

pshafer-als:
As best as I can tell, the failing tests in the Documentation build appears to be an error with the build environment.  Please let me know if it's something on my end that needs fixing.

Here's a snippet from the [full output](https://github.com/bluesky/ophyd/pull/991/checks?check_run_id=2604828514)...

```
Exception occurred:
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/matplotlib/sphinxext/plot_directive.py", line 280, in setup
    app.add_css_file('plot_directive.css')
AttributeError: 'Sphinx' object has no attribute 'add_css_file'
```

prjemian:
Might be a sphinx version issue

pshafer-als:
> Might be a sphinx version issue
> […](#)

So I ended up updating `sphinx` to `>=1.8.1` in 'requirements-docs.txt', to allow the `matplotlib` docs to build.  I'm surprised that building the `matplotlib` docs is a part of the build tests for `ophyd`, but hey, I'm out of my league here.  :)

If this is overkill, then we could just use the previous commit.  Other suggestions welcomed.

requirements-docs.txt
```diff
@@ -3,5 +3,5 @@ ipython
 ipywidgets
 matplotlib
 numpydoc
-sphinx==1.6.7
+sphinx>=1.8.1,!=2.0.0
```
mrakitin, line 6: Why do we need to exclude v2.0.0?

mrakitin (commented):
Looks like it's addressing https://github.com/bluesky/ophyd/issues/986 as well!
@prjemian, have you done any sphinx updates on your side, or can we close that issue by this PR?

mrakitin:
FYI, it picked the latest sphinx [v4.0.1](https://pypi.org/project/Sphinx/4.0.1/), released on May 10, 2021:
image

requirements-docs.txt
mrakitin: It picked a much newer version 4.0.1 anyway, so I don't think it's a blocking question.

mrakitin (approved):
Looks good to me, thanks for the contribution, @pshafer-als!

requirements-docs.txt
pshafer-als: > Why do we need to exclude v2.0.0?

That comes from [`matplotlib` docs requirements](https://github.com/matplotlib/matplotlib/blob/master/requirements/doc/doc-requirements.txt).

mrakitin: Cool! Posting the permalink here for the future us, just in case: https://github.com/matplotlib/matplotlib/blob/4723f8a42c8190f5cb287aca8fdf80ec6c37b102/requirements/doc/doc-requirements.txt#L10.

prjemian: I'm using Sphinx 3.? or higher on my side.  If the server builds the docs, I'm OK with closing #986.

prjemian (approved):
LGTM!

mrakitin:
@tacaswell, do you have any objections to the sphinx pin?

mrakitin:
I spoke to Tom, and he gave his OK on merging it if we don't have concerns about it. The docs are building, and that's a good fix.

Closed #986.

pshafer-als:
Woohoo, my first PR!  :)

prjemian:
Great job!

PR #992, opened 2021-06-01
Adding classes to ophyd to support the areaDetector Kafka plugin

jwlodek:
This plugin allows areaDetector to post image data to a kafka broker server that can then be consumed by a consumer. The Signals for the class were created with the AD Boilerplating script. I have already built and started testing the plugin, and can already get data via a consumer - so next step would be to try to get some tests setup to try and collect data throught bluesky in this way.

For reference below is the CSS screen for the plugin, and my fork (the one I have working) is [here](https://github.com/jwlodek/ADPluginKafka).

image


ophyd/areadetector/plugins.py
```diff
@@ -836,6 +837,28 @@ class JPEGPlugin(FilePlugin, version=(1, 9, 1), version_type='ADCore'):
...
+    reconnect_flush_time = Cpt(SignalWithRBV, 'ReconnectFlushTime')
+    kafka_buffer_size = Cpt(SignalWithRBV, 'KafkaBufferSize')
+    kafka_max_message_size = Cpt(SignalWithRBV, 'KafkaMaxMessageSize')
+    unsent_packets = Cpt(EpicsSignalRO, 'UnsentPackets_RBV')
+    connection_status = Cpt(EpicsSignalRO, 'ConnectionStatus_RBV')
+    connection_message = Cpt(EpicsSignalRO, 'ConnectionMessage_RBV')
+    kafka_topic = Cpt(SignalWithRBV, 'KafkaTopic')
+    kafka_broker_address = Cpt(SignalWithRBV, 'KafkaBrokerAddress')
+    kafka_stats_interval_time = Cpt(SignalWithRBV, 'KafkaStatsIntervalTime')
+    kafka_max_queue_size = Cpt(SignalWithRBV, 'KafkaMaxQueueSize')
```
mrakitin, line 859: I'd personally sort the entries alphabetically for an easier search of a parameter, but it's up to you. If it's based on the order from the AD source, we can leave it as is.

mrakitin (approved):
Awesome feature! I'd like to try it at a beamline.

mrakitin:
The tests are failing pretty badly with the segmentation fault errors. @jwlodek, do you mind printing the values of `ulimit -a` before this line: https://github.com/jwlodek/ophyd/blob/0e4a86c0538719b340ba4f7ce9e188908728f5bf/.github/workflows/testing.yml#L49? It may help to set the stack size to `unlimited` (with `ulimit -s unlimited`).

ophyd/areadetector/plugins.py
jwlodek: We can make them alphabetical if that's preferable - as you said, the current order is just the order they were defined in the Plugin Driver code.

mrakitin: I am in favor of the sorted items, or, at least, the same order in which the parameters appear in CSS screens.

mrakitin (approved):

danielballan (approved):

PR #901, opened 2020-09-01
Add way to escape 'Another set in progress'.

danielballan:
See #757. This is one possible approach.

TO DO
- [ ] Needs broad input about whether this is the right way to address #757
- [ ] Needs tests
- [x] As written this leaves a polling thread dangling. To fix that properly, we'd need `set_and_wait` to accept a poison pill `threading.Event` or some other means of signaling that it should stop waiting.

PR #994, opened 2021-06-04
Define last_status_object property in Signal class

### LINKED ISSUE #993: ophyd.Signal._status does not exist until .set() is called
In the `Signal()` class, the `_status` attribute is not set until the `set()` method is called.  https://github.com/bluesky/ophyd/blob/474037df5e73f7fce5c3da57f8e0822bcc4e25db/ophyd/signal.py#L322-L323

Can the `_status` be defined as a class attribute? Then, tests such as https://github.com/APS-USAXS/ipython-usaxs/issues/501 do not also need to test if this attribute has yet been defined.

prjemian:
FIX #993 by adding a `last_status_object ` property (to provide access to the internal `_status` attribute, if it exists).

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -85,6 +85,7 @@ class Signal(OphydObject):
     _default_sub = SUB_VALUE
     _metadata_keys = None
     _core_metadata_keys = ('connected', 'read_access', 'write_access', 'timestamp')
+    _status = None
 
     def __init__(self, *, name, value=0., timestamp=None, parent=None,
                  labels=None, kind=Kind.hinted, tolerance=None,
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -85,7 +85,6 @@ class Signal(OphydObject):
     _default_sub = SUB_VALUE
     _metadata_keys = None
     _core_metadata_keys = ('connected', 'read_access', 'write_access', 'timestamp')
-    _status = None
 
     def __init__(self, *, name, value=0., timestamp=None, parent=None,
                  labels=None, kind=Kind.hinted, tolerance=None,
@@ -326,6 +325,11 @@ class Signal(OphydObject):
         self._set_thread.daemon = True
         self._set_thread.start()
         return self._status
+
+    @property
+    def last_status_object(self):
+        '''The most recent status object, if available, or ``None``.'''
+        return getattr(self, "_status", None)
 
     @property
     def value(self):
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -601,13 +601,13 @@ def test_import_ro_signal_class():
     assert SignalRoFromPkg is SignalRoFromModule
 
 
-def test_last_status_object():
+def test_signal_last_status_object():
     signal = Signal(name="signal", value=1)
     assert not hasattr(signal, "_status")
     assert signal.last_status_object is None
 
     signal.put(0)
-    assert hasattr(signal, "_status")
+    assert not hasattr(signal, "_status")
 
     signal.set(2)
     assert hasattr(signal, "_status")
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -599,3 +599,14 @@ def test_import_ro_signal_class():
     from ophyd.signal import SignalRO as SignalRoFromModule
 
     assert SignalRoFromPkg is SignalRoFromModule
+
+
+def test_last_status_object():
+    signal = Signal(name="signal", value=1)
+    assert not hasattr(signal, "_status")
+    assert signal.last_status_object is None
+
+    signal.put(0)
+    assert hasattr(signal, "_status")
+    assert signal.last_status_object is not None
+    assert signal.last_status_object == signal._status
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -608,5 +608,8 @@ def test_last_status_object():
 
     signal.put(0)
     assert hasattr(signal, "_status")
+
+    signal.set(2)
+    assert hasattr(signal, "_status")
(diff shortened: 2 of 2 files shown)


prjemian:
APS USAXS is [interested](https://github.com/APS-USAXS/ipython-usaxs/issues/501) in this PR.

klauer:
It's a "semi-private" attribute that isn't _really_ intended to be accessed external to the signal. 

Is there a way your external code could be restructured to instead use the return value of `set()`?

prjemian:
`Another call to set() in progress` is the underlying issue, so no, 
unless there is a way to resolve that problem by other means.  This has 
been the recommended path so far.

Today, with the USAXS instrument, we discovered that a default of 
`Status()` is better than `None`, and then call `.set_finished()` so it 
appears to have been successful.

The alternative is that we define this immediately after defining any 
Signal (or subclass) objects where we expect to find `_status`.

ZLLentz:
Some related thoughts (no real opinion on the PR itself):

I think I'd find a public api to get the "most recent status" returned by a settable device useful.

Does calling `set_finished` on `._status` "unstick" a device that is stuck with "Another call to set() in progress", including cleaning up the thread? That's a useful trick to know if so. Maybe there should also be a public API for cancelling a move? Personally I've switched to always defining short timeouts instead, but I understand this isn't possible for all use cases.

prjemian:
In this case, we are not certain of the underlying cause as it only seems to present the `Another call to set() in progress` exception after a long sequence of repeated scans; otherwise identical scans at that.  (I'm suspicious of some unreported resource consumption.)  I do not believe that calling the `set_finished()` method of a `Status()` object will clear the timeout after the timeout has elapsed.  With research, I can dig up the issues where we explored the timeout situation here in ophyd related to `Another call to set() in progress`.

While it is a "private" attribute, here is an example of beam time lost since it was defined only because `set()` was called.  What would be the intent of making this status object an attribute of the class: https://github.com/bluesky/ophyd/blob/89baea23125939b55cf20046fd5337fcb3b73fbd/ophyd/signal.py#L324 if such external access was not intended?  The caller receives this object: https://github.com/bluesky/ophyd/blob/89baea23125939b55cf20046fd5337fcb3b73fbd/ophyd/signal.py#L328

prjemian:
@ZLLentz : In this case, the use is to watch for completion of a fly scan operated by an EPICS database.  `self.flying` is the `ophyd.Signal`, [set to `True`](https://github.com/APS-USAXS/ipython-usaxs/blob/master/profile_bluesky/startup/instrument/devices/usaxs_fly_scan.py#L215) when pushing the ["Go" button (PV)](https://github.com/APS-USAXS/ipython-usaxs/blob/master/profile_bluesky/startup/instrument/devices/usaxs_fly_scan.py#L200-L205) and then [set to to `False`](https://github.com/APS-USAXS/ipython-usaxs/blob/master/profile_bluesky/startup/instrument/devices/usaxs_fly_scan.py#L220) when [EPICS says the fly scan is complete](https://github.com/APS-USAXS/ipython-usaxs/blob/master/profile_bluesky/startup/instrument/devices/usaxs_fly_scan.py#L219).

The fly scan takes a user-selectable time (60-240 s) and is allowed an additional 100s if a new trajectory needs to be loaded from EPICS to the motor controller (which is the low-level controller conducting the fly scan).

klauer:
There is a significant difference between making something part of the public API in terms of attribute names, and returning the object itself in a method. Lost beam time is also an entirely different story.

`._status` should _not_ be relied upon externally. Adding a `get_last_status_object()` or a `.last_status_object` property to the supported API would be a better start to such a discussion.

In your current case, the burden is on the caller to ensure that the internals are as expected. That is to say, your code should just read `getattr(obj, "_status", None)`.

> Maybe there should also be a public API for cancelling a move? Personally I've switched to always defining short timeouts instead, but I understand this isn't possible for all use cases.

I think the original intention was `.stop()` could be used even in non-positioners. Perhaps we should be implementing it here.

prjemian:
I'm good with exposing this via a public property.  I'll modify the PR.

prjemian:
Lots of TimeoutError exceptions raised for area detector PVs for py36 testing: not related to this PR.

@klauer : This is ready to review again.

prjemian:
Note, only the py36 (push) tests failed, not the py36(pull_request) tests.  Re-running all CI jobs, just in case.

prjemian:
This time on py38 (push) only.  Inconsistency is on the CI server?  Run again.

prjemian:
Continued CI failures due to TimeoutError and EPICS area detector features -- not related to this PR.

danielballan:
I’m late to this discussion, but in October we (@prjemian, @ZLLentz, and I) developed a solution to “Another set in progress” that has been approved. It just needs a rebase and a test.

https://github.com/bluesky/ophyd/pull/901#issuecomment-759019053

Reaching into the status object and calling set_finished is a clever solution, but I prefer where we landed in #901: a new method clear_set() that explicitly says “we’re canceling this because the part of the code that was _supposed_ to mark it as either finished or errored never did so.” I like that it can issue a warning with specifics, so that if it happens often and perhaps one or two layers down in the code, there is a way for controls staff to become aware of a recurring issue.

If we could nudge #901  across the finish line, would it solve your problem, @prjemian?

tacaswell:
I agree with @danielballan , #901 is a better solution to this problem and points people at the actual problem (that a `set` method is broken) while still providing a "get-on-with-life" escape hatch.

prjemian:
@danielballan Thanks, big time, for picking that up.  I'll pivot to #901.  If that resolves the OP (https://github.com/APS-USAXS/ipython-usaxs/issues/501, where a short-term work-around was implemented), then I can drop this PR.

prjemian:
dropping this branch & PR

PR #998, opened 2021-06-30
cast config_attrs as list(config_attrs)

### LINKED ISSUE #997: TypeError using configuration_attrs
In https://github.com/BCDA-APS/apstools/issues/518, a `TypeError` exception was raised when calling the `.summary()` method of an instance of a custom `ophyd.Device` subclass (ophyd 1.6.1).  In that issue, the problem was traced back to the type of `Device.configuration_attrs` as used in: https://github.com/bluesky/ophyd/blob/d0bb265e7e032e73cb3a778194fe167e58c932df/ophyd/device.py#L1082

The exception reports:

```python
TypeError: can only concatenate list (not "OphydAttrList") to list
```

This python test (where `calc` is the custom subclass instance) shows the situation:

```python
>>> type(calc.read_attrs), type(calc.configuration_attrs), type(list(calc.configuration_attrs))
(list, ophyd.device.Device.OphydAttrList, list)
```

This shows the problem is resolved with a simple wrapping as `list(config_attrs)`.

prjemian:
FIX #997

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1079,7 +1079,7 @@ class Device(BlueskyInterface, OphydObject):
         "Return a string summarizing the structure of the Device."
         read_attrs = self.read_attrs
         config_attrs = self.configuration_attrs
-        used_attrs = set(read_attrs + list(config_attrs))
+        used_attrs = list(read_attrs) + list(config_attrs)
         extra_attrs = [a for a in self.component_names
                        if a not in used_attrs]
         hints = getattr(self, 'hints', {}).get('fields', [])
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1079,7 +1079,7 @@ class Device(BlueskyInterface, OphydObject):
         "Return a string summarizing the structure of the Device."
         read_attrs = self.read_attrs
         config_attrs = self.configuration_attrs
-        used_attrs = set(read_attrs) and set(config_attrs)
+        used_attrs = set(read_attrs + config_attrs)
         extra_attrs = [a for a in self.component_names
                        if a not in used_attrs]
         hints = getattr(self, 'hints', {}).get('fields', [])
@@ -1495,6 +1495,9 @@ class Device(BlueskyInterface, OphydObject):
             return repr(list(self))
 
         def __add__(self, other):
+            return list(self) + list(other)
+
+        def __radd__(self, other):
             return list(self) + list(other)
 
 
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1079,7 +1079,7 @@ class Device(BlueskyInterface, OphydObject):
         "Return a string summarizing the structure of the Device."
         read_attrs = self.read_attrs
         config_attrs = self.configuration_attrs
-        used_attrs = list(read_attrs) + list(config_attrs)
+        used_attrs = set(read_attrs) and set(config_attrs)
         extra_attrs = [a for a in self.component_names
                        if a not in used_attrs]
         hints = getattr(self, 'hints', {}).get('fields', [])
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1079,7 +1079,7 @@ class Device(BlueskyInterface, OphydObject):
         "Return a string summarizing the structure of the Device."
         read_attrs = self.read_attrs
         config_attrs = self.configuration_attrs
-        used_attrs = set(read_attrs + config_attrs)
+        used_attrs = set(read_attrs + list(config_attrs))
         extra_attrs = [a for a in self.component_names
                        if a not in used_attrs]
         hints = getattr(self, 'hints', {}).get('fields', [])


jklynch:
@prjemian do you usually see test failures with python 3.6? I re-ran the tests and got different failures.

prjemian:
No, I don't use Py36 on my end now so testing with it locally is a challenge.

This error (from the CI py36 process) looks like an area detector has not been _warmed up_:

```python
 TimeoutError: Attempted to set EpicsSignalWithRBV(read_pv='ADSIM:TIFF1:EnableCallbacks_RBV', name='test_tiff1_enable', parent='test_tiff1', value='Disable', timestamp=1625075787.23148, auto_monitor=False, string=True, write_pv='ADSIM:TIFF1:EnableCallbacks', limits=False, put_complete=False) to value 1 and timed out after 10 seconds. Current value is 'Disable'.
```

Without looking at the details, it seems likely all of these are related to the same error (not related to this issue):

```python
FAILED ophyd/tests/test_areadetector.py::test_basic[pyepics] - TimeoutError: ...
FAILED ophyd/tests/test_areadetector.py::test_invalid_plugins[pyepics] - Time...
FAILED ophyd/tests/test_areadetector.py::test_default_configuration_smoke[pyepics]
FAILED ophyd/tests/test_areadetector.py::test_fstiff_plugin[pyepics-None-/tmp/ophyd_AD_test/data1/%Y/%m/%d-None-False0]
FAILED ophyd/tests/test_areadetector.py::test_fstiff_plugin[pyepics-None-/tmp/ophyd_AD_test/data1/%Y/%m/%d-None-False1]
FAILED ophyd/tests/test_areadetector.py::test_fstiff_plugin[pyepics-/tmp/ophyd_AD_test/data1-%Y/%m/%d-None-False]
FAILED ophyd/tests/test_areadetector.py::test_fstiff_plugin[pyepics-/tmp/ophyd_AD_test/data1-/tmp/ophyd_AD_test/data1/%Y/%m/%d-%Y/%m/%d-False]
FAILED ophyd/tests/test_areadetector.py::test_fstiff_plugin[pyepics-/-/tmp/ophyd_AD_test/data1/%Y/%m/%d-None-False]
FAILED ophyd/tests/test_areadetector.py::test_posix_path[pyepics-/some/path/here]
FAILED ophyd/tests/test_areadetector.py::test_posix_path[pyepics-/here/is/another/]
```

tacaswell:
I am not sure that this is the correct fix as I would expect `obj.read_attrs` to also be an OphydAttrList, I would propose

```diff
✔ 15:13:09 $ git diff
diff --git a/ophyd/device.py b/ophyd/device.py
index 04acb396..b2e3ec07 100644
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -1497,6 +1497,9 @@ class Device(BlueskyInterface, OphydObject):
         def __add__(self, other):
             return list(self) + list(other)
 
+        def __radd__(self, other):
+            return list(self) + list(other)
+
 
 # Device can be used on its own in trivial cases; ensure that it is ready
 # out-of-the-box for this scenario.

```

instead.

ophyd/device.py
```diff
@@ -1079,7 +1079,7 @@ def _summary(self):
         "Return a string summarizing the structure of the Device."
         read_attrs = self.read_attrs
         config_attrs = self.configuration_attrs
-        used_attrs = set(read_attrs + config_attrs)
+        used_attrs = list(read_attrs) + list(config_attrs)
```
mrakitin, line 1082: We lost `set` here to make a unique set of attrs. Was it intentional?

prjemian: Good catch, not sure it was intentional but now I see the value of `set()` here.

klauer:
@tacaswell has the right idea here

ophyd/device.py
prjemian: But, cannot use `+` operation with sets.  Will replace with `and`.

prjemian:
CI failures are in unrelated code due to area detector tests:

```
FAILED ophyd/tests/test_areadetector.py::test_invalid_plugins[pyepics] - Time...
FAILED ophyd/tests/test_areadetector.py::test_read_configuration_smoke[pyepics]
FAILED ophyd/tests/test_areadetector.py::test_default_configuration_smoke[pyepics]
= 3 failed, 1725 passed, 125 skipped, 1854 deselected, 1 xfailed, 4 warnings in 333.77s (0:05:33) =
Error: Process completed with exit code 1.
```

prjemian:
@klauer You're suggesting I revert that last change, then?  @mrakitin Agree?

mrakitin:
> @klauer You're suggesting I revert that last change, then? @mrakitin Agree?

@prjemian, I checked the whole `_summary` method, and the `used_attrs` is only used to identify `extra_attrs` (later printed as "unused attrs"). So I think it's fine to revert. My original concern was that there may be duplicated entries if you just add two lists together without applying `set(...)` on them to make all entries unique.

prjemian:
@mrakitin The current changes, using `set() and set()` accomplish the goal of a unique set of items.  Let's leave that in place without further changes then.  Agree?  @klauer?

ZLLentz:
The current changes no longer seem to address the issue you've referenced here. We're just making the same set in a different way, right?

@tacaswell 's suggestion seems like it's the only diff that's needed to address the linked issue.

prjemian:
I missed that the change proposed by @tacaswell does not modify _this_ line in the `_summary()` method:

```python
    used_attrs = set(read_attrs + config_attrs)
```

As I see it now, it adds the `__radd__()` method which is called to resolve this for any such use in `Device`.

mrakitin (approved):
Looks good to me now!
@klauer, @ZLLentz, are you happy with the outcome?

ZLLentz (approved):
@mrakitin I'm happy with this, we should merge to resolve the bug.

PR #999, opened 2021-07-07
MNT: set_current_position basic error handling

ZLLentz:
If the user puts in a bad value to `EpicsMotor.set_current_position`, this method can abort before restoring the `.SET` field (`set_use_switch`) to the normal value of `0` (`Use`), leaving it at the value of `1` (`Set`). This can cause confusion when someone tries to move a motor manually later, only for the position to be set instead of used.

This is just a simple try/finally block, we still raise the exception caused by the bad input but we clean up after ourselves.

### DIFF
--- a/ophyd/epics_motor.py
+++ b/ophyd/epics_motor.py
@@ -199,8 +199,10 @@ class EpicsMotor(Device, PositionerBase):
 
         '''
         self.set_use_switch.put(1, wait=True)
-        self.user_setpoint.put(pos, wait=True, force=True)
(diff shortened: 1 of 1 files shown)


klauer (approved):
Looks good to me 👍

prjemian (approved):
LGTM, thanks for the save!

mrakitin:
Was this PR number coordinated with https://github.com/bluesky/databroker/pull/666? 😆

Nice addition, @ZLLentz!

PR #1001, opened 2021-07-23
make AD Eiger SequenceId a writeable PV

JunAishima:
-  this is a writeable PV when probed with `cainfo`, so it should not be RO

Example from FMX at NSLS-II:
```bash
$ cainfo XF:17IDC-ES:FMX{Det:Eig16M}cam1:SequenceId
XF:17IDC-ES:FMX{Det:Eig16M}cam1:SequenceId
    State:            connected
    Host:             xf17id2-ioc2-1723.nsls2.bnl.local:41309
    Access:           read, write
    Native data type: DBF_LONG
    Request type:     DBR_LONG
    Element count:    1
```


mrakitin (approved):

mrakitin:
The test failure is unrelated, so I'll go ahead and merge it.

PR #1004, opened 2021-08-17
TST: test against bluesky protocols

untzag:
Adding test which explicitly checks ophyd objects against new bluesky protocols using static analysis. I'm sure there are more objects that could be checked, willing to add if desired.

### DIFF
--- a/ophyd/tests/test_bluesky_protocols.py
+++ b/ophyd/tests/test_bluesky_protocols.py
@@ -1,0 +1,92 @@
+from mypy import api
+
+
+imports = ""
+imports += "import ophyd;"
+imports += "from ophyd import flyers, sim;"
+imports += "hw = sim.hw();"
+
+
+def run(cmd):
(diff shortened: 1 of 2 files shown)


untzag:
Passes locally, don't immediately understand why it doesn't pass in CI. Don't have time to think about it at the moment, will return.

ksunden:
perhaps try putting the test file in a separate directory and running mypy from that directory?

untzag:
Played around, the solution ended up being `--follow-imports=skip`. Don't know why I don't need that locally, but if it works it works :shrug:

This is ready for review I'm curious if y'all think this is a good idea.

I've been seeing some pytest failures which I don't think is related to this change.

untzag:
Single TimeoutError for 3.6, believe it's unrelated.

This is ready for review again. Everything goes via pytest now :smile:

mrakitin (approved):
Looks good to me!

mrakitin:
@tacaswell, @danielballan, any objections to merging it?

PR #1005, opened 2021-08-23
DOC #591 add ScalerCH docs

### LINKED ISSUE #591: ScalerCH needs some documentation
There is zero documentation about this device.  It's an improvement over `EpicsScaler`, right?  Ready for production use?

prjemian:
Fix #591 (from 2018) and finally reveal the `ScalerCH` class in the documentation.  Also update the URLs to the EPICS synApps `scaler`, since in the interim, it has been moved out of the synApps `std` module.

### DIFF
--- a/docs/source/reference/builtin-devices.rst
+++ b/docs/source/reference/builtin-devices.rst
@@ -40,6 +40,8 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
    ophyd.epics_motor.EpicsMotor
    ophyd.epics_motor.MotorBundle
 
+.. _built_in.EpicsScaler:
+
 EpicsScaler
 -----------
 
@@ -51,13 +53,16 @@ Create an ``EpicsScaler`` to control an EPICS `scaler record
     from ophyd import EpicsScaler
     scaler = EpicsScaler('XF:28IDC-ES:1{Sclr:1}', name='tth')
 
-Note that :ref:`ScalerCH` is an alternative ``ophyd`` representation of
-the EPICS `scaler record.
+Note that :ref:`built_in.ScalerCH` is an alternative ``ophyd`` representation of
+the EPICS `scaler record
+<https://htmlpreview.github.io/?https://github.com/epics-modules/scaler/blob/master/documentation/scalerRecord.html>`_.
 
 .. autosummary::
    :toctree: ../generated
 
    ophyd.scaler.EpicsScaler
+
+.. _built_in.ScalerCH:
 
 ScalerCH
 --------
@@ -77,12 +82,12 @@ An important difference between the `ScalerCH` and the `EpicsScaler` is
 in how the channels names are represented.  See this table for a
 comparison:
 
-==================  =======================================  ====================================================
-class               channel naming                           examples
-==================  =======================================  ====================================================
-:ref:`EpicsScaler`  numbered                                 ``scaler_channels_chan2``, ``scaler_channels_chan3``
-:ref:`ScalerCH`     EPICS scaler record channel name fields  ``I0``, ``diode``
-==================  =======================================  ====================================================
+===========================  =======================================  ====================================================
+class                        channel naming                           examples
+===========================  =======================================  ====================================================
+:ref:`built_in.EpicsScaler`  numbered                                 ``scaler_channels_chan2``, ``scaler_channels_chan3``
+:ref:`built_in.ScalerCH`     EPICS scaler record channel name fields  ``I0``, ``diode``
+===========================  =======================================  ====================================================
 
 .. autosummary::
    :toctree: ../generated
--- a/docs/source/reference/builtin-devices.rst
+++ b/docs/source/reference/builtin-devices.rst
@@ -1,14 +1,14 @@
 Devices with Built-in Support
 =============================
 
-These devices are have ready-made classes in Python. To configure them, the
+These devices have ready-made classes in Python. To configure them, the
 user need only provide a PV prefix and a name.
 
-EpicsMotor
-----------
+EPICS motor
+-----------
 
-Create an ``EpicsMotor`` to communicate with a single `EPICS motor record
-<http://www.aps.anl.gov/bcda/synApps/motor/>`_:
+The EPICS `motor record <http://www.aps.anl.gov/bcda/synApps/motor/>`_
+is supported by the ``EpicsMotor`` object in ``ophyd``:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


mrakitin:
@prjemian, can the following error be fixed before proceeding with the reviews, please?

```
Warning, treated as error:
/home/runner/work/ophyd/ophyd/docs/source/reference/builtin-devices.rst:54:Inline interpreted text or phrase reference start-string without end-string.
make: *** [Makefile:61: html] Error 2
```

docs/source/reference/builtin-devices.rst
```diff
@@ -44,18 +44,51 @@ EpicsScaler
...
-<http://www.aps.anl.gov/bcda/synApps/std/scalerRecord.html>`_:
+<https://htmlpreview.github.io/?https://github.com/epics-modules/scaler/blob/master/documentation/scalerRecord.html>`_:
 
 .. code-block:: python
 
     from ophyd import EpicsScaler
     scaler = EpicsScaler('XF:28IDC-ES:1{Sclr:1}', name='tth')
 
+Note that :ref:`ScalerCH` is an alternative ``ophyd`` representation of
+the EPICS `scaler record.
```
mrakitin:
```suggestion
the EPICS' scaler record.
```

docs/source/reference/builtin-devices.rst
```diff
@@ -40,22 +40,60 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
...
-<http://www.aps.anl.gov/bcda/synApps/std/scalerRecord.html>`_:
+<https://htmlpreview.github.io/?https://github.com/epics-modules/scaler/blob/master/documentation/scalerRecord.html>`_:
 
 .. code-block:: python
 
     from ophyd import EpicsScaler
     scaler = EpicsScaler('XF:28IDC-ES:1{Sclr:1}', name='tth')
 
+Note that :ref:`built_in.ScalerCH` is an alternative ``ophyd`` representation of
+the EPICS `scaler record
```
mrakitin, line 57: I think sphinx is not happy with a single backtick hanging around.
```suggestion
the EPICS' scaler record
```

docs/source/reference/builtin-devices.rst
```diff
@@ -40,22 +40,60 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
...
    :toctree: ../generated
 
    ophyd.scaler.EpicsScaler
 
+.. _built_in.ScalerCH:
+
+ScalerCH
+--------
+
+Create a ``ScalerCH`` to control an EPICS `scaler record
```
mrakitin, line 70:
```suggestion
Create a ``ScalerCH`` to control an EPICS' scaler record
```

docs/source/reference/builtin-devices.rst
```diff
@@ -40,22 +40,60 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
...
+<https://htmlpreview.github.io/?https://github.com/epics-modules/scaler/blob/master/documentation/scalerRecord.html>`_:
+
+.. code-block:: python
+
+    from ophyd import ScalerCH
+    scaler = ScalerCH('XF:28IDC-ES:1{Sclr:1}', name='tth')
+
+Note that :ref:`EpicsScaler` is an alternative ``ophyd`` representation of
+the EPICS `scaler record.
```
mrakitin, line 79:
```suggestion
the EPICS' scaler record.
```

mrakitin (commented):
Left my suggestions to fix the docs building.

docs/source/reference/builtin-devices.rst
mrakitin: This one is for the older commit. Please disregard.

prjemian: That's not the fix (it is not punctuation). It's part of a reference on the next lines.  I'll look later today.

docs/source/reference/builtin-devices.rst
prjemian: same thing, not punctuation

docs/source/reference/builtin-devices.rst
prjemian: same thing

docs/source/reference/builtin-devices.rst
prjemian: same thing, might be better to switch to a named footnote, all can refer to same

mrakitin, line 79:
```suggestion
the EPICS `scaler record <https://htmlpreview.github.io/?https://github.com/epics-modules/scaler/blob/master/documentation/scalerRecord.html>`_.
```

mrakitin (commented):
Thanks, @prjemian, I did not check the next lines to understand it's a part of the reference. With my local build, sphinx complained about line 79, so it looks like the other 2 entries are correct, it's just the third one is missing the URL. Here is a straightforward (although repeating) suggestion to fix it.

But I better like your idea of having a named reference.

docs/source/reference/builtin-devices.rst
prjemian: The URL grew too complicated to expect that further edits will not become problems.  I'm refactoring the scaler support into a *EPICS scaler record* section with subsections for *EpicsScaler* and *ScalerCH*.  People need guidance about which to choose (since there are two) and that should come first.

prjemian:
- Py3.6 (pull_request) failed due to AD (not this work).
- Py3.7 (push) failed due to attempt to access obj.value which has not gotten value from the control system yet. (not this work).

Ready for review now.

docs/source/reference/builtin-devices.rst
```diff
@@ -40,22 +40,38 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
...
 
+.. _built_in.Epics.Scaler.record:
+
+EPICS Scaler
+------------
+
+The EPICS ``scaler`` record ([#scaler]_) is supported by two alternative
+``ophyd`` Devices.  (You need only choose one of these two.)  An important
+difference between the :ref:`built_in.EpicsScaler` and the :ref:`built_in.ScalerCH`
+Devices is in how each channel's names is represented, as summarized
```
mrakitin, line 51: "names are represented" or "name is represented"?

docs/source/reference/builtin-devices.rst
```diff
@@ -40,22 +40,38 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
    ophyd.epics_motor.EpicsMotor
    ophyd.epics_motor.MotorBundle
 
+.. _built_in.Epics.Scaler.record:
```
mrakitin, line 43: Is it a convention to point to the part of the documentation within the same doc (such as `#<...>` in HTML)?

mrakitin (approved):
Good addition! I have a single optional comment/suggestion.

docs/source/reference/builtin-devices.rst
prjemian: good catch

docs/source/reference/builtin-devices.rst
prjemian: Envisioning someone who searched for one of the two scaler support classes.  The link is used to find the other (in case it was off-screen).  The NeXus manual, for example, is rife with these links.  The rendered view is not distracting.

The fiendishly-long URL to the scaler documentation is a growing pain as synApps migrates into Github-style documentation.  Original docs are in HTML.  Github shows the html source code so an additional handler (htmlpreview.github.io) is needed to render the page.

prjemian:
ready for review

docs/source/reference/builtin-devices.rst
mrakitin: Thanks for the explanation. I just wonder why the name is different. I thought prepending the anchor with `_` would be enough, however, that should look like `_built_in.EpicsScaler` (without the dot between `Epics` and `Scaler` and `.record`). I don't see how `built_in.Epics.Scaler.record` would be used in the references below.

docs/source/reference/builtin-devices.rst
```diff
@@ -40,32 +40,71 @@ This must be sub-classed (like :class:`~ophyd.device.Device`) to be useful.
...
+``ophyd`` Devices.  (You need only choose one of these two.)  An important
+difference between the :ref:`built_in.EpicsScaler` and the :ref:`built_in.ScalerCH`
+Devices is in how each channel's name is represented, as summarized 
+in the next table:
+
+===========================  =======================================  ====================================================
+class                        channel naming                           examples
+===========================  =======================================  ====================================================
+:ref:`built_in.EpicsScaler`  numbered                                 ``scaler_channels_chan2``, ``scaler_channels_chan3``
+:ref:`built_in.ScalerCH`     EPICS scaler record channel name fields  ``I0``, ``diode``
```
mrakitin, line 58: Should not those two refs point to the classes in https://github.com/bluesky/ophyd/blob/master/ophyd/scaler.py like in the example of EpicsMotor above?

mrakitin (commented):
A couple of more comments regarding the references, mostly for my understanding to where you want to point the reader. Thanks!

docs/source/reference/builtin-devices.rst
prjemian: The extra anchor is not referenced as of now.  It was added to differentiate between the two sections _EPICS scaler_ and _EpicsScaler_.  By default, Sphinx creates anchors for every section based on a simple conversion of the section title to lower case and handling replacement of non-allowed characters.  The anchors in this case would be `epics_scaler` and `epicsscaler`, respectively.  These are too similar to be useful anchors.  The custom anchors avoid that ambiguity by prefixing the document name, delimiting with `.` (which is an allowed character in an anchor), and suffixing for clarity.

Since the anchor is not used, it could easily be removed.  I suggest leaving it in should a need arise in the future for a reference such as this:

    :ref:`built_in.Epics.Scaler`

docs/source/reference/builtin-devices.rst
prjemian: That's a good addition.

prjemian: The table will refer to the subsections.  Each subsection will point to its class API documentation which, in turn, provides a `[source]` link to a view of the Python code.

prjemian: @mrakitin Please check again.  Merge if OK.  Thanks!

docs/source/reference/builtin-devices.rst
mrakitin: Sounds good, thanks for the explanation!

mrakitin (approved):

prjemian:
@mrakitin Thanks for the thoughtful review!

mrakitin:
Always welcome, @prjemian! Thanks for the good explanations and improvements to the missing docs.

PR #1010, opened 2021-08-31
MNT: pin networkx >= 2.5

untzag:
versions of networkx earlier than 2.5 fail at import in Python 3.9

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
       fail-fast: false
     env:
       TEST_CL: pyepics
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 inflection
-networkx>=2.0
+networkx>=2.5
 numpy
 pint
--- a/setup.py
+++ b/setup.py
@@ -59,4 +59,6 @@ setup(name='ophyd',
           "Development Status :: 5 - Production/Stable",
           "Programming Language :: Python :: 3.6",
           "Programming Language :: Python :: 3.7",
+          "Programming Language :: Python :: 3.8",
+          "Programming Language :: Python :: 3.9",
       ])


untzag:
Don't understand the timeout errors, won't investigate further. I'll leave this open but maintainers feel free to close :smile:

tacaswell:
I am -0 on pushing the minimum version to be the minimum version of a library supported on the newest version of Python as this may be only the newest version of the dependency (for example if they were slow to account for API changes to CPython or for things with hand-rolled c extensions that may need to changes at the c Python object struct level!).

MikeHart85 (approved):
This looks good to me, and I think we should merge it so we can include it in the imminent 1.6.2 release. Fairly certain test failures were unrelated, but rerunning just in case.

`networkx 2.5` dropped support for `python 3.5`, which I think should be fine since we don't support it either.

tacaswell:
networkx 2.5 was release August 2020 so a year of support is reasonable.  I still a bit grumpy about adding floors to versions we claim to support due to their support windows rather than our demands of the API (I think reporting a lowest supported version as the first version of dependency as the oldest version that supports the oldest version of Python we support, but that floor is more for our  

This failure is on py39, I'm restarting the tests to see if it was a fluke...

prjemian:
I'm in agreement with @tacaswell , requirements are from _this_ project's point of view.  We raise the minimum when we find a limitation _in this project's code_ with the current minimum.

MikeHart85 (changes requested):
Ah, I see now. I misunderstood Tom's earlier comment.

I agree, there shouldn't be a need to raise the minimum, as there's nothing preventing 2.5 from being used with Ophyd as-is. Lower versions if networkx work fine with Ophyd on other supported Python versions.

"Requesting changes" just as a means to retract approval.

prjemian:
One type of _limitation_ could be that we just don't want to support any problems that trace back to a lower version ([hklpy did that](https://github.com/bluesky/hklpy/issues/193#issuecomment-890041029) recently).

untzag:
For what it's worth, my direct experience was to upgrade to Python 3.9 and find that I couldn't `import ophyd`. Conda was happy to let me keep networkx 2.3 even though that version is broken by changes to the stdlib `fractions` package in 3.9. Others may get caught by the same trap.

mrakitin:
The [latest CI run](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=310034&view=logs&jobId=656edd35-690f-5c53-9ba3-09c10d0bea97&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d) on conda-forge feedstock seemed to pick networkx 2.5+ successfully for Python 3.9 (fresh install, not upgrade).

tacaswell:
When pinning versions I think there are two extremes.  

On one side there are libraries which I think should pin as _lightly_ as they possible can (basically only forbid things that are known to not work (either because they lack a feature we know we want or have known bugs we do not want to work-around for setting a floor / banning specific versions or because we _know_ there is a API change in a newer version that will break us to put in a ceiling).   The goal is to exclude things you know will not work with out needlessly trapping your (future) users in constraints that do not have a strict purpose.  

On the other side we have application pinning where you should PIN EVERYTHING because your goal is to define an environment that is guaranteed to work!  The management of the meta-data on binaries is a problem for the packaging tools (it sounds like you upgraded python in-place in a conda env !?) to get right and making sure any given users environment works is on the user.

I think a lot of the pain / dependency quagmires you can get your self into are when you start to soften from either of those stances (libraries start to try to be "helpful" and pin dependencies tightly) or applications are just like "YOLO, any version of X!" without testing that it works.

Issue #1011, opened 2021-08-31
Move `inflection` requirement from `requirements.txt`


mrakitin:
See relevant discussion in https://github.com/conda-forge/ophyd-feedstock/pull/31#issuecomment-909593909.

PR #975, opened 2021-03-09
Add simple script for helping generate boilerplate detector and cam classes

jwlodek:
I'm not sure if this should live in this ophyd repository, but essentially this script allows for generating a bunch of boilerplate for new Detector and Cam classes by simply parsing the `.template` files for a specified areaDetector driver. To get an idea of what it does, you can take a look at a PR I am making adding the generated classes (with some minor hand-done fixes after) to add ADPICam detector and cam classes.

The ADPICam driver: https://github.com/areaDetector/ADPICam

I think long term this script will be replaced by some other, better, solution, like utilizing [PVI](https://pvi.readthedocs.io/en/latest/) from Diamond, but for now this should really simplify adding support for any new drivers.

PR #1014, opened 2021-10-01
PKG: move `inflection` package from requirements.txt

mrakitin:
We don't have `requirements-dev.txt` file in this repo, so I moved it to the best match `requirements-test.txt`.

Closes https://github.com/bluesky/ophyd/issues/1011.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,1 @@
+inflection
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -5,6 +5,7 @@ databroker>=1.0.0b1
 epics-pypdb
 flake8
 h5py
+inflection
 matplotlib
 pyepics>=3.4.2
 pytest
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -5,7 +5,6 @@ databroker>=1.0.0b1
 epics-pypdb
 flake8
 h5py
-inflection
 matplotlib
 pyepics>=3.4.2
 pytest
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-inflection
 networkx>=2.0
 numpy
 pint
--- a/scripts/collect_ad_boilerplate.py
+++ b/scripts/collect_ad_boilerplate.py
@@ -8,7 +8,11 @@ import logging
 
 def write_detector_class(boilerplate_file, dev_name, det_name, cam_name):
     """
-    Writes boilerplater 'Detector' class for ophyd/areadetector/detectors
+    Writes boilerplate 'Detector' class for ophyd/areadetector/detectors.
+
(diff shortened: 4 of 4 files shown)


jwlodek (approved):

dmgav (approved):

mrakitin:
> I would be fine with this change but, on inspection, cannot find where `inflection` is used in `ophyd`. It's in `scripts/collect_ad_boilerplate.py` but that is a suggestion for code to be used, isn't it? It is not a unit test.
> 
> So, I can't approve yet.

Yes, it's used in https://github.com/bluesky/ophyd/blob/dd4b3e389a0202ecacce39fc3965d703c616b0d4/scripts/collect_ad_boilerplate.py#L5 and I mistakenly guided @jwlodek to put the dependency into the `requirements.txt` rather than in a separate place while he was working on https://github.com/bluesky/ophyd/pull/975.

@prjemian, what would be your suggestions? Another `requirements-dev.txt` file with that single dependency? I don't think it should stay in `requirements.txt`...

prjemian:
How is file `scripts/collect_ad_boilerplate.py` used?  Looks as if it is an example.  If that is the case, it is either a conditional requirements for `scripts` or additional documentation in this file.  As it stands, it is not necessary for use or testing of `ophyd`.

jwlodek:
The script is used to auto-create ophyd classes for areaDetector drivers by scraping their `*.template` files. I wrote it mostly because I was writing classes for the `ADPICam` driver and it had over 200 PVs, so it felt like it made sense to just automate the process. I've used it to add a few other detector classes since then as well.

prjemian:
@jwlodek That's a good addition for the [docstring](https://github.com/bluesky/ophyd/blob/dd4b3e389a0202ecacce39fc3965d703c616b0d4/scripts/collect_ad_boilerplate.py#L11).

This seems a justification for inclusion in `requirements-dev.txt`

prjemian:
revised doc string

```python
    Writes boilerplate 'Detector' class for ophyd/areadetector/detectors.

    This script is automates the creation of ophyd classes for areaDetector drivers by scraping
    their *.template files. It is called by developers as needed.
```

ZLLentz (approved):
I agree with @prjemian that `requirements-dev.txt` would be most appropriate, but I'm just happy to see it removed from `requirements.txt` to make the setup.py requirements list accurate again.

mrakitin:
I moved `inflection` to `requirements-dev.txt` (new file) and updated the docstring as discussed. @prjemian, @ZLLentz, are you OK with that outcome?

prjemian (approved):
LGTM, Thanks!

jwlodek (approved):

mrakitin:
Neat! @ZLLentz, you have the honor of merging it then 😄.

PR #1016, opened 2021-10-05
Added release notes for v1.6.3

MikeHart85:
Release notes for (tentatively) 1.6.3. 

Patch-level increment proposed, since there are only a few small updates.

Release date prospectively set to Oct 8, since that's when release tagging is scheduled.


mrakitin (approved):
Looks good to me, thanks for handling it, @MikeHart85!

PR #1022, opened 2021-12-10
Make DeviceStatus report a fraction 0 rather than 1 once complete

callumforrester:
`MoveStatus` reports a `fraction` starting at 1 and tending towards 0 as the underlying operation progresses toward its goal. You would expect a `MoveStatus` to report a final fraction of 0, but instead it reports a 1. This is because it inherits from `DeviceStatus`, which reports a fraction of 1 as part of its `_settled` hook. 

One class treats `fraction` as "how far have we come?". The other treats it as "how far do we have to go?". This PR takes the behaviour of `MoveStatus` as canonical and modifies `DeviceStatus` in accordance.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -625,7 +625,7 @@ class DeviceStatus(StatusBase):
     def _settled(self):
         '''Hook for when status has completed and settled'''
         for watcher in self._watchers:
-            watcher(name=self.device.name, fraction=1)
+            watcher(name=self.device.name, fraction=0.)
 
     __repr__ = __str__
 
--- a/ophyd/tests/test_epicsmotor.py
+++ b/ophyd/tests/test_epicsmotor.py
@@ -274,7 +274,7 @@ def test_watchers(motor):
     st.add_callback(callback)
     ev.wait()
     assert collector
-    assert collector[-1] == 1
+    assert collector[-1] == 0.
     assert len(collector) > 1
 
 


tacaswell (approved):
I want eyes from another facility than NSLS-II on this before we merge.

tacaswell:
This is a minor API change and I can imagine that it could surprise someone, I don't think we need everyone I tagged to sign off (and my choice of people was arbitrary), but we do need a to make sure everyone has a fair chance to object ;).

mrakitin (approved):

ZLLentz (approved):
This makes sense to me and will not break any LCLS code

prjemian (approved):
I do not foresee this causing any problems at APS.

untzag (approved):
Fine with me. Given that we were surprised by this inconsistency, is this a place where additional documentation and testing would serve us well? I can't see anywhere where we've written down the expected behavior of `fraction` (haven't looked too hard).

tacaswell:
"power cycled" to pickup recent improvements to CI.

tacaswell:
The failure looks real.  It may be a test that has landed on main from when the PR was opened (the tests run an the proposed merged commit not the tip of this branch).

callumforrester:
@tacaswell Not sure why 3.7 would hang, would you mind rerunning?

tacaswell:
Merged!  This may be the highest number of approvals per LoC changed we have had in ophyd yet.

PR #1025, opened 2022-01-11
BUG: Fixed fraction in ADTriggerStatus to match docs

DominicOram:
Fixes the ADTriggerStatus to match the behaviour as documented in https://github.com/bluesky/bluesky/pull/1496/files

Also, added a test for this behaviour.


danielballan (approved):

danielballan:
@DominicOram Would you mind rebasing to fix the merge conflict? Should be a simple one. Sorry that this sat here way too long.

PR #1028, opened 2022-02-09
ENH: add a "busy" device

tacaswell:
We had a request for a way to set a sleep in the RE that would give a
progress bar.

You can get an interuptable sleep via `bps.sleep`, but that provides no
feedback.  Rather than trying to invent another way to do progress bars, re-use
the progress bar machinery for hardware motion.

This still needs tests, but @bruceravel is using this at his beamline so I wanted to get a PR for ASAP.

### DIFF
--- a/ophyd/busy.py
+++ b/ophyd/busy.py
@@ -1,0 +1,72 @@
+import time
+import threading
+
+from ophyd.device import Device
+from ophyd.status import DeviceStatus
+
+
+class BusyStatus(DeviceStatus):
+    """
+    A "busy" device that takes a fixed amount of time in seconds to report complete.
+
+    The clock starts as soon as the device is created (so do not hold onto these things!)
+
+    Parameters
+    ----------
+    device : Device
+        This object this status object belongs to
+
+    delay : float
+        Total delay in seconds
+
+    tick : float, default=0.1
+        Time between updating the
+
+    """
+
+    def __init__(self, device, delay, *, tick=0.1, **kwargs):
+        super().__init__(device, **kwargs)
+        start = time.monotonic()
+        deadline = start + delay
+
+        def busy_status_loop():
+            current = time.monotonic()
+            while current < deadline:
+                elapsed = current - start
+                for w in self._watchers:
+                    w(
+                        name=self.device.name,
+                        curent=elapsed,
+                        initial=0,
+                        target=delay,
+                        units="s",
+                        fraction=(1 - elapsed / delay),
+                        time_elapsed=elapsed,
+                        time_remaining=delay - elapsed,
+                    )
+                time.sleep(tick)
+                current = time.monotonic()
+            w(
+                name=self.device.name,
+                curent=delay,
+                initial=0,
+                target=delay,
+                unit="s",
+                fraction=0,
+                time_elapsed=current - start,
+                time_remaining=0,
+            )
+            self.set_finished()
+
+        threading.Thread(target=busy_status_loop).start()
+
+
+class Busy(Device):
+    """
+    A "busy" device that takes a fixed amount of time in seconds to report complete.
+
+
+    """
+
+    def set(self, delay):
+        return BusyStatus(self, delay, tick=max(1, min(0.1, delay / 100)))


prjemian:
Different than support for the EPICS busy record from synApps.  https://github.com/BCDA-APS/apstools/blob/main/apstools/synApps/busy.py

tacaswell:
Correct, it was inspired _by_ the busy record, but re-using the name is a probably a terrible idea.  I think this is a very good idea and I can see it being used both as a way to provide a delay on its own and as a way to put a "time box" around some other activity (because it is a "settable" it can be set in parallel with either other sets or a trigger!).  \\

What should we call this instead?

 - SleepWithProgressBar
 - TimeMover
 - TimeTraveler
 - Delay
 - SleepMachine
 - ??

prjemian:
The name is ok. Just leaving a trail to thee other one.

ophyd/busy.py
```diff
@@ -0,0 +1,72 @@
...
+    Parameters
+    ----------
+    device : Device
+        This object this status object belongs to
+
+    delay : float
+        Total delay in seconds
+
+    tick : float, default=0.1
+        Time between updating the
```
mrakitin, line 23: Incomplete description here

ophyd/busy.py
```diff
@@ -0,0 +1,72 @@
...
+class BusyStatus(DeviceStatus):
+    """
+    A "busy" device that takes a fixed amount of time in seconds to report complete.
+
+    The clock starts as soon as the device is created (so do not hold onto these things!)
+
+    Parameters
+    ----------
+    device : Device
+        This object this status object belongs to
```
mrakitin, line 17: Double "this" may be hard to read.
```suggestion
        The object this status object belongs to
```

ophyd/busy.py
```diff
@@ -0,0 +1,72 @@
...
+        deadline = start + delay
+
+        def busy_status_loop():
+            current = time.monotonic()
+            while current < deadline:
+                elapsed = current - start
+                for w in self._watchers:
+                    w(
+                        name=self.device.name,
+                        curent=elapsed,
```
mrakitin, line 39:
```suggestion
                        current=elapsed,
```

ophyd/busy.py
```diff
@@ -0,0 +1,72 @@
...
+                        initial=0,
+                        target=delay,
+                        units="s",
```
mrakitin, line 42: That's what I found as a working example with the correct kwargs: https://github.com/bluesky/ophyd/blob/7937cd46e92e9ec2f7eec321dd5b1b1b71115e20/ophyd/areadetector/trigger_mixins.py#L67-L76
```suggestion
                        unit="s",
```

ophyd/busy.py
```diff
@@ -0,0 +1,72 @@
...
+                        fraction=(1 - elapsed / delay),
+                        time_elapsed=elapsed,
+                        time_remaining=delay - elapsed,
+                    )
+                time.sleep(tick)
+                current = time.monotonic()
+            w(
+                name=self.device.name,
+                curent=delay,
```
mrakitin, line 51:
```suggestion
                current=delay,
```

ophyd/busy.py
```diff
@@ -0,0 +1,72 @@
...
+
+class Busy(Device):
+    """
+    A "busy" device that takes a fixed amount of time in seconds to report complete.
+
+
+    """
+
+    def set(self, delay):
+        return BusyStatus(self, delay, tick=max(1, min(0.1, delay / 100)))
```
mrakitin, line 72: Is this `max(..., min(...))` condition even needed? With `delay=100`, it'll pick the min=0.1, and the tick will be 1. With `delay=0`, it'll pick min=0, but the tick will still be 1. Why not hard-code the `tick=1`?

Did you have another corner case in mind?

mrakitin (commented):
I like the idea! Below are a few comments/suggestions to address what Bruce pointed to.

ophyd/busy.py
tacaswell: I think I messed the logic up here.  The goal was to never tick faster that 10hz (people can not see _much_ faster and no need to stress the system) and never slower than 1hz (1s is about as long as I want to when I watch something to see if it is dead or not) and to get as close as possible to 100 updates as possible in between (because that felt like a "nice" number).

tacaswell:
I am not sure when I will have time to work on this again, I would be very 💯 👍🏻 if someone was motivated and wanted to take this PR over!

PR #1032, opened 2022-04-11
ENH: move the recursive collect_asset_docs up to Device

tacaswell:
While talking to @jklynch about how asset document collection works I noted that the recursive fallback was implemented in the `AreaDetector` related classes.  As part of a push to refactor and (hopefully) simplify how we manage external resources it would be good to pull the recursive discovery and accumulation up to the `Device` level.


ZLLentz:
This looks very sensible :+1:

Functionally, there's a small change here of only checking kind='normal'- could that be related to some of the area detector test failures?

ZLLentz:
This looks very sensible :+1:

Functionally, there's a small change here of only checking kind='normal'- could that be related to some of the area detector test failures?

tacaswell:
rebased and tweaked the test a bit.

It makes sense that collect asset docs and read should hit the same children.

ophyd/device.py
```diff
@@ -1237,6 +1237,11 @@ def read(self):
             res.update(component.read())
         return res
 
+    def collect_asset_docs(self):
```
klauer, line 1240: docstring would be nice!

tacaswell:
Yup, this is super broken and I have no idea why :(

I assume multiple inheritance is involved....

PR #1036, opened 2022-04-14
Make SynAxis produce an updating MoveStatus

callumforrester:
* Change `SynAxis` to make a `MoveStatus` rather than a `DeviceStatus`.
* Update this status periodically during motion (period is customisation with a constructor parameter, defaults to 6 times per move)
* Example use: progress bars with Bluesky

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -470,7 +470,7 @@ class SynAxis(Device):
         self._events_per_move = events_per_move
         self.egu = egu
 
-    def set(self, value: float) -> StatusBase:
+    def set(self, value: float) -> MoveStatus:
         old_setpoint = self.sim_state["setpoint"]
         distance = value - old_setpoint
         self.sim_state["setpoint"] = value
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -11,11 +11,6 @@ import weakref
 from collections import OrderedDict, deque
 from functools import partial
 from tempfile import mkdtemp
-
-from .signal import Signal, EpicsSignal, EpicsSignalRO
-from .areadetector.base import EpicsSignalWithRBV
-from .status import DeviceStatus, MoveStatus, StatusBase
-from .device import Device, Component as Cpt, DynamicDeviceComponent as DDCpt, Kind
 from types import SimpleNamespace
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


ophyd/sim.py
```diff
@@ -418,39 +420,51 @@ def readback_func(x):
         super().__init__(name=name, parent=parent, labels=labels, kind=kind,
                          **kwargs)
         self.readback.name = self.name
+        self._events_per_move = events_per_move
+        self.egu = egu
+
+    def set(self, value: float) -> None:
```
tacaswell, line 426: This returns a Status object not None

ophyd/sim.py
```diff
@@ -418,39 +420,51 @@ def readback_func(x):
         super().__init__(name=name, parent=parent, labels=labels, kind=kind,
                          **kwargs)
         self.readback.name = self.name
+        self._events_per_move = events_per_move
+        self.egu = egu
+
+    def set(self, value: float) -> StatusBase:
```
ZLLentz, line 426: Small detail: we can be more precise here
```suggestion
    def set(self, value: float) -> MoveStatus:
```

ZLLentz (approved):
I checked out this branch to give it a spin- works as advertised. This could be very useful.

We've accumulated merge conflicts since this PR was created.

klauer:
Fixed the merge conflict

callumforrester:
After some thought, I've changed the default value of `events_per_move` to 1 so that existing `SynAxis` objects behave exactly as before and deliberate intervention is needed to invoke the new behaviour (e.g. `SynAxis(name="x", events_per_move=6)`).

@tacaswell @ZLLentz Let me know if you have any thoughts on this

ophyd/sim.py
callumforrester: Interestingly I thought this wouldn't typecheck but turns out it does: https://mypy-play.net/?mypy=latest&python=3.7&gist=9b1b36bb342d46dacc8296f77746ee8e

ophyd/sim.py
```diff
@@ -422,6 +426,8 @@ def readback_func(x):
         super().__init__(name=name, parent=parent, labels=labels, kind=kind,
                          **kwargs)
         self.readback.name = self.name
+        if events_per_move < 1:
+            raise KeyError("At least 1 event per move is required")
```
klauer, line 430: `KeyError` seems like the wrong exception here; `ValueError` instead perhaps?

ZLLentz:
I think changing the default to be more similar to the original behavior is wise. Note that it isn't quite exactly the same- previously we resolved the set without creating a thread if `delay == 0`. The tests still pass so I think this is :+1:

tacaswell:
@callumforrester This needs a rebase.

callumforrester:
@tacaswell Not sure why 3.9 hung, mind restarting?

tacaswell:
Hmm, this is already restarted once.

PR #1038, opened 2022-04-18
AD's FilePath PV **must** end with a path delimiter specific to the IOC's OS

### LINKED ISSUE #1037: BUG: write_path_template MUST end with directory path delimiter
Writes to the area detector string PV `$(IOC)Hdf1:FilePath` will timeout if the string does not end with the path delimiter used by the IOC's host operating system.  This code: https://github.com/bluesky/ophyd/blob/c3ae3dac6f37a9befeed0ce80a7684dd645f4bbc/ophyd/areadetector/filestore_mixins.py#L288

continues to fail because `ret` is an instance of [`pathlib.PurePath`](https://docs.python.org/3.8/library/pathlib.html#pathlib.PurePath), where the `str()` representation does not return a path delimiter at the end of the string.  In the area detector IOC, the HDF file writing plugin will append the delimiter to the supplied path (`$(IOC)Hdf1:FilePath`) if it is not provided, then write the final value to `$(IOC)Hdf1:FilePath_RBV`.  From ophyd, since this readback value is not the same as the setpoint value it requested, ophyd waits until the timeout expires.  They will never be equal.  We continue to encounter this problem.

One fix might be to determine the `path_semantics`.  But this is not correct if the workstation running the Python code has different semantics than the IOC.  And that is the logic in the code now: https://github.com/bluesky/ophyd/blob/c3ae3dac6f37a9befeed0ce80a7684dd645f4bbc/ophyd/areadetector/filestore_mixins.py#L268-L278

Instead, to get this right, it is necessary for the Python client to inform the `FileStoreBase()` object which semantics are to be used *from the perspective of the IOC operating system*.  Then, the same decision can be modified to use the IOC's semantics and return the `write_path_template` with the proper delimiter.  Here is an example of what that return must provide:

```py
return f"{ret}{delimiter}"  # THIS is the fix, MUST end with delimiter
```

prjemian:
- FIX #1037

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -267,12 +267,15 @@ class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
         rootp = self.reg_root
         if self.path_semantics == 'posix':
             ret = PurePosixPath(self._write_path_template)
+            delimiter = "/"
         elif self.path_semantics == 'windows':
             ret = PureWindowsPath(self._write_path_template)
+            delimiter = "\\"
         elif self.path_semantics is None:
             # We are forced to guess which path semantics to use.
             # Guess that the AD driver is running on the same OS as this client.
             ret = PurePath(self._write_path_template)
+            delimiter = pathlib.os.sep
         else:
             # This should never happen, but just for the sake of future-proofing...
             raise ValueError(f"Cannot handle path_semantics={self.path_semantics}")
@@ -285,7 +288,7 @@ class FileStoreBase(BlueskyInterface, GenerateDatumInterface):
                     ('root: {!r} in not consistent with '
                      'read_path_template: {!r}').format(rootp, ret))
 
-        return str(ret)
+        return f"{ret}{delimiter}"  # per AD IOC code, MUST end with delimiter
 
     @write_path_template.setter
     def write_path_template(self, val):


tacaswell:
We have https://github.com/bluesky/ophyd/blob/7937cd46e92e9ec2f7eec321dd5b1b1b71115e20/ophyd/areadetector/paths.py#L87-L97 for dealing with the AD IOC adding a trailing deliminator.

If we are going to do this normalization here, I would much rather do it on the way in than on the way out.

prjemian:
Can you explain how `str(ret)` will ever return a trailing delimiter?  The problem appears during staging and tracks back to this line.

prjemian:
That is, we believe we have it set properly on [the way in](https://github.com/BCDA-APS/bdp_controls/blob/cf8586bcca4320685d38402aa3a3eec65021db6d/qserver/instrument/devices/area_detector.py#L38-L44)

```py
AD_IOC_MOUNT_PATH = pathlib.Path(iconfig["ADIOC_IMAGE_ROOT"])
BLUESKY_MOUNT_PATH = pathlib.Path(iconfig["BLUESKY_IMAGE_ROOT"])
IMAGE_SUBDIR = iconfig["AD_IMAGE_SUBDIR"]

# MUST end with a `/` (which pathlib will not provide)
WRITE_PATH_TEMPLATE = f"{AD_IOC_MOUNT_PATH / IMAGE_SUBDIR}/"
READ_PATH_TEMPLATE = f"{BLUESKY_MOUNT_PATH / IMAGE_SUBDIR}/"

# ...

class MySimDetector(SingleTrigger, DetectorBase):
    """ADSimDetector"""

    cam = ADComponent(MyFixedCam, "cam1:")
    image = ADComponent(ImagePlugin_V34, "image1:")
    hdf1 = ADComponent(
        MyHDF5Plugin,
        "HDF1:",
        write_path_template=WRITE_PATH_TEMPLATE,
        read_path_template=READ_PATH_TEMPLATE,
    )
    pva = ADComponent(PvaPlugin_V34, "Pva1:")
```

(and python and IOC are both of default `posix` semantics) yet we experience timeout when writing `write_path_template` with no trailing delimiter.  The trailing delimiter was traced directly to the `str(ret)` code.

prjemian:
```p
In [17]: p = pathlib.PurePath("path/with/trailing/char/")

In [18]: str(p)
Out[18]: 'path/with/trailing/char'
```

tacaswell:
https://github.com/bluesky/ophyd/blob/7909d08e28a671eea4f3499568977ebf8c1d4cc3/ophyd/areadetector/filestore_mixins.py#L431-L435
is where the standard stage method sets the filepath (note we use `obj.set(val).wait()` rather that `set_and_wait(obj, val)` ) and https://github.com/bluesky/ophyd/blob/7909d08e28a671eea4f3499568977ebf8c1d4cc3/ophyd/areadetector/plugins.py#L794-L795 is where the standard `FileBase` plugin defines this path to be 

> Can you explain how str(ret) will ever return a trailing delimiter?

fair, but the core of the problem in the IOC mutating the value and that should be handled as close to where the problem lies so that other paths that might set the file path can also benefit from it rather than relying on a property someplace else to follow a non-standard convention of always including the trailing delimiter.

If we do not want to keep fixing this where ever we set it, then I think we should try to get this fixed in AD so it does not mutate the input and add the delimiter someplace internal and not reflect this to the PVs.

prjemian:
In this line, `write_path` is the value returned from `self.write_path_template`: https://github.com/bluesky/ophyd/blob/7909d08e28a671eea4f3499568977ebf8c1d4cc3/ophyd/areadetector/filestore_mixins.py#L426

prjemian:
> ... but the core of the problem in the IOC mutating the value and that should be handled as close to where the problem lies so that other paths that might set the file path can also benefit from it rather than relying on a property someplace else to follow a non-standard convention of always including the trailing delimiter.
> 
> If we do not want to keep fixing this where ever we set it, then I think we should try to get this fixed in AD so it does not mutate the input and add the delimiter someplace internal and not reflect this to the PVs.

@tacaswell : I agree with you must whole heartedly.  When a feature such as this (the area detector FilePath feature that modifies the requested FilePath by adding a trailing directory path separator) cause problems for downstream clients, the *fix* should be applied where this editing occurs.  An error would be returned when a non-compliant `FilePath` is requested.  @MarkRivers, @ulrikpedersen: your thoughts?

However, the python code here must resolve such problems with existing area detector versions, as well as future versions.  This PR addresses the current problem in a general way at the point where the problem is created.

MarkRivers:
Please explain the issue since don't know the context.

To be clear, AD does not modify the FilePath PV itself.  If the user neglects to add the proper terminator at the end of FilePath it will be added into the value in the parameter library, and FilePath_RBV will reflect the modified value.  

IMHO the current behavior is a significant convenience for the user using the GUI because they will often forget to add the terminator, or they may add the wrong terminator for the IOC OS.

prjemian:
Context is described in #1037, that when ophyd puts to `FilePath` (without trailing separator), the put results in a timeout because ophyd is waiting for `FilePath_RBV` to become equal to the value it put to `FilePath`.

Seems we have two groups, each believing their implementation is correct.

The implementation in AD is a nice user convenience on which more than just Python users have come to rely.  Not likely to change.

The implementation in ophyd is that `set_and_wait` expects that readback will become what has been `set()`, eventually, or timeout.  This model works well in general.

Here, the assumptions in ophyd fail because ophyd has been setup to expect that `FilePath_RBV` will eventually become the value to which `FilePath` has been set.  Mostly works but one type of failure is reported here for which a repair is presented.

Another way to *repair* ophyd's this is to relax how ophyd is setup, separating the `FilePath_RBV` entirely to a separate attribute.  Then, **this** problem would go away.  Would another appear?

MarkRivers:
> The implementation in ophyd is that set_and_wait expects that readback will become what has been set(), eventually, or timeout. This model works well in general.

That is not a safe assumption for many PVs in areaDetector and other drivers.  For example, a device I was working on today, the  Measurement Computing USB-CTR08 has a "rate" parameter that specifies the sampling rate in Hz.  Because of the granularity of the clock when the requested rate is 1.15e6 the device will return the actual rate which might be 1.10e6.  The RBV value is set to the actual rate.  In this case Ophyd would also wait forever?

prjemian:
Agree with your example.  Most PVs work well with the basic model in ophyd.  Many PV pairs (PV and PV_RBV) also work well.  Some PVs, as you describe will not and need special handling.  In ophyd.  Seems that ophyd has created this problem (the expectation that the readback will match exactly what was put).  If that is the case, ophyd must be configured to resolve such minor discrepancies.

MarkRivers:
I agree.  Here is another simple example from areaDetector, setting the frame rate on an ADSpinnaker camera:
```
corvette:CARS/iocBoot/ioc13Mjolnir>caput 13BMDPG1:cam1:FrameRate 109.5
Old : 13BMDPG1:cam1:FrameRate        109
New : 13BMDPG1:cam1:FrameRate        109.5
corvette:CARS/iocBoot/ioc13Mjolnir>caget -f6 13BMDPG1:cam1:FrameRate 13BMDPG1:cam1:FrameRate_RBV
13BMDPG1:cam1:FrameRate        109.500000
13BMDPG1:cam1:FrameRate_RBV    109.664352
```

ZLLentz:
Can we make an alternative to `set_and_wait` that simply waits for the readback to change/update, but not to any particular value? This would be more appropriate for many two-PV signals, even outside the context of Area Detector. It'd be a nice non-default setting to have access to when making a device class.

Or is there some other way to configure signals that already exists? Put completion, perhaps? Right now every single `set` call routes through `set_and_wait`, as far as I can tell.

I had a similar problem recently on an unrelated IOC that I had to resolve via removing the readback side of the PV (luckily, in my case, this was appropriate).

MarkRivers:
> Can we make an alternative to set_and_wait that simply waits for the readback to change/update, but not to any particular value? 

I don't think that is possible.  In asynPortDriver if you write the same value to the PV twice the second put will not result in a callback to device support, because the value in the driver did not change.  That means the _RBV record won't process, and there won't be a callback to your client.

> Or is there some other way to configure signals that already exists? Put completion, perhaps?

put completion would probably work in some conditions, but it relies on the record calling recGblFwdLink.  In the case of complex records like the motor, mca, scaler, etc. recGblFwdLink is not called until the move or acquisition is complete.  So if you write to a motor record field while the motor is moving I don't think the put will complete until the motion stops.

tacaswell:
To my understanding there is no generic, guaranteed always to work method of telling when an IOC has actually done what you asked it to (sometimes put-complete works, sometimes literal readback works, sometimes we need software dead-bands, sometimes we need to track oscillations being within deadbands, sometimes you have to watch some unrelated PV, sometimes you have to normalize between enum values and enum strings, ...).  The purpose of the `obj.set` method is to do what ever it is that needs to be done to kick off the change and then return a `Status` object that will be marked as done.   What ever bespoke thing has to be done on the Python side to know when it is done (this can be a simple as subscribing to the put complete on the Acquire PV for AD (with the right other settings) to watching for the DMOV field to flash to trying to fit a damped oscillation to the readback trace in the case of at least one temperature controller).

This is already the code path that gets used when you do `yield from bsp.mv(obj, val)` (which waits for it to be done) or `yeield from bps.abs_set(obj, value, group=...)` (which does not wait).  I am pretty sure that way  back in the day when we started `Status` objects did not have a `wait` method and the `set_and_wait` function pre-dates the logic around `set` fully settling down (@danielballan and @klauer may remember better than I do).

I think the fix here is to leave this code and deprecate the public use of `set_and_wait(obj, value)` in favor of `obj.set(value).wait()`.  

@prjemian  If you change your usage to the new method does it work?

----

@MarkRivers I do not disagree that ensuring the implicit trailing slash gets put in the path is useful, but I do not agree it should be pushed back to the RBV.  Rather I would expect that `file_template` would require the delimiter to be included so it does not matter if the path includes the trailing delimiter or not.  However, I also see how changing that convention at this point would cause a whole lot of pain and thrashing for not much gain so we should just make sure we always use `EpicsPathSignal` in ophyd and move on.

prjemian:
I'll give it a try.

tacaswell:
I opened https://github.com/bluesky/ophyd/pull/1039 to do that deprecation.

prjemian:
> Ran tests using existing [client] and did not encounter the TimeoutError as reported. Will [close the ophyd PR] and revert the override in the bdp repository.

As stated in https://github.com/BCDA-APS/bdp_controls/issues/36#issuecomment-1102844012,

> It is important to document when an incident occurs.

Apologies for the bother.

PR #1039, opened 2022-04-18
API: deprecated use of set_and_wait function

tacaswell:
The functionality of the free function `set_and_wait(obj, val)` can be better
accomplished via `obj.set(val).wait()` and the existence of this function leads
to confusion.

### DIFF
--- a/docs/source/reference/release_notes.rst
+++ b/docs/source/reference/release_notes.rst
@@ -2,6 +2,18 @@
  Release History
 =================
 
+1.7.0 (????-??-??)
+==================
+
+Changes
+-------
+
+* Deprecated ``ophyd.utils.epics_pv.set_and_wait(obj, val)`` in favor of
+  ``obj.set(val).wait()``.  This ensures that the same logic is used to
+  determine if the signal is set as requested when used in a plan (via
+  ``yield from bps.abs_set(...)`` or ``yield from bps.mv(...)``) as from
+  within ophyd methods.
+
 
 1.6.4 (2022-04-08)
 ==================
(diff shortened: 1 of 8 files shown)


klauer (approved):
I agree that `set()` is generally preferable and am mostly in favor of this. It seems fine overall. Some thoughts:
* We do have a few direct `set_and_wait` calls in our repos which might specify their own tolerance/timeout settings.
* I also would have expected `_set_and_wait` to become part of the `Signal` class itself so that it could be overridden more easily.
* Given that ophyd v1 is effectively slated for deprecation in its entirety, I do wonder if we should be creating extra work with function-by-function deprecation.

ZLLentz (approved):
I am also in favor of this.

Quick note about overriding:

> I also would have expected _set_and_wait to become part of the Signal class itself so that it could be overridden more easily.

Signal does currently have an overridable `_set_and_wait` method. Somehow I had also missed this previously. The default `Signal._set_and_wait` simply calls `_set_and_wait`, and the `set` method will call `self._set_and_wait` in a thread.

So if we do have a poorly-behaving 2-PV signal as in the AD classes, then we absolutely can make a special handler for it by subclassing `EpicsSignal` and overriding `_set_and_wait`. I've been convinced that the details of this are very application-specific.

tacaswell:
I could be talked out of actually deprecating `set_and_wait`, but `Signal` is aware of the tolerances at it is not clear to me why you would want to change the tolerances as a function of time!

I did not move the implementation to `Signal` so that you could still do

```
try:
    import ophyd.utils.epics_pv._set_and_wait as set_and_wait
except ImportError:
    import ophyd.utils.epics_pv.set_and_wait as set_and_wait
```

at the top of a file and YOLO your way to back-compatibility.

PR #1040, opened 2022-04-25
MNT: ensure the input to set in EpicsPathSignal converted to str

tacaswell:
If a Pathlib.Path object is passed in it will fail the test and be implicitly
converted to a string, but for clarity this always tries to cast the input to
str at the top of the set_and_wait helper function.


AbbyGi (approved):
Looks good to me.

ZLLentz (approved):
I'll click the merge button here- seems to be sitting still despite looking :+1:

Issue #1021, opened 2021-11-05
Is there an ophyd.EpicsSignalWO class?


prjemian:
EPICS PV fields such as `.PROC` are signals for the EPICS server to run a function.  They never return the value that you put to them.  This breaks the current model for an ophyd EpicsSignal that requires the get to reach the put value.

tacaswell:
No, but there probably should be.

prjemian:
The `.PROC` field was a lousy example.  Better is the [`.SYNC`](https://github.com/epics-modules/motor/blob/da3bfab4c118226719f756779cc3e84a3a405ddf/motorApp/MotorSrc/motorRecord.dbd#L799-L802) field from EPICS motor.  That's the one we repeatedly run into the problem of `another call to set() in progress`.

prjemian:
To be clear (in the implied request):

> The proposed `ophyd.EpicsSignalWO` would be a write-only signal that does not check the EPICS PV for a final value.  It only does a `caput` and returns without checking for an updated value from EPICS.

tacaswell:
Can you still get put-completion callbacks on the .SYNC or .PROC fields?

prjemian:
Good question but I do not know.  But I know who to ask.

prjemian:
Might be easier just to test it though.

tacaswell:
I mean, at the protocol level, I can not image that they would reject a `WriteNotifyRequest`, the question is if the IOC does anything meaningful about it!

tangkong:
I believe this is closed by #1044 , but I don't have the authority to close the issue

mrakitin:
Closed via #1044.

PR #1044, opened 2022-05-09
ENH: Add EpicsSignalNoValidation

tangkong:
## Description
- Adds a no-validation EpicsSignal, which does not wait to verify the readback on set.

## Motivation and Context
#1021 

We're trying to work on this as part of the 2022 Epics Codeathon.

## How Has This Been Tested?
Tests added.  

## Where Has This Been Documented?
This PR, docstrings, etc

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -8,7 +8,7 @@ import weakref
 import numpy as np
 
 from .utils import (ReadOnlyError, LimitError, DestroyedError,
-                    WriteOnlyError, doc_annotation_forwarder)
+                    doc_annotation_forwarder)
 from .utils.epics_pvs import (waveform_to_string,
                               raise_if_disconnected, data_type, data_shape,
                               AlarmStatus, AlarmSeverity, validate_pv_name, _set_and_wait)
@@ -1428,8 +1428,10 @@ class EpicsSignalRO(EpicsSignalBase):
             self._run_metadata_callbacks()
 
 
-class EpicsSignalWO(EpicsSignalBase):
-    '''A write-only EpicsSignal -- that is, one with no `read_pv`
+class EpicsSignalNoValidation(EpicsSignalBase):
+    '''An EpicsSignal that does not verify values on set.
+    This signal does support readback, but does not guarantee that
+    the readback will match the set value.
 
     Keyword arguments are passed on to the base class (Signal) initializer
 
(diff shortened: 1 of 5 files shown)


tangkong:
I apologize for the tiny commits, I have not yet gotten ophyd CI to work locally.

ophyd/tests/test_signal.py
```diff
@@ -227,6 +236,24 @@ def test_epicssignal_readonly(cleanup, signal_test_ioc):
     time.sleep(0.2)
 
 
+def test_epicssignal_novalidation(nv_signal):
+    print('EpicsSignalNoValidation.metadata=', nv_signal.metadata)
+
+    nv_signal.put(10)
+    st = nv_signal.set(11)
+
+    time.sleep(0.05)
```
mrakitin, line 245: As the waiting is happening on the object level already, it probably makes sense to remove this sleep.
```suggestion
```

tangkong:
Tests seem to be stuck, maybe an issue similar to what Zach was seeing on his PR's?

mrakitin (approved):
Looks good to me! We did some pair-coding/development on this with @tangkong for a bit.

The test failures seem to be transient.

ophyd/signal.py
```diff
@@ -1428,6 +1428,59 @@ def _pv_access_callback(self, read_access, write_access, pv):
...
+        '''
+        Set the value of this signal, and return a completed Status
+        object, bypassing any readback verification
+
+        Returns
+        -------
+        st : Status
+            This status object will be finished
+        '''
+        super().set(value)
```
ZLLentz, line 1476: This leads us through the following code path:
https://github.com/bluesky/ophyd/blob/0b1570d08d6dd2616d041646171886ec737880db/ophyd/signal.py#L290-L292

Which means we'll create a set_and_wait thread that may never resolve (We picked write-only, right? There's something weird with our PV, probably), culminating in:
https://github.com/bluesky/ophyd/blob/0b1570d08d6dd2616d041646171886ec737880db/ophyd/signal.py#L322-L324
I think this PR needs to override `_set_and_wait` or needs to route around the `super().set` call entirely

ophyd/signal.py
```diff
@@ -1428,6 +1428,59 @@ def _pv_access_callback(self, read_access, write_access, pv):
...
+        object, bypassing any readback verification
+
+        Returns
+        -------
+        st : Status
+            This status object will be finished
+        '''
+        self.put(value)
+
+        st = Status()
```
ZLLentz, line 1478: One last small suggestion from me: link this back to the signal object for better str on the status
```suggestion
        st = Status(self)
```
ref https://github.com/bluesky/ophyd/blob/0b1570d08d6dd2616d041646171886ec737880db/ophyd/status.py#L571-L576

ZLLentz (approved):

ZLLentz:
Ok let's get this in :+1:

ophyd/signal.py
mrakitin: That's a good point I overlooked, thanks for the suggestion, @ZLLentz :)

PR #1046, opened 2022-05-10
ENH: Upstream helpful PV Positioners, add wait_for_value util

ZLLentz:
Part of the May 2022 EPICS codeathon effort
Relies on the `InternalSignal` from #1045, which would need to be merged first

- Adds `PVPositionerIsClose`, a helpful PV positioner for when your IOC doesn't supply a `done` signal.
- Adds `PVPositionerDone`, a dummy class for when the user wants their signal to "look like" a motor and cannot be dissuaded.
- split out the "waiting" part of `ophyd.utils.epics_pvs._set_and_wait` into its own standalone utility, `wait_for_value` because I needed it to write a reasonable test

### DIFF
--- a/ophyd/pv_positioner.py
+++ b/ophyd/pv_positioner.py
@@ -309,8 +309,7 @@ class PVPositionerComparator(PVPositioner):
 
     The done state is set by a comparison function defined in the class body.
     The comparison function takes two arguments, readback and setpoint,
-    returning True if we are close enough to be considered done or False if we
-    are too far away.
+    returning True if we are considered done or False if we are not.
 
     This class is intended to support `PVPositionerIsClose`, but exists to
     allow some flexibility if we want to use other metrics for deciding if
--- a/ophyd/pv_positioner.py
+++ b/ophyd/pv_positioner.py
@@ -1,6 +1,7 @@
 # vi: ts=4 sw=4
 
 import logging
+from typing import Any, Optional
 
 import numpy as np
 
@@ -10,7 +11,7 @@ from .device import Component as Cpt
 from .device import (Device, required_for_connection)
 from .ophydobj import Kind
 from .positioner import PositionerBase
-from .signal import InternalSignal
+from .signal import EpicsSignal, InternalSignal
 from .status import wait as status_wait
 
 logger = logging.getLogger(__name__)
@@ -305,34 +306,31 @@ class PVPositionerPC(PVPositioner):
 class PVPositionerComparator(PVPositioner):
     """
     PV Positioner with a software done signal.
+
     The done state is set by a comparison function defined in the class body.
     The comparison function takes two arguments, readback and setpoint,
     returning True if we are close enough to be considered done or False if we
     are too far away.
-    """
 
-    # Override setpoint, readback in subclass
-    setpoint = None
-    readback = None
+    This class is intended to support `PVPositionerIsClose`, but exists to
+    allow some flexibility if we want to use other metrics for deciding if
+    the PVPositioner is done.
+
+    Internally, this will subscribe to both the `setpoint` and `readback`
+    signals, updating `done` as appropriate.
+    """
 
     done = Cpt(InternalSignal, value=0)
     done_value = 1
 
-    # Optionally override limits to a 2-element tuple in subclass
-    limits = None
-
-    def __init__(self, prefix, *, name, **kwargs):
-        self._last_readback = None
-        self._last_setpoint = None
-        super().__init__(prefix, name=name, **kwargs)
+    def __init__(self, prefix: str, *, name: str, **kwargs):
         if None in (self.setpoint, self.readback):
             raise NotImplementedError('PVPositionerComparator requires both '
                                       'a setpoint and a readback signal to '
                                       'compare!')
-
-    def done_comparator(self, readback, setpoint):
-        """Override done_comparator in subclass."""
-        raise NotImplementedError('Must implement a done comparator!')
+        self._last_readback = None
+        self._last_setpoint = None
+        super().__init__(prefix, name=name, **kwargs)
 
     def __init_subclass__(cls, **kwargs):
         """Set up callbacks in subclass."""
@@ -341,7 +339,16 @@ class PVPositionerComparator(PVPositioner):
             cls.setpoint.sub_value(cls._update_setpoint)
             cls.readback.sub_value(cls._update_readback)
 
-    def _update_setpoint(self, *args, value, **kwargs):
+    def done_comparator(self, readback: Any, setpoint: Any) -> bool:
+        """
+        Override done_comparator in your subclass.
+
+        This method should return True if we are done moving
+        and False otherwise.
+        """
+        raise NotImplementedError('Must implement a done comparator!')
+
+    def _update_setpoint(self, *args, value: Any, **kwargs) -> None:
         """Callback to cache the setpoint and update done state."""
         self._last_setpoint = value
         # Always set done to False when a move is requested
@@ -350,12 +357,12 @@ class PVPositionerComparator(PVPositioner):
         self.done.put(0, internal=True)
         self._update_done()
 
-    def _update_readback(self, *args, value, **kwargs):
+    def _update_readback(self, *args, value: Any, **kwargs) -> None:
         """Callback to cache the readback and update done state."""
         self._last_readback = value
         self._update_done()
 
-    def _update_done(self):
+    def _update_done(self) -> None:
         """Update our status to done if we pass the comparator."""
         if None not in (self._last_readback, self._last_setpoint):
             is_done = self.done_comparator(self._last_readback,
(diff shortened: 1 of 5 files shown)


ophyd/pv_positioner.py
```diff
@@ -296,3 +301,150 @@ def done_moving(**kwargs):
...
+        """Callback to cache the readback and update done state."""
+        self._last_readback = value
+        self._update_done()
+
+    def _update_done(self) -> None:
+        """Update our status to done if we pass the comparator."""
+        if None not in (self._last_readback, self._last_setpoint):
+            is_done = self.done_comparator(self._last_readback,
+                                           self._last_setpoint)
+            self.done.put(int(is_done), internal=True)
```
tacaswell, line 370: Is it worth only updating if it has changed?  I am 95% sure that the down-stream callbacks are going to fire on over put, I'm not sure ("not done yet", "not done yet", ... )  or ("done", "done", ...) is interesting to most things that would subscribe to the done status?

tacaswell:
The test failure is from the new tests, but looks like CI related timing issues....

ophyd/pv_positioner.py
ZLLentz: Yes, it's worth modifying this so that it only updates when changed. It's easy to do, makes the situation more clear, and removes burden on downstream callbacks. I'll make this change.

ZLLentz:
I'll modify the test a bit to make it less susceptible to timing errors. I will claim that these tests passed locally every time.

ZLLentz:
Rebased on current master

ZLLentz:
We're having a strange issue here where the 3.8 build is getting permanently stuck without any apparent text output. Yesterday this happened to a 3.9 build and I only noticed once the runtime had crept up to 2 hours with no activity. I'm not sure how to investigate this.

tacaswell:
I manually cancled it and we got

```
coverage run --concurrency=thread --parallel-mode -m pytest -v -k "${TEST_CL}"
+ coverage run --concurrency=thread --parallel-mode -m pytest -v -k pyepics
============================= test session starts ==============================
platform linux -- Python 3.8.12, pytest-7.1.2, pluggy-1.0.0 -- /opt/hostedtoolcache/Python/3.8.12/x64/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/ophyd/ophyd, configfile: pytest.ini
plugins: rerunfailures-10.2
collecting ... collected 3738 items / 1869 deselected / 1869 selected

ophyd/tests/test_areadetector.py::test_basic[pyepics] PASSED             [  0%]
ophyd/tests/test_areadetector.py::test_stubbing[pyepics] SKIPPED (co...) [  0%]
ophyd/tests/test_areadetector.py::test_detector[pyepics] PASSED          [  0%]
ophyd/tests/test_areadetector.py::test_tiff_plugin[pyepics] PASSED       [  0%]
ophyd/tests/test_areadetector.py::test_hdf5_plugin[pyepics] PASSED       [  0%]
ophyd/tests/test_areadetector.py::test_subclass[pyepics] PASSED          [  0%]
ophyd/tests/test_areadetector.py::test_getattr[pyepics] PASSED           [  0%]
ophyd/tests/test_areadetector.py::test_invalid_plugins[pyepics] PASSED   [  0%]
ophyd/tests/test_areadetector.py::test_validate_plugins_no_portname[pyepics] PASSED [  0%]
ophyd/tests/test_areadetector.py::test_get_plugin_by_asyn_port[pyepics] PASSED [  0%]
ophyd/tests/test_areadetector.py::test_visualize_asyn_digraph_smoke[pyepics] PASSED [  0%]
ophyd/tests/test_areadetector.py::test_read_configuration_smoke[pyepics] PASSED [  0%]
ophyd/tests/test_areadetector.py::test_str_smoke[pyepics] PASSED         [  0%]
```

which suggests the problem is the next test.

tacaswell:
Which looking at a different run (and assuming they are deterministically orderded) gives us 

```
ophyd/tests/test_areadetector.py::test_default_configuration_smoke[pyepics] PASSED [  0%]
```

and then a block of area detector tests...which suggests we need better timeouts on those?

ZLLentz:
I think I agree, it must be area detector tests getting stalled out. I'm surprised that GH actions doesn't give us partial text output from these stages and requires a cancellation to see where we are stuck.

I don't think this is related to this PR- I see the same thing in the log statements PR.

ZLLentz:
This actually stalls even earlier in the logs PR:
```
coverage run --concurrency=thread --parallel-mode -m pytest -v -k "${TEST_CL}"
+ coverage run --concurrency=thread --parallel-mode -m pytest -v -k pyepics
============================= test session starts ==============================
platform linux -- Python 3.9.[12](https://github.com/bluesky/ophyd/runs/6375385206?check_suite_focus=true#step:7:12), pytest-7.1.2, pluggy-1.0.0 -- /opt/hostedtoolcache/Python/3.9.12/x64/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/ophyd/ophyd, configfile: pytest.ini
plugins: rerunfailures-10.2
collecting ... collected 3734 items / [18](https://github.com/bluesky/ophyd/runs/6375385206?check_suite_focus=true#step:7:18)67 deselected / 1867 selected

ophyd/tests/test_areadetector.py::test_basic[pyepics] PASSED             [  0%]
ophyd/tests/test_areadetector.py::test_stubbing[pyepics] SKIPPED (co...) [  0%]
ophyd/tests/test_areadetector.py::test_detector[pyepics] PASSED          [  0%]
ophyd/tests/test_areadetector.py::test_tiff_plugin[pyepics] PASSED       [  0%]
```

ophyd/pv_positioner.py
```diff
@@ -296,3 +301,151 @@ def done_moving(**kwargs):
...
+
+    The done state is set by a comparison function defined in the class body.
+    The comparison function takes two arguments, readback and setpoint,
+    returning True if we are considered done or False if we are not.
+
+    This class is intended to support `PVPositionerIsClose`, but exists to
+    allow some flexibility if we want to use other metrics for deciding if
+    the PVPositioner is done.
+
+    Internally, this will subscribe to both the `setpoint` and `readback`
```
klauer, line 318:
```suggestion
    Internally, this will subscribe to both the ``setpoint`` and ``readback``
```

(Single backtick means an `:any:` reference, right? If I'm misremembering, ignore this)

ophyd/pv_positioner.py
```diff
@@ -296,3 +301,151 @@ def done_moving(**kwargs):
...
+    signals, updating `done` as appropriate.
+    """
```
klauer, line 320: Should add the parameters to the docstring here.

ophyd/pv_positioner.py
```diff
@@ -296,3 +301,151 @@ def done_moving(**kwargs):
...
+
+    atol is a measure of absolute tolerance. If atol is 0.1, then you'd be
+    able to be up to 0.1 units away and still count as done. This is
+    typically the most useful parameter for calibrating done tolerance.
+
+    rtol is a measure of relative tolerance. If rtol is 0.1, then you'd be
+    able to deviate from the goal position by up to 10% of its value. This
+    is useful for small quantities. For example, defining an atol for a
+    positioner that ranges from 1e-8 to 2e-8 could be somewhat awkward.
+    """
```
klauer, line 393: Parameters in docstring again

ophyd/pv_positioner.py
```diff
@@ -296,3 +301,151 @@ def done_moving(**kwargs):
...
+
+class PVPositionerDone(PVPositioner):
+    """
+    PV Positioner with no readback that reports done immediately.
+
+    This is for the case where you'd like a PV to look like a
+    positioner, but the truth is that it is just a PV.
+
+    When the user asks for a move, set done to 0 and then back to 1.
+    This simluates the normal positioner behavior.
```
klauer, line 441:
```suggestion
    This simulates the normal positioner behavior.
```

ophyd/utils/epics_pvs.py
```diff
@@ -232,6 +233,34 @@ def _set_and_wait(signal, val, poll_time=0.01, timeout=10, rtol=None,
     TimeoutError if timeout is exceeded
     """
     signal.put(val, **kwargs)
+    wait_for_value(signal, val, poll_time=poll_time, timeout=timeout, rtol=rtol, atol=atol)
+
+
+def wait_for_value(signal, val, poll_time=0.01, timeout=10, rtol=None,
```
klauer, line 239: `set_and_wait` is now internal API. Should this be the same?

klauer (approved):
This LGTM overall 👍

ophyd/pv_positioner.py
ZLLentz: Looks like default_role is not set for ophyd: https://github.com/bluesky/ophyd/blob/cf8e8750e07689fe8335bdf477ad914cd5c31d3e/docs/source/conf.py#L81-L82

ophyd/utils/epics_pvs.py
ZLLentz: Need to think on this for a moment. Unlike `set_and_wait`, `wait_for_value` should always have the desired behavior in any context: "wait for the value of this signal to be this value", whereas for `set_and_wait` it had weird interaction specifically with signal `set` methods, particularly when we need to customize `_set_and_wait`. I could see an argument for `wait_for_value` being an overridable method on `Signal` objects, I'm not sure I'm ready to commit to a change like that in this PR? If that's desirable in the future, it could be wise to preemptively internalize this.

ophyd/pv_positioner.py
ZLLentz: I when-in-rome'd a bit too much here- a lot of the surrounding code was also missing parameters. I'll update.

ophyd/utils/epics_pvs.py
klauer: I don't have a feel for how these might be used, if I'm being honest. The only thing I could say is that it's easier to go from private->public than it is public->private in a library.

PR #1047, opened 2022-05-10
MAINT: reduce log level of spammy self.log.info log messages to debug

ZLLentz:
This is part of the May 2022 EPICS Codeathon effort

I've taken all of the `self.log.info` log messages and dropped them to `debug` level.
At LCLS we've instituted a log filter that drops all of these from info to debug because they are too noisy to be useful. Semantically, I take `INFO` to be equivalent to a logged `print` statement, so I claim that anything that would be annoying in a `print` statement should be dropped to `DEBUG` level.

I'm going to talk through each of these calls and why I want to reduce the log level, in case there is disagreement or misunderstanding of use cases:

- ophyd/ophydobj.py
  - "first instance of OphydObject: id=%s"
    - I'd never put something like this a print statement, it's more of a debug helper for developers. The user doesn't need to care about object ids or which ophyd object was created first.
- ophyd/signal.py
  -  'set(value=%s, timeout=%s, settle_time=%s, kwargs=%s)'
     - This is run every time we `set` any signal, which seems excessive for a print-equivalent.
  - '%s: _set_and_wait(value=%s, timeout=%s, atol=%s, rtol=%s, kwargs=%s) succeeded => %s'
    - This is run when the set succeeds, which makes it the second print-equivalent we typically see on `set`
  - 'settling for %d seconds'
    - This is run if the signal has a settling time after the set succeed, making it the third print-equivalent we can see on `set`
- ophyd/sim.py
  - 'sleep_and_finish %s'
    - This is run every time we trigger a `SynSignal`, which seems a big excessive if you're running bluesky plans, etc., it's something you'd only really want while debugging.

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -186,7 +186,7 @@ class OphydObject:
         self.control_layer_log = LoggerAdapter(control_layer_logger, {'ophyd_object_name': name})
 
         if not self.__any_instantiated:
-            self.log.info("first instance of OphydObject: id=%s", id(self))
+            self.log.debug("first instance of OphydObject: id=%s", id(self))
             OphydObject._mark_as_instantiated()
         self.__register_instance(self)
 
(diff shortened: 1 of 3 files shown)


tacaswell (approved):

ZLLentz:
CI stalls out on the area detector tests again

PR #1052, opened 2022-06-09
PERF: set subscriptions only once per class

klauer:
`OphydObj.subscriptions`:
* Run once per object instantiation, this can add up to a significant amount of time
* Instead, set once when subclassed

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -1,11 +1,10 @@
 import functools
 import time
 import weakref
-
 from enum import IntFlag
 from itertools import count
 from logging import LoggerAdapter, getLogger
-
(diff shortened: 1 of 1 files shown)


ZLLentz (approved):
This seems reasonable to me. It's good to trim off these per-device-instance bits as we identify them because some sessions can have many many devices.

tacaswell:
:+1:

My only concern was the posibility of spooky action-at-a-distance effects due to the set not be effectively global, but the `frozenset` takes fare of that.

tacaswell:
I guess we may also catch anyone who was monkeypatching subscription names on in `__init__`....but they probably should not be doing that anyway....

PR #1053, opened 2022-06-15
ENH: Component[T] generic support for type hinting

klauer:
## Proposal

Make `Component` a generic user-defined type `Component[T]`, with the single type parameter bound to `OphydObject`.

## Benefits

* Backward-compatible: no code changes required to existing codebases.
* Allows for static analysis of ophyd Device classes and instantiated  devices.
* Enables analyzers such as pyright/pylance to check signal attributes  and methods.
* Proposed API mirrors dataclass ``field()`` returning> ``Field`` with  ``component(cls)`` returning> ``Component[cls]``.

## Migration

Migration as simple as changing ``from ophyd import Component`` to  ``from ophyd import component as Component``, or alternatively changing individual `Component` instantiations to more specific `Component[EpicsSignal]` (for example).

## Screenshots

WIth something simple like:
```python
    det1 = AdscDetector("", name="abc")
    det1.cam.gain.setpoint_pvname
```

pyright is now able to determine types and go to the source of the components, even if nested:
image

image


ophyd/device.py
```diff
@@ -207,22 +257,33 @@ def __repr__(self):
...
     __str__ = __repr__
 
-    def __get__(self, instance, owner):
+    @typing.overload
+    def __get__(self, instance: None, owner: type) -> Component:
+        ...
+
+    @typing.overload
+    def __get__(self, instance: Any, owner: type) -> K:
+        ...
```
ZLLentz, line 274: Is this the core bit that makes the static typing analysis work? I feel like I need to read up more on the internals of how type hinting generics work to give a comprehensive review. I'm going to poke around with with branch locally to observe the effects.

klauer: This `__get__` block is what allows the [descriptor](https://docs.python.org/3/howto/descriptor.html) mechanism utilized by `Component` to describe what it does to the static analyzer: 
* Overloading indicates the two scenarios (1) access by class where you get back a `Component` instance itself, and (2) access by instance where you get back the instantiated `cpt.cls`.
* Both the generic `Component[K]` and this section are required for the instance `__get__` to work; otherwise the typevar `K` couldn't be used here.

ZLLentz:
Interactively, this works as advertised. In VSCode I'm getting helpful hints that show what types the components will be and all of their associated methods/members, as opposed to previously where components were marked as "Component | Any" without any helpful method/member hints. I'm going to take some time to understand the pieces of the diff here for my own learning benefit.

ophyd/areadetector/base.py
```diff
@@ -134,9 +137,35 @@ def _array_shape_callback(self, **kwargs):
                        **self._metadata)
 
 
-class ADComponent(Component):
-    def __init__(self, cls, suffix=None, *, lazy=True, **kwargs):
-        super().__init__(cls, suffix=suffix, lazy=lazy, **kwargs)
+K = TypeVar("K", bound=OphydObject)
+
+
+class ADComponent(Component[K]):
```
ZLLentz, line 143: Is this representative of what downstream libraries with custom `Component` subclasses need to do to propagate the type checking niceties?

ophyd/device.py
```diff
@@ -207,22 +257,33 @@ def __repr__(self):
         suffix = (repr(self.suffix) if self.suffix
                   else '')
 
-        component_class = self.cls.__name__
+        component_class = self.cls.__name__ if self.cls else ""
```
ZLLentz, line 260: What situation is being handled with this change here? Someone passing in a weird value for cls in `__init__`?

ophyd/device.py
```diff
@@ -207,22 +257,33 @@ def __repr__(self):
...
+    ) -> Union[Component, K]:
         if instance is None:
             return self
 
-        try:
-            return instance._signals[self.attr]
-        except KeyError:
+        if self.attr not in instance._signals:
             return instance._instantiate_component(self.attr)
+        return instance._signals[self.attr]
```
ZLLentz, line 286: Is there a deeper motivation behind this inversion here? (try/except -> attr in). To me it seems purely stylistic.

ophyd/device.py
```diff
@@ -264,8 +325,87 @@ def sub_value(self, func):
...
+            lazy=lazy,
+            trigger_value=trigger_value,
+            add_prefix=add_prefix,
+            doc=doc,
+            kind=kind,
+            **kwargs
+        )
+
+
+class FormattedComponent(Component[K]):
```
ZLLentz, line 408: Why does `ADComponent` have duplicated annotations from `Component`, while `FormattedComponent does not?

ophyd/device.py
```diff
@@ -303,7 +443,11 @@ def maybe_add_prefix(self, instance, kw, suffix):
         return suffix.format(**format_dict)
 
 
-class DynamicDeviceComponent(Component):
+component = Component.create
+formatted_component = FormattedComponent.create
```
ZLLentz, line 447: Are these the best names for these? It's ok for them to be slightly more verbose, I'm a bit worried about confusion resulting from  what amounts to just a case shift.

`typed_component`, maybe?

ZLLentz (commented):
I added some questions/thoughts after trying it out, reading through the diff, and leaning on the typing generics docs a bit.

I am overall :+1: on this. I have been bothered by the static analysis issues for a while now as we've moved to add type annotations to other libraries and I've been relying on them more.

ophyd/device.py
klauer: I had a branch where I allowed `cls` to be derived from the annotation alone, e.g., `Cpt[EpicsSignal]()` would be valid. I thought that would be deemed too magical so I removed it from the proposal.

This is an unintended change left in from the above; I'll remove it.

ophyd/device.py
klauer: I'll revert it for the sake of an easier-to-review PR. It doesn't need to be here.

ophyd/device.py
klauer: Solely for sphinx (note the failed docs build prior to [aaa7183](https://github.com/bluesky/ophyd/pull/1053/commits/aaa718335f5f8ac67d00d806527869790d654d9c)); I believe it to be a bug and this was the quickest way forward.

ophyd/device.py
klauer: I like the simplicity of the naming here, but I understand your hesitation. I'm open to input if anyone feels strongly on names.

ophyd/device.py
klauer: Resolved in 1acca934a43fc8ca5577a421e9d082cb70d10966

ophyd/device.py
klauer: Resolved in 1acca934a43fc8ca5577a421e9d082cb70d10966

ZLLentz:
One oddity that it took my brain a few hours to realize: I don't actually need to use the new `component` import path to get good suggestions in VSCode. It seems to work out of the box as-is with the previous imports. I'm not sure if VSCode is doing something extra here or if downstream device definitions can get this for free.

ophyd/areadetector/base.py
klauer: Yes, my understanding is that this is required for subclasses of `Component`

ophyd/areadetector/base.py
```diff
@@ -134,9 +137,35 @@ def _array_shape_callback(self, **kwargs):
...
+    #: Default laziness for the component.  All AreaDetector components are
+    #: by default lazy - as they contain thousands of PVs that aren't
+    #: necesssary for everyday functionality.
+    lazy_default: ClassVar[bool] = True
+
+    #: The attribute name of the component.
```
tacaswell, line 149: There are not inherited from the parent?

klauer:
@ZLLentz and I found that this could be drastically simplified. 

Apologies if you've already taken a look at the PR - but please reload as the contents of have been changed significantly with the last few commits.

ophyd/areadetector/base.py
klauer: This is related to https://github.com/bluesky/ophyd/pull/1053#discussion_r899422182 - would be happy to try any other suggestions you might have.

ophyd/device.py
```diff
@@ -119,7 +158,18 @@ def __init__(self, cls, suffix=None, *, lazy=False, trigger_value=None,
         self.add_prefix = tuple(add_prefix)
         self._subscriptions = collections.defaultdict(list)
 
-    def __set_name__(self, owner, attr_name):
+    def get_class_from_annotation(self) -> Optional[Type[K]]:
```
tacaswell, line 161: Is this only for testing?

ophyd/device.py
```diff
@@ -215,7 +265,19 @@ def __repr__(self):
...
+    def __get__(self, instance: None, owner: type) -> Component[K]:
+        ...
+
+    @typing.overload
+    def __get__(self, instance: Any, owner: type) -> K:
+        ...
+
+    def __get__(
+        self,
+        instance: Optional[Any],
```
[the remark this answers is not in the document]
tacaswell: "yes, but later" is a fine with me answer.

PR #1060, opened 2022-08-03
Adopt black

coretl:
Adopt black for code formatting, and use pre-commit to run black and flake8 on checkin and in CI. Blacken the whole codebase in a single commit. Put that commit in .git-blame-ignore-revs so it can be [ignored](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)


coretl:
@danielballan @tacaswell @mrakitin @klauer @ZLLentz @prjemian @dylanmcreynolds @whs92 @coretl @callumforrester please could someone review. The failing CI is fixed in #1058

tacaswell:
Rather than decorating every import with the `#noqa` can we skip those on the `__init__` files at the configuration level?  Matplotlib has https://github.com/matplotlib/matplotlib/blob/ffc55ae5c3b62b36fd994800404fbad98cb9e853/.flake8#L43-L133 in our config

coretl:
> Rather than decorating every import with the `#noqa` can we skip those on the `__init__` files at the configuration level? 

We can do, but I don't think I added any more `#noqa`s, I just moved them to different lines. I'd prefer to leave v1 as it is, and namespace v2 more heavily so we don't need as many `#noqa`s...

tacaswell (approved):
I have resisted this before, however my position has moved from "against" to "neutral".  I think we do want this for the ophyd v2 work (it is basically a new project), we want to do that work in the same tree (because it is not actually a new project and we are going to make sure they inter-operate), and formatting the whole code base is going to be way less hassle than trying to only format _some_ files.

One of my big concerns with blanket formatting is that this is in a very real sense un-reviewable (+10k/-8k LoC) and I have seen formatting tools put bugs in.  However black is widely enough used and checks the consistency at the AST level so I am confident that the reformatting will be correct.

PR #1064, opened 2022-08-05
Adopt isort

coretl:
Adopt isort for import ordering, run as a plugin to flake8 using flake8-isort. Adopt in a single commit as per black. Put that commit in .git-blame-ignore-revs so it can be [ignored](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -4,3 +4,5 @@
 #
 # Run black==22.3.0 on the codebase
 20e1b14b541a0331109158c77bc319248be28989
+# Use isort to sort imports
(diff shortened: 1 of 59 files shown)


coretl:

coretl:
Note the changes I had to make in [6e503c9](https://github.com/bluesky/ophyd/pull/1064/commits/6e503c9478bf36474e19589177629c5d3cec379c) as a result of `ophyd/__init__.py` imports being re-ordered (otherwise you get a circular import error). If these are too invasive I can back out and tell isort to ignore imports in `ophyd/__init__.py`

tacaswell (approved):

PR #1065, opened 2022-08-19
Move settings to setup.cfg and use setuptools-scm

coretl:
Move config from `.flake8`, `pytest.ini`, `setup.py`, `coveragerc`, and `requirements*.txt` to `setup.cfg` and adopt setuptools-scm instead of versioneer (which will also allow us to delete `MANIFEST.in`).

Rewrite CI to share steps in less workflows and test the packaging.


setup.cfg
```diff
@@ -1,9 +1,124 @@
...
+# Specify any package data to be included in the wheel below.
+[options.package_data]
+ophyd =
+    # Include our documentation helpers:
+    "*.rst"
+
+[options.entry_points]
+# Include a command line script
+console_scripts =
+    python3-pip-skeleton = python3_pip_skeleton.__main__:main
```
ZLLentz, line 72: I'm not super familiar with these files, what exactly is this doing? Naively it looks to me like ophyd is providing a console script called `python3-pip-skeleton` defined in a different library?

ZLLentz (commented):
I did my best to review the parts I understood. I also tried to double-check that all the elements from the previous setup scripts found a new home and I didn't notice any discrepancies that made me worry.

I don't have any experience with the github actions configuration and I'm much more familiar with the older setup scripts than the setuptools-scm ways, so hopefully others can review too.

I left a few questions that might be more for my understanding than anything else.

Also: the inline codecov is a bit obnoxious here in the review process. I suspect it's more useful if your repo has high coverage? I hope there's some way to hide these...

ophyd/__init__.py
```diff
@@ -127,5 +125,13 @@ def get_cl():
...
-del get_versions
+try:
+    # Use live version from git
+    from setuptools_scm import get_version
+
+    # Warning: If the install is nested to the same depth, this will always succeed
+    __version__ = get_version(root="..", relative_to=__file__)
+    del get_version
+except (ImportError, LookupError):
+    # Use installed version
```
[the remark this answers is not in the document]
coretl: The setuptools_scm hook adds a `_version.py` file on install, which we don't add to git. This means that from a clean checkout, running from source, it won't exist. However, I have moved all dev dependencies into extra_requires, and we do an editable install to pull these in, so the `_version.py` file will be created. If we subsequently make some commits, but keep on running from this editable install, then the `_version.py` file will be wrong, so we use the version from git in preference. This is a bit of an edge case, but one that we have come across locally (running something from a random editable install somewhere, how do we tell what version is actually running)

ophyd/__main__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from . import __version__
+
+__all__ = ["main"]
+
+
+def main(args=None):
+    parser = ArgumentParser()
+    parser.add_argument("--version", action="version", version=__version__)
+    args = parser.parse_args(args)
```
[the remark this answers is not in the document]
coretl: This is here purely so someone can type `python -m ophyd --version` and get the installed version of ophyd. It is used mainly in CI to test that ophyd imports and gives a version number. It is here because most of my projects sprout a CLI at some point, but I don't know if ophyd ever will, so maybe it is out of place here. If this is too intrusive, I could replace the invocation in CI with `python -c "import ophyd; print(ophyd.__version__)"` to have the same effect.

setup.cfg
coretl: Oops, that's a copy-paste fail, there's not meant to be a console_script for ophyd. Deleted in 9f0847e

coretl:
> Also: the inline codecov is a bit obnoxious here in the review process. I suspect it's more useful if your repo has high coverage? I hope there's some way to hide these...

I was surprised to see that there was already a `.codecov` file for this repo, and all I did was correct the syntax of it, and codecov started running its checks. I tend to add it to my projects as I keep coverage close to 100%, but it may be less useful here. At least it is only set to run as a check here, rather than eagerly commenting all the time... When you refer to "hiding these", do you mean the inline comments that it makes on other repos, or something that's happening to the ophyd repo?

ophyd/__init__.py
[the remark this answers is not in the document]
ZLLentz: This makes sense to me

ophyd/__main__.py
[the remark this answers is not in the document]
ZLLentz: After 1 weekend of brain churning I think its reasonable, I was mostly confused because I hadn't seen an entrypoint sprout up just for versions before. In retrospect with more time to consider it I'm thinking that some of my projects would benefit too, particularly for quickly checking if I'm running some sort of dev version.

setup.cfg
```diff
@@ -1,9 +1,121 @@
-[versioneer]
-VCS = git
-style = pep440-post
-versionfile_source = ophyd/_version.py
-versionfile_build = ophyd/_version.py
-tag_prefix = v
+[metadata]
```
prjemian, line 1: Can any of this be moved to and used from `pyproject.toml`?  There is a [recommendation](https://pypi.org/project/setuptools-scm/) to deprecate `setup.cfg`:

> usage in setup.cfg is considered deprecated, please use pyproject.toml whenever possible

coretl: I think most things can read from `pyproject.toml`. [tox](https://tox.wiki/en/latest/example/basic.html#pyproject-toml-tox-legacy-ini) just embeds the ini under a specific key, which is a bit ugly, but flake8 [won't add support for pyproject.toml](https://github.com/PyCQA/flake8/issues/234#issuecomment-1206730688) until pip changes its behaviour on the presence of pyproject.toml. I'll try moving some things and see if it works

coretl: Actually, as the `[tool.setuptools]` table in pyproject.toml [is still in beta](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#beta) I think it's best to leave things in setup.cfg until they have stabilized.

coretl: > There is a [recommendation](https://pypi.org/project/setuptools-scm/) to deprecate `setup.cfg`:

It looks like this recommendation is from setuptools-scm, and we already put its configuration in pyproject.toml

PR #1066, opened 2022-08-24
Fix capture signal was always triggered

physwkim:
* Capture signal is always triggered even in Single write mode.
* In single mode, the following error message is displayed
```ERROR: capture not supported in Single mode```


tacaswell (approved):

prjemian:
@tacaswell Can we get a new patch release ASAP that includes this fix?

PR #1067, opened 2022-09-02
Add v2 dir

coretl:
The v1/v2 import compatibility turned out to be a bit of a rabbithole. 
`__getattr__` in `ophyd/__init__.py` seems only to work for `import ophyd`. Things like `import ophyd.utils` fail with `ModuleNotFoundError`. I did try sys.modules["ophyd"] = sys.modules["ophyd.v1"] but that meant ophyd.v1.Component != ophyd.Component and many other subtle bugs. I considered manufacturing modules programmatically, but then had another idea:

- At the moment we `import ophyd` for v1. Keep this for 1.x releases, but allow early adopters to `import ophyd.v2`.
- When v2 stabilizes, we make a 2.x release that moves the whole of v1 into a subpackage and requires `import ophyd.v1` for legacy code. Structure v2 to have minimal public subpackages so it is easy to make a handful of top level subpackages that just `from ophyd.v2.subpackage import *`

Any thoughts on this approach?


ZLLentz:
I think your approach is good and practical. I'm somewhat surprised that everything fell over when you tried to stuff it into a v1 directory. I don't see much of a reason to do anything other than :+1: this pr since it simply adds a bare submodule.

ZLLentz (approved):

coretl:
> I'm somewhat surprised that everything fell over when you tried to stuff it into a v1 directory

This surprised me too. I thought a module level `__getattr__` would be sufficient, but apparently not. I thought about making a `sys.modules` tree that mirrors that of v1, but that would confuse IDEs. The only thing I could think of that would work for both IDEs and runtime is making a parallel tree of `.py` files with `from ophyd.v1.pkg.subpkg import *`. Avoiding all of this by making v2 (which will be much smaller) importable as `import ophyd.v2` on a 1.x release, then `import ophyd` or `import ophyd.v2` on a 2.x release (with a parallel tree of `.py` files) seems an easier thing to do.

> I don't see much of a reason to do anything other than +1 this pr since it simply adds a bare submodule.

Yes, This is more about transition strategy for the future rather than any work for now, and so I have a way to find it again when I need it!

PR #1068, opened 2022-09-05
Fix SynAxisNoPosition

coretl:
PR #1036 made SynAxis produce a MoveStatus. Unfortunately SynAxisNoPosition inherits from this, but has no position, which is required for MoveStatus. This reverts to using DeviceStatus just for SynAxisNoPosition and adds a test for it

This was spotted in https://github.com/bluesky/bluesky/pull/1542

Credit to @callumforrester for the test

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -470,6 +470,9 @@ class SynAxis(Device):
         self._events_per_move = events_per_move
         self.egu = egu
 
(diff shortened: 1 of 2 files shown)


tacaswell (approved):

tacaswell:
Merging to unbreak bluesky tests.

PR #1073, opened 2022-09-13
Signal settle time

maffettone:
Currently settle time only exists as an attribute in Positioners and not Signals. However, `Signal.set()` uses this as a keyword argument, which obscures it from `bps.mv` or other plans. This adds `settle_time` as a property of Signal, that can be overridden in the call to `set()`. It keeps the defaults as None, and should maintain backward compatibility with existing devices.

### DIFF
--- a/ophyd/serial.py
+++ b/ophyd/serial.py
@@ -1,0 +1,92 @@
+import threading
+import time
+from ophyd import Kind
+from ophyd.ophydobj import OphydObject
+import serial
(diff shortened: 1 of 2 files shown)


maffettone:
This was built off an outdated version of ophyd which inhibits a clean merge. Discarding and revisiting.

tacaswell:
you rebase and force-push as well.

PR #1076, opened 2022-09-23
Use pydata docs theme and make a space for v2 docs

coretl:


coretl:
Here's the output, unzip and point a web browser at index.html
[html.zip](https://github.com/bluesky/ophyd/files/9633120/html.zip)

docs/conf.py
```diff
@@ -0,0 +1,115 @@
...
+version = ophyd.__version__
+
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = "sphinx"
+
+# -- Options for HTML output ---------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+html_theme = "pydata_sphinx_theme"
```
mrakitin, line 61: Are we sure we don't need a package from https://pypi.org/project/pydata-sphinx-theme/?

mrakitin (approved):
Nice addition! The built docs looks pretty 🙂. I have one comment below, but if it builds with the standard sphinx, it's good, just wanted to confirm.

To check the built docs from CI, I recently added [this step](https://github.com/NSLS-II/sirepo-bluesky/blob/38cd3515ecac50cb2772fa57f2b2c0b07d270a77/.github/workflows/docs.yml#L88-L91) into another project, which can be helpful for reviews without manual uploads of the archives. The artifacts then can be found in the Actions Summary for the run ([example](https://github.com/NSLS-II/sirepo-bluesky/actions/runs/3101472652)).

setup.cfg
```diff
@@ -48,16 +48,16 @@ dev =
     myst-parser
     numpydoc
     pre-commit
+    pydata-sphinx-theme
```
coretl, line 51: @mrakitin here's the theme dependency, it moved from a `requirements.txt` file into a `[dev]` extra in a previous PR

coretl:
@mrakitin I've added in that artifact step, so the built docs get uploaded to https://github.com/bluesky/ophyd/suites/8470735727/artifacts/376304136

PR #1081, opened 2022-12-07
DOC: added a shutter example of a PVPositionerComparator

clintonroy:


ZLLentz (approved):
This is a good addition to the docs. The CI failure is unrelated and looks like a race condition to me.

PR #1085, opened 2022-12-16
Fix docs linking

coretl:
It's more fussy since v2 changes merged in...


coretl:
codecov coverage difference looks spurious...

tacaswell (approved):

PR #996, opened 2021-06-23
AD: add "create_directory" component to stage_sigs

mrakitin:
This PR adds the `create_directory` component to the `stage_sigs` of the `FileStorePluginBase` class, so that the requested directory can be created by the AreaDetector. See https://github.com/NSLS-II-SIX/profile_collection/pull/36 for the original implementation, ported to upstream here.

PR #1087, opened 2023-01-10
Change `make_dir_tree` default dir mode to `755`

mrakitin:
We had an internal discussion in NSLS-II Slack, and thought it would be more useful/secure to have the default mode of the created directory tree without writing permissions for "others". It's questionable what we should leave for the "group" writing permissions, for now, I set it to `755` (=`rwxr-x-rx`), but open for discussion/arguments.

### DIFF
--- a/ophyd/utils/paths.py
+++ b/ophyd/utils/paths.py
@@ -2,7 +2,7 @@ import calendar
 import os
 
 
-def makedirs(path, *, mode=0o777, mode_base=None):
+def makedirs(path, *, mode=0o755, mode_base=None):
     """Recursively make directories and set permissions
 
     Parameters
@@ -36,7 +36,7 @@ def makedirs(path, *, mode=0o777, mode_base=None):
     return ret
 
 
-def make_dir_tree(year, *, base_path=None, mode=0o777):
+def make_dir_tree(year, *, base_path=None, mode=0o755):
     """Make full directory tree for the year
 
     Parameters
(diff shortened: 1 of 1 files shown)


cryos (approved):
This is a much more reasonable default in my opinion. Thanks!

prjemian:
I f there is any controversy over 775 vs 755, make it a module attribute that can be changed locally.

ZLLentz:
Just curious: what are these used for? I don't see any calls to them in `ophyd` outside of test contexts, so it's not immediately obvious to me how it's being used in the context of a hardware abstraction library. That might inform if/how to parameterize it.

My gut feeling is to leave it as is in this PR or defer to user's `umask` but I don't know the context here.

tacaswell:
This dates from before AD would auto-make directories for you so we had to create them by hand and the default format in the file store plugins in `/base/year/month/day`.

mrakitin:
Thanks all for the reviews!

Right, that creates the directory structure explained by Tom. An example directory path can be found in the screenshot at https://github.com/bluesky/ophyd/pull/996#issuecomment-867025902. The dir structure can be created automatically by AD with the "Create dir. depth" field (or the corresponding PV), but some detectors do not support that, so we pre-create the directory for each year/month/date ahead of time.

Another example of where we use it (both the structure creation and filling it with the detector data):
- [creation](https://github.com/NSLS-II/sirepo-bluesky/blob/da7a04f938341bb7abe90e230b718787677a1db6/examples/prepare_det_env.py#L35-L36)
- [filling](https://github.com/NSLS-II/sirepo-bluesky/blob/da7a04f938341bb7abe90e230b718787677a1db6/sirepo_bluesky/sirepo_ophyd.py#L110-L120)

PR #1088, opened 2023-01-23
Add status repr to WaitTimeoutError message

dperl-dls:
In a similar situation to #1086 , we had an issue where the error message `Status has failed to complete` was not very illuminating as to which status that was, so I have added this information to the `raise WaitTimeoutError` calls in the same manner as it is added to other exception messages in `status.py`. This would be really helpful to us in avoiding wrapping every `wait()` with try/except to record what we are waiting on or subclassing every kind of status we have to modify `wait()` and I hope it might be useful to others too.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -399,7 +399,7 @@ class StatusBase:
             when this method was called, not from the beginning of the action).
         """
         if not self._event.wait(timeout=timeout):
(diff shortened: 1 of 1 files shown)


tacaswell (approved):

tacaswell:
Thank you @dperl-dls !

PR #1091, opened 2023-02-13
Added conditional to check kwargs of EpicsSignalRW

RAYemelyanova:
I thought it would be useful to have EpicsSignalRW set the write_pv to read_pv by default, if no write_pv is specified.


coretl:
Apologies, I can't remember my own code. It appears this already works:
https://github.com/bluesky/ophyd/blob/5a0f1f3e8c2f453eb5a3262c610714bee67f84db/ophyd/v2/epicsdemo/__init__.py#L30
Because the connection code checks those variables:
https://github.com/bluesky/ophyd/blob/5a0f1f3e8c2f453eb5a3262c610714bee67f84db/ophyd/v2/epics.py#L211-L226

PR #1093, opened 2023-02-17
BUG: Let channelsim work with enums

tizayi:
Channel sim can now initialise simulated values for python Enum data types.

### DIFF
--- a/ophyd/v2/_channelsim.py
+++ b/ophyd/v2/_channelsim.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import time
+from enum import Enum
 from typing import Dict, Generic, List, Sequence, Type, TypeVar
 
 from bluesky.protocols import Descriptor, Dtype, Reading
@@ -23,9 +24,16 @@ def make_sim_descriptor(source: str, value) -> Descriptor:
         dtype = primitive_dtypes[type(value)]
         shape = []
     except KeyError:
-        assert isinstance(value, Sequence), f"Can't get dtype for {type(value)}"
-        dtype = "array"
-        shape = [len(value)]
+        if isinstance(value, Sequence):
+            dtype = "array"
+            shape = [len(value)]
+        elif isinstance(value, Enum):
+            dtype = "string"
+            shape = []
+            choices = [e.value for e in type(value)]
+            return dict(source=source, dtype=dtype, shape=shape, choices=choices)
+        else:
+            assert False, f"Can't get dtype for {type(value)}"
     return dict(source=source, dtype=dtype, shape=shape)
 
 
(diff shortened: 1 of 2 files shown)


coretl:
Please can we have a test for this? I suggest something like this:
https://github.com/bluesky/ophyd/blob/5a0f1f3e8c2f453eb5a3262c610714bee67f84db/ophyd/v2/tests/test_epicsdemo.py#L86-L101
But for `epicsdemo.Sensor` that already has an Enum type defined for it, but has never been created in tests with `sim=True`

ophyd/v2/_channelsim.py
```diff
@@ -23,9 +24,14 @@ def make_sim_descriptor(source: str, value) -> Descriptor:
...
     except KeyError:
-        assert isinstance(value, Sequence), f"Can't get dtype for {type(value)}"
-        dtype = "array"
-        shape = [len(value)]
+        if isinstance(value, Sequence):
+            dtype = "array"
+            shape = [len(value)]
+        elif issubclass(type(value), Enum):
+            dtype = "enum"
+            shape = []
```
coretl, line 32: @danielballan @tacaswell event model currently has no concept of an Enum. Should we represent as a string, or extend event model? I don't know if downstream should care if it is an Enum as event docs are read-only, so it will never have to set it. The `get_value` and `set` methods in ophyd support Enum types

coretl: Decided on `dtype=string` but add `choices=[""one","two"]` into descriptor

ophyd/v2/_channelsim.py
```diff
@@ -59,7 +67,11 @@ def __init__(self, pv: str, datatype: Type[T]):
         self.put_proceeds = asyncio.Event()
         self.put_proceeds.set()
         self._listeners: List[SimMonitor[T]] = []
-        self.set_value(datatype())
+
+        if issubclass(datatype, Enum):
+            self.set_value(list(datatype)[0])
```
coretl, line 72: Didn't know you could list enums, neat!

ophyd/v2/_channelsim.py
```diff
@@ -23,9 +24,16 @@ def make_sim_descriptor(source: str, value) -> Descriptor:
...
-        assert isinstance(value, Sequence), f"Can't get dtype for {type(value)}"
-        dtype = "array"
-        shape = [len(value)]
+        if isinstance(value, Sequence):
+            dtype = "array"
+            shape = [len(value)]
+        elif isinstance(value, Enum):
+            dtype = "string"
+            shape = []
+            choices = [e.value for e in type(value).__members__.values()]
```
coretl, line 33: As we can list enums this can be simplified then
```suggestion
            choices = [e.value for e in type(value)]
```

coretl (approved):

Issue #1098, opened 2023-03-01
Ophyd.v2: Work out how to propogate cause of exception into FailedStatus


coretl:
c.f. https://github.com/bluesky/ophyd-epics-devices/actions/runs/4295440949/jobs/7485860571

coretl:
Suggest that we add:
```python
@property
def exception(self) -> Exception: ...
```
to the bluesky/protocols.py, then in the code check if it is a method and call it with `timeout=0` for back-compat with ophyd.v1

Please add this as a PR to bluesky for general comment

PR #1103, opened 2023-03-16
Ophyd v2: add _exception to AsyncStatus to propagate errors through FailedStatus.

RAYemelyanova:
added ._exception to AsyncStatus to match with method in v1 Status object

Solves https://github.com/bluesky/ophyd/issues/1098

This will be followed by a PR in bluesky, which will change the Status object that ophyd v2 AsyncStatus inherits from as well as how FailedStatus gets instantiated. Once I have done this, I will link it here so the changes can be viewed together.

### DIFF
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -38,6 +38,7 @@ from bluesky.protocols import (
     Stageable,
     Status,
     Subscribable,
+    StatusException,
 )
 from bluesky.run_engine import call_in_bluesky_event_loop
 
@@ -60,6 +61,7 @@ class AsyncStatus(Status):
         self.task.add_done_callback(self._run_callbacks)
         self._callbacks = cast(List[Callback[Status]], [])
         self._watchers = watchers
+        self._exception = None
 
     def __await__(self):
         return self.task.__await__()
@@ -76,6 +78,10 @@ class AsyncStatus(Status):
                 callback(self)
 
     @property
+    def exception(self) -> Optional[StatusException]:
+        return self._exception
+
+    @property
     def done(self) -> bool:
         return self.task.done()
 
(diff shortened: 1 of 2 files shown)


tacaswell:
Per https://github.com/matplotlib/matplotlib/issues/25406 we should at have a think about keeping a reference to the exception.  I think in this case we (and the v1 status objects) we want to do this so we can move the exceptions around, but we should be sure.

ophyd/v2/core.py
```diff
@@ -75,6 +77,10 @@ def _run_callbacks(self, task: asyncio.Task):
             for callback in self._callbacks:
                 callback(self)
 
+    @property
+    def exception(self) -> Optional[StatusException]:
```
tacaswell, line 81: Note to self, this matches https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.exception

ophyd/v2/core.py
```diff
@@ -75,6 +77,10 @@ def _run_callbacks(self, task: asyncio.Task):
             for callback in self._callbacks:
                 callback(self)
 
+    @property
+    def exception(self) -> Optional[Union[Exception, asyncio.CancelledError]]:
+        return self._exception
```
coretl, line 82: I'd prefer to get this from the task rather than storing it...
```suggestion
        try:
            self.task.result()
        except (Exception, asyncio.CancelledError) as e:
            return e
        else:
            return None
```

RAYemelyanova: Sure, this makes sense. I'll add that in now.

RAYemelyanova: I've changed the success method, so that it doesnt store the exception in ._exception either. Presumably, the self.task attribute will contain that information in self.task.exception().

coretl: yes

RAYemelyanova:
At this point, I think I've tested every line I've added in the ophyd v2 core.py file... Should I write some more tests for anything else? Maybe for the v1 part?

coretl (approved):

Issue #1090, opened 2023-02-10
Ophyd.v2: Move set_name and connect functionality out of StandardReadable to a more generic class


RAYemelyanova:
### Overview
This is a bit of a philosophical issue. I'm currently working on writing some Ophyd v2 devices for Zebras, for a simple tomography scan. I have been using ophyd-epics-device's implementation of an area detector (MyHDFFlyerSim) for this aim. 

The StandardReadable class has a set_name method which conveniently prepends a prefix to all children of a Device. This is very handy, because any class that inherits from it can just define lots of PV's with names relative to the prefix given to it, and calling ` super().__init__(prefix, name)` in its constructor will automatically set the names of those PVs correctly.

For both my Zebra Flyable and the area detector in ophyd-epics-devices, the rest of the functionality of StandardReadable is unwanted and in fact un-needed. It therefore makes sense to extract the set_name and connect methods of a StandardReadable to a more generic base class.

### An Example

As mentioned, ophyd-epics-devices contains a flyable area detector definition, called `MyHDFFlyerSim`. It looks like so:

```
from bluesky.protocols import Flyable, WritesExternalAssets
from ophyd.v2.core import StandardReadable
...

class MyHDFFlyerSim(StandardReadable, Flyable, WritesExternalAssets):
    def __init__(self, prefix: str, name="") -> None:
        self.drv = ADDriver("CAM")
        self.hdf = NDFileHDF("HDF5:")
        self._asset_docs = collections.deque()  # type: ignore
        self._stage_task: Optional[asyncio.Task] = None
        self._capture_status: Optional[AsyncStatus] = None
        self._start_status: Optional[AsyncStatus] = None
        self._compose_datum: Optional[Callable] = None
        self.stream: str = "primary"
        super().__init__(prefix, name, config=[self.drv.acquire_time])

    async def _stage(self) -> None:
        ...

    ...
```
See here for the code: **https://github.com/bluesky/ophyd-epics-devices/blob/main/src/ophyd_epics_devices/areadetector.py#L236**

This class does not contain `read`, `describe`, `read_configuration` or `describe_configuration` methods which are defined in StandardReadable. As it is a Flyable, those methods are not intended to be used. The only reason it inherits StandardReadable is for the set_name and collect functionality. This seems inefficient and could probably be improved upon.

### Potential Solutions
[...]

PR #1104, opened 2023-04-05
v2: Add dict support

RAYemelyanova:
The purpose of this PR is to be able to define v2 Devices with dictionary attributes containing (nested) devices, instead of just plain devices. i.e. something like:

```python
class myDevice(StandardReadable):
    def __init__(self, prefix: str, name: str = ""):
        self.child1 = DummyDevice("device1")
        self.dict_with_children =
            {
                "abc": DummyDevice("device2"),
                123: DummyDevice("device3"),
            }
      
        super().__init__(prefix, name)
```

The idea is, that these sub-devices will all be correctly named and connected. With the current master branch, "device2" and "device3" would be ignored (i.e. the devices in the dictionary).

## Initial approach
Initially, I allowed a user to create a dictionary as an attribute to a Device, as shown in the code snippet above. This worked OK; standard readable has a set_name and connect method that finds all the children of the object, so I extracted this into its own generator function (instead of duplicating this login in both). I then added a checker in this function which would extract device children from inside dictionaries, like so:

```python
def get_device_children(device: Device) -> Generator[Tuple[str, Device], None, None]:
    for attr_name, attr in device.__dict__.items():
        if isinstance(attr, Device):
            yield f"{attr_name.rstrip('_')}", attr
        elif isinstance(attr, Dict):
            for dict_key, dict_value in attr.items():
                if isinstance(dict_value, Device):
                    yield f"{attr_name.rstrip('_')}-{dict_key}", dict_value
```

The outcome of this approach, along with tests, can be seen by checking out commit 04ff4934.

## Current approach
It was discussed with @coretl that perhaps a better approach would be to make the dictionary typed, or write a wrapper around a dictionary which would have set_name and connect methods. I've experimented with this in the latest commit.

There are some benefits to this approach. Most importantly, with a wrapped dictionary users know exactly what they need to put in; the signature of this class explicitly states the types it allows. I've decided to type this class as:

```python
KT = TypeVar("KT")

class DeviceDict(MutableMapping[KT, Device]):
    def __init__(self, dictionary=None ...):
        ...

    def set_name(self, parent_name: str):
        ...
    async def connect(self, prefix: str, sim: bool):
        for _, device in self.data.items():
            await device.connect(prefix, sim)
```
this feels like a more pythonic approach, as it extracts connecting/setting names of dict elements into a seperate class (DeviceDict) rather than relying on the get_device_children function to do this for us.

## Further discussion...

I don't know if how I've written it here is the best way. The tests should provide a good overview of how they would be used, and I don't know if people would find this verbose.

This also touches on [this issue](https://github.com/bluesky/ophyd/issues/1090), and I don't know if it just runs parallel with it, because `DeviceDict` is only used for the purpose of the StandardReadable properly naming things and connecting them. And it feels like it should be more generic.

### DIFF
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -1,6 +1,7 @@
 """Core Ophyd.v2 functionality like Device and Signal"""
 
 from __future__ import annotations
+import abc
 
 import asyncio
 import functools
@@ -19,7 +20,9 @@ from typing import (
     Generator,
     Generic,
     Iterable,
+    Iterator,
     List,
+    MutableMapping,
     Optional,
     Protocol,
     Sequence,
@@ -201,12 +204,8 @@ async def connect_children(device: Device, prefix: str, sim: bool):
 
 def get_device_children(device: Device) -> Generator[Tuple[str, Device], None, None]:
     for attr_name, attr in device.__dict__.items():
-        if isinstance(attr, Device):
+        if isinstance(attr, Device) or isinstance(attr, DeviceDict):
             yield f"{attr_name.rstrip('_')}", attr
-        elif isinstance(attr, Dict):
-            for dict_key, dict_value in attr.items():
-                if isinstance(dict_value, Device):
-                    yield f"{attr_name.rstrip('_')}-{dict_key}", dict_value
 
 
 class DeviceCollector:
@@ -618,3 +617,81 @@ class StandardReadable(Readable, Configurable, Stageable, Device):
 
     async def read_configuration(self) -> Dict[str, Reading]:
         return await merge_gathered_dicts(sig.read() for sig in self._conf_signals)
+
+
+KT = TypeVar("KT")
+
+class DeviceDict(MutableMapping[KT, Device]):
+    def __init__(self, dictionary=None, /, **kwargs) -> None:
+        self.data: Dict[KT, Device] = {}
+        if dictionary is not None:
+            self.update(dictionary)
+        if kwargs:
+            self.update(kwargs)
+    
+    def __contains__(self, key: KT) -> bool:
+        return key in self.data
+
+    def __delitem__(self, key: KT) -> None:
+        del self.data[key]
+
+    def __getitem__(self, key: KT) -> Device:
+        if key in self.data:
+            return self.data[key]
+        raise KeyError(key)
+
+    def __len__(self) -> int:
+        return len(self.data)
+
+    def __iter__(self) -> Iterator[KT]:
+        return iter(self.data)
+
+    def __setitem__(self, key: KT, value: Device) -> None:
+        self.data[key] = value
+    
+    @classmethod
+    def fromkeys(cls, iterable: Iterable[KT], value: Device) -> "DeviceDict":
+        """Create a new dictionary with keys from `iterable` and values set 
+        to `value`.
+
+        Args:
+            iterable: A collection of keys.
+            value: The default value. All of the values refer to just a single 
+                instance, so it generally does not make sense for `value` to be a 
+                mutable object such as an empty list. To get distinct values, use 
+                a dict comprehension instead.
+
+        Returns:
+            A new instance of MyDict.
+        """
+        d = cls()
+        for key in iterable:
+            d[key] = value
+        return d
+
+    def update(self, other=(), /, **kwds) -> None:
+        """Updates the dictionary from an iterable or mapping object."""
+        if isinstance(other, MutableMapping):
+            for key in other:
+                self.data[key] = other[key]
+        elif hasattr(other, "keys"):
+            for key in other.keys():
+                self.data[key] = other[key]
+        else:
+            for key, value in other:
+                self.data[key] = value
+        for key, value in kwds.items():
+            self.data[key] = value
+
+    def set_name(self, parent_name: str):
+        for name, device in self.data.items():
+            device.set_name(f"{parent_name}-{name}")
+            device.parent = self
+
+    async def connect(self, prefix: str, sim: bool):
+        for _, device in self.data.items():
+            await device.connect(prefix, sim)
+        
+
+    
+    
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -19,9 +19,7 @@ from typing import (
     Generator,
     Generic,
     Iterable,
-    Iterator,
     List,
-    MutableMapping,
     Optional,
     Protocol,
     Sequence,
@@ -193,17 +191,19 @@ async def connect_children(device: Device, prefix: str, sim: bool):
             await connect_children(self, prefix + self.prefix, sim)
     """
 
-    coros = {
-        name: child_device.connect(prefix, sim)
-        for name, child_device in get_device_children(device)
-    }
+    coros = {}
+    for name, child_device in get_device_children(device):
+        if isinstance(child_device, DeviceDict):
+            coros.update(child_device.nested_connect(prefix, sim))
+        else:
+            coros.update({name: child_device.connect(prefix, sim)})
 
     await wait_for_connection(**coros)
 
 
 def get_device_children(device: Device) -> Generator[Tuple[str, Device], None, None]:
     for attr_name, attr in device.__dict__.items():
-        if isinstance(attr, Device) or isinstance(attr, DeviceDict):
+        if isinstance(attr, Device):
             yield f"{attr_name.rstrip('_')}", attr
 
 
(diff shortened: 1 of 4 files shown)


ophyd/v2/core.py
```diff
@@ -607,3 +616,78 @@ async def describe_configuration(self) -> Dict[str, Descriptor]:
 
     async def read_configuration(self) -> Dict[str, Reading]:
         return await merge_gathered_dicts(sig.read() for sig in self._conf_signals)
+
+
+KT = TypeVar("KT")
+
+
+class DeviceDict(MutableMapping[KT, Device]):
```
coretl, line 624: Could this be written smaller if it subclassed `dict` and added a few methods? Also, if you subclass `Device` you don't need to specialcase the isinstance check in `get_device_children`:

```python
class DeviceDict(Dict[KT, Device], Device):
    def set_name(self, name: str):
        for child_name, device in self.items():
            device.set_name(f"{name}-{child_name}")
            device.parent = self

    async def connect(self, prefix: str, sim: bool):
        coros = {
            k: d.connect(sim)
            for k, d in self.items()
        }
        await wait_for_connection(**coros)
```

RAYemelyanova: interesting you wrote the connect method in this way; in the latest commit, the connect method just loops through all the keys in the dictionary and connects them in series. This is obviously not practical, but I found that if I used the method you've written above in the past this led to things not being connected properly.

I'm really not sure why exactly it's failing. But if you put that bit of code in and try to run the bottom test of test_core in ophyd/v2/tests it fails because the dict devices actually don't have their connect() methods run at all.

RAYemelyanova: note: have just checked and with the DeviceDict as above (except with `d.connect(prefix, sim)` instead) that test only works for this case:

```python
async def test_children_of_standard_readable_have_set_names_and_get_connected():
    parent = DummyStandardReadable("parent")
    parent.set_name("parent")
    assert parent.child1.name == "parent-child1"
    assert parent.dict_with_children[123].name == "parent-dict_with_children-123"
    assert parent.dict_with_children["abc"].name == "parent-dict_with_children-abc"

    await parent.connect()

    await parent.dict_with_children.connect(parent._init_prefix, False) #have to connect this otherwise test fails.

    assert parent.child1.connected
    assert parent.dict_with_children[123].connected
    assert parent.dict_with_children["abc"].connected

```

RAYemelyanova: In the latest commit I have proposed a solution to this problem... but I don't think it's pretty.

ophyd/v2/core.py
```diff
@@ -188,14 +190,23 @@ async def connect_children(device: Device, prefix: str, sim: bool):
...
-        for k, c in device.__dict__.items()
-        if k != "parent" and isinstance(c, Device)
-    }
+
+    coros = {}
+    for name, child_device in get_device_children(device):
+        if isinstance(child_device, DeviceDict):
+            coros.update(child_device.nested_connect(prefix, sim))
+        else:
+            coros.update({name: child_device.connect(prefix, sim)})
```
coretl, line 199: We call this when implementing the `connect` method of a Device, so this will never be called with a `DeviceDict`, so we can go back to the simpler implementation:

```suggestion
    coros = {
        name: child_device.connect(prefix, sim) 
        for name, child_device in get_device_children(device)
    }
```

ophyd/v2/core.py
```diff
@@ -188,14 +190,23 @@ async def connect_children(device: Device, prefix: str, sim: bool):
...
+
     await wait_for_connection(**coros)
 
 
+def get_device_children(device: Device) -> Generator[Tuple[str, Device], None, None]:
+    for attr_name, attr in device.__dict__.items():
+        if isinstance(attr, Device):
+            yield f"{attr_name.rstrip('_')}", attr
```
coretl, line 207: I would do the `parent` check here, but not the attribute naming change
```suggestion
        if attr_name != "parent" and isinstance(attr, Device):
            yield attr_name, attr
```

ophyd/v2/core.py
```diff
@@ -570,11 +581,9 @@ def name(self) -> str:
     def set_name(self, name: str = ""):
         if name and not self._name:
             self._name = name
-            for attr_name, attr in self.__dict__.items():
-                # TODO: support lists and dicts of devices
-                if isinstance(attr, Device):
-                    attr.set_name(f"{name}-{attr_name.rstrip('_')}")
-                    attr.parent = self
+            for child_name, child in get_device_children(self):
+                child.set_name(f"{name}-{child_name}")
```
coretl, line 585: This is where I'd restore the naming
```suggestion
                child.set_name(f"{name}-{child_name.rstrip('_')}")
```

ophyd/v2/core.py
```diff
@@ -607,3 +616,21 @@ async def describe_configuration(self) -> Dict[str, Descriptor]:
...
+            device.set_name(f"{parent_name}-{name}")
+            device.parent = self
+
+    async def connect(self, prefix: str, sim: bool):
+        coros = {str(k): d.connect(prefix, sim) for k, d in self.items()}
+        await wait_for_connection(**coros)
+
+    def nested_connect(self, prefix: str, sim: bool):
+        coros = {str(k): d.connect(prefix, sim) for k, d in self.items()}
+        return coros
```
coretl, line 636: With the above changes we don't need this

ophyd/v2/core.py
```diff
@@ -607,3 +616,21 @@ async def describe_configuration(self) -> Dict[str, Descriptor]:
...
     async def read_configuration(self) -> Dict[str, Reading]:
         return await merge_gathered_dicts(sig.read() for sig in self._conf_signals)
+
+
+KT = TypeVar("KT")
+
+
+class DeviceDict(Dict[KT, Device], Device):
+    def set_name(self, parent_name: str):
+        for name, device in self.items():
```
coretl, line 626: Also need to set `self._name` and implement a `def name(self) -> str:` property

ophyd/v2/core.py
```diff
@@ -607,3 +616,21 @@ async def describe_configuration(self) -> Dict[str, Descriptor]:
 
     async def read_configuration(self) -> Dict[str, Reading]:
         return await merge_gathered_dicts(sig.read() for sig in self._conf_signals)
+
+
+KT = TypeVar("KT")
+
+
+class DeviceDict(Dict[KT, Device], Device):
```
coretl, line 624: @tacaswell this started as a `DeviceVector(d1, d2, d3)`, and became `DeviceDict({1: d1, 2: d2, 3: d3})` so we could be 1-indexed and gappy. I think using a dict is correct, but should we allow any typed keys, or should we only support int keys and call it a `DeviceVector`? If so, should we support both `DeviceVector(d1, d2, d3)` and `DeviceVector({1: d1, 2: d2, 3: d3})` to construct the same thing?

ophyd/v2/tests/test_core.py
```diff
@@ -1,10 +1,39 @@
...
+    StandardReadable,
+    get_device_children,
+)
+
+
+@pytest.fixture(scope="function", params=[False, True])
+def RE(request):
+    loop = asyncio.new_event_loop()
+    loop.set_debug(True)
+    RE = RunEngine({}, call_returns_result=request.param, loop=loop)
```
coretl, line 23: Why do we want to test it with `call_returns_result` true and false?

coretl: Wait, do we even use this?

RAYemelyanova: no - will remove.

RAYemelyanova:
tests seem to be failing because 'codecov'  in the dev packages on the setup.cfg file doesn't seem to exist on pypi... Have now changed this to pytest-codecov, although I'm not sure why we want codecov in the first place. I thought that was a github CI thing.

coretl:
> tests seem to be failing because 'codecov' in the dev packages on the setup.cfg file doesn't seem to exist on pypi... Have now changed this to pytest-codecov, although I'm not sure why we want codecov in the first place. I thought that was a github CI thing.

It looks like this goes back to travis days. Try deleting it from setup.cfg, if it passes CI then I'm sure it will be fine

RAYemelyanova:
Tests all pass after changing codecov to pytest-codecov. This is ready for a review/merge.

callumforrester (approved):

tacaswell:
The context of codecov is that they used to have a bash uploader and a bunch of language specific uploaders (on the theory that you should upload using a tool in the language you were already working in to make the installation easy).  codecov has been trying to killoff the language specific uploaders and we switch to using the GHA a while ago but did not purge the dependency.

I suspect that `pytest-codecov` is also no longer needed and can also be removed.

context: https://community.codecov.com/t/codecov-yanked-from-pypi-all-versions/4259

PR #1111, opened 2023-04-25
v2: Make sim signal source look like real source

coretl:


RAYemelyanova (approved):
looks good to me. feel free to press big green button.

PR #1114, opened 2023-04-28
v2 Fix traceback of a failed AsyncStatus object

coretl:
Something in the catch-throw-catch-throw cycle was confusing the traceback. Either way, this is clearer. Added a test to check.

### DIFF
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -28,6 +28,7 @@ from typing import (
     Tuple,
     Type,
     TypeVar,
+    Union,
     cast,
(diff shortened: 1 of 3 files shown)


coretl:
CI will not pass on this unless we point CI at bluesky master, or we get a bluesky release

RAYemelyanova (approved):
looks good to me! Thanks for doing this

coretl:
I'm going to fix the lint errors resulting from Stageable protocol updates in another PR, so will merge this one now

Issue #1117, opened 2023-05-17
v2: SimSignalBackend throws an error if it's given a datatype of npt.NDArray


RAYemelyanova:
## Overview
some PV's return arrays of values, for example npt.NDArray[np.float64].

If this value is passed to a SimSignalBackend, it will fail with a TypeError. This is a problem if I want to use any simulated devices with array PVs like this example.

## Steps to reproduce

```python
import numpy as np
import numpy.typing as npt
from ophyd.v2.core import SimSignalBackend

sim_backend = SimSignalBackend(npt.NDArray[np.float64], "SOME-IOC:PV")
```

## Output

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/.../venv/lib/python3.9/site-packages/ophyd/v2/core.py", line 438, in __init__
    self._initial_value = datatype()
TypeError: ndarray() missing required argument 'shape' (pos 1)
```

## The Fix

The fix to this, is to simply add another elif clause to the existing __init__ method of the `SimSignalBackend`:

```python
class SimSignalBackend(SignalBackend[T]):
    """An simulated backend to a Signal, created with ``Signal.connect(sim=True)``"""

    _initial_value: T
    _value: T
    _timestamp: float

    def __init__(self, datatype: Optional[Type[T]], source: str) -> None:
        self.datatype = datatype
        self.source = f"sim://{source.split('://', 1)[-1]}"
        #: If cleared, then any ``put(wait=True)`` will wait until it is set
        self.put_proceeds = asyncio.Event()
        self.put_proceeds.set()
        self.listeners: List[SimMonitor[T]] = []
        if datatype is None:
            self._initial_value = cast(T, None)
        elif issubclass(datatype, Enum):
            self._initial_value = cast(T, list(datatype)[0])
        elif get_origin(datatype) == np.ndarray:
            self._initial_value = datatype(shape=1) #I dont know if 1 is best here.
        else:
            self._initial_value = datatype()
        self.set_value(self._initial_value)
```

I'm going to make a PR to address this.

PR #1118, opened 2023-05-17
Let SimSignalBackend accept npt.NDArray datatypes

RAYemelyanova:
solves https://github.com/bluesky/ophyd/issues/1117

### DIFF
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -534,6 +534,15 @@ class SimSignalBackend(SignalBackend[T]):
         if wait:
             await asyncio.wait_for(self.put_proceeds.wait(), timeout)
 
+    def _set_value(self, value: T):
+        """Method to bypass asynchronous logic, designed to only be used in tests."""
+        self._value = value
+        self._timestamp = time.monotonic()
+        self._reading = self.converter.reading(self._value, self._timestamp, self._severity)
+
+        if self.callback:
+            self.callback(self._reading, self._value)
+
     async def get_descriptor(self) -> Descriptor:
         return self.converter.descriptor(self.source, self._value)
 
@@ -550,9 +559,9 @@ class SimSignalBackend(SignalBackend[T]):
         self.callback = callback
 
 
-async def set_sim_value(signal: Signal[T], value: T):
-    """Set the value of a signal that is in sim mode"""
-    await _sim_backends[signal].put(value)
+def set_sim_value(signal: Signal[T], value: T):
+    """Set the value of a signal that is in sim mode."""
+    _sim_backends[signal]._set_value(value)
 
 
 def set_sim_put_proceeds(signal: Signal[T], proceeds: bool):
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -477,7 +477,7 @@ class DisconnectedSimConverter(SimConverter):
 
 
 def make_converter(datatype):
-    is_array = get_origin(datatype) == np.ndarray
+    is_array = get_dtype(datatype) is not None
     is_sequence = get_origin(datatype) == abc.Sequence
     is_enum = issubclass(datatype, Enum) if inspect.isclass(datatype) else False
 
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -422,9 +422,9 @@ class SimSignalBackend(SignalBackend[T]):
         if datatype is None:
             self._initial_value = cast(T, None)
         elif issubclass(datatype, Enum):
-            self._initial_value = cast(T, list(datatype)[0])
+            self._initial_value = cast(T, list(datatype)[0].value)
         elif get_origin(datatype) == np.ndarray:
-            self._initial_value = datatype(shape=1)  # type: ignore
+            self._initial_value = cast(T, datatype(shape=0))  # type: ignore
         else:
             self._initial_value = datatype()
         self.set_value(self._initial_value)
@@ -445,15 +445,15 @@ class SimSignalBackend(SignalBackend[T]):
             dtype = primitive_dtypes[type(self._value)]
             shape = []
         except KeyError:
-            if isinstance(self._value, Sequence):
-                dtype = "array"
-                shape = [len(self._value)]
-            elif isinstance(self._value, Enum):
+            if isinstance(self._value, Enum):
                 dtype = "string"
                 shape = []
                 choices = [e.value for e in type(self._value)]
                 # Ignore type until https://github.com/bluesky/event-model/issues/235#issuecomment-1497353265
                 return dict(source=self.source, dtype=dtype, shape=shape, choices=choices)  # type: ignore
+            elif isinstance(self._value, Sequence):
+                dtype = "array"
+                shape = [len(self._value)]
             else:
                 assert False, f"Can't get dtype for {type(self._value)}"
         return dict(source=self.source, dtype=dtype, shape=shape)
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -424,7 +424,7 @@ class SimSignalBackend(SignalBackend[T]):
         elif issubclass(datatype, Enum):
             self._initial_value = cast(T, list(datatype)[0])
         elif get_origin(datatype) == np.ndarray:
-            self._initial_value = datatype(shape=1)
+            self._initial_value = datatype(shape=1)  # type: ignore
         else:
             self._initial_value = datatype()
         self.set_value(self._initial_value)
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -5,6 +5,7 @@ import asyncio
 import functools
 import inspect
 import logging
+import re
 import sys
 import time
 from abc import abstractmethod
@@ -448,11 +449,11 @@ class SimArrayConverter(SimConverter):
 class SimEnumConverter(SimConverter):
     enum_class: Type[Enum]
 
-    def write_value(self, value: Union[Enum, str]):
+    def write_value(self, value: Union[Enum, str]) -> Enum:
         if isinstance(value, Enum):
-            return value.value
-        else:
             return value
+        else:
+            return self.enum_class(value)
 
     def descriptor(self, source: str, value) -> Descriptor:
         choices = [e.value for e in self.enum_class]
@@ -493,10 +494,11 @@ class SimSignalBackend(SignalBackend[T]):
     _reading: Reading
 
     def __init__(self, datatype: Optional[Type[T]], source: str) -> None:
+        pv = re.split(r"://", source)[-1]
+        self.source = f"sim://{pv}"
         self.datatype = datatype
         self.pv = source
         self.converter: SimConverter = DisconnectedSimConverter()
-        self.source = f"sim://{source}"
         self.put_proceeds = asyncio.Event()
         self.put_proceeds.set()
         self.callback: Optional[ReadingValueCallback[T]] = None
@@ -508,7 +510,7 @@ class SimSignalBackend(SignalBackend[T]):
         self._severity = 0
         self._reading = self.converter.reading(self._initial_value, time.monotonic(), self._severity)
 
-        await self.put(None)
+        await self.put(self._initial_value)
 
     async def put(self, value: Optional[T], wait=True, timeout=None):
         if value is None:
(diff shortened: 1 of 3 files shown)


RAYemelyanova:
I would ideally like a test to check that, for the pva/ca case, a type of npt.NDArray also works as expected. However, I'm struggling thinking of how to do this we this would mean ophyd/v2/tests/test_record.db would need to output something like a npt.NDArray[np.float64].... This db file has waveforms which are tested in the form of TypedDicts (e.g. the `Table` object) but not quite the same as, for example, `PC_OUT` on a zebra, which is just a stream of floats (or an array of them).

ophyd/v2/core.py
```diff
@@ -433,7 +435,9 @@ def __init__(self, datatype: Optional[Type[T]], source: str) -> None:
         if datatype is None:
             self._initial_value = cast(T, None)
         elif issubclass(datatype, Enum):
-            self._initial_value = cast(T, list(datatype)[0])
+            self._initial_value = cast(T, list(datatype)[0].value)
+        elif get_origin(datatype) == np.ndarray:
+            self._initial_value = cast(T, datatype(shape=0))  # type: ignore
```
coretl, line 440: Do you still need the type ignore?

ophyd/v2/core.py
```diff
@@ -433,7 +435,9 @@ def __init__(self, datatype: Optional[Type[T]], source: str) -> None:
         if datatype is None:
             self._initial_value = cast(T, None)
         elif issubclass(datatype, Enum):
-            self._initial_value = cast(T, list(datatype)[0])
+            self._initial_value = cast(T, list(datatype)[0].value)
```
coretl, line 438: Why the value of an Enum rather than an Enum itself?

ophyd/v2/tests/test_core.py
coretl, line 1: I don't have time to look at the parameterized tests at the moment, if you need this merged I suggest you take them back out and leave just the single test you first wrote, then make an issue for adding them back in

RAYemelyanova:
I've decided to change how `SimSignalBackend` works, to be more inline with the other backends, i.e. to have converters. 
In reality these converters don't actually need to 'convert' anything but it's a neat way to keep logic separated. I have kept methods like '.value' on the converters because I like the idea of the `SimSignalBackend` giving you a `DisconnectedError` if the converter hasn't been set properly, and the easiest way of doing this is by going through the converter for things like get_value and maybe even get_reading.

What do you think? @coretl and @evalott100

RAYemelyanova:
Jobs will fail for now, there's a couple of things that still need sorting out.

evalott100:
> What do you think?

I'll defer to Tom for that kind of decision (I haven't worked much with ophyd), but I'll follow the PR :slightly_smiling_face:

ophyd/v2/core.py
```diff
@@ -416,82 +418,135 @@ def close(self):
...
+
+class DisconnectedSimConverter(SimConverter):
+    def __getattribute__(self, __name: str) -> Any:
+        raise NotImplementedError("No PV has been set as connect() has not been called")
+
+
+def make_converter(datatype):
+    if datatype is None:
+        return SimConverter()
+    if (get_origin(datatype) == np.ndarray) or (get_origin(datatype) == abc.Sequence):
```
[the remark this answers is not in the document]
coretl: There's a function in core to do this:
https://github.com/bluesky/ophyd/blob/7612b2c9de9d5bc16cf28eea79ba5c12553f3cc2/ophyd/v2/core.py#L952-L964

ophyd/v2/core.py
```diff
@@ -416,82 +418,135 @@ def close(self):
...
+
+@dataclass
+class SimEnumConverter(SimConverter):
+    enum_class: Type[Enum]
+
+    def write_value(self, value: Union[Enum, str]):
+        if isinstance(value, Enum):
+            return value.value
+        else:
+            return value
```
coretl, line 461: As we discussed, I don't think we should do any actual conversions in the converter

ophyd/v2/core.py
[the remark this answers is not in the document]
evalott100: Makes sense, seems like `typing_extensions.get_origin` doesn't event work for `npt.NDArray`s anyway, while `__origin__` does, weird.

``` python
>>> x = npt.NDArray[np.uint8]
>>> x
numpy.ndarray[typing.Any, numpy.dtype[numpy.uint8]]
>>> get_origin(x) == None
True
>>> x.__origin__
<class 'numpy.ndarray'>
```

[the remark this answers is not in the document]
coretl: Interesting, this probably needs a comment in the code to explain this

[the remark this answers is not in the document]
RAYemelyanova: In the above, I am just interested in whether or not my array is a datatype...

```python
>>> datatype = npt.NDArray[np.int32]
>>> get_origin(datatype) == npt.NDArray
False
>>> get_origin(datatype) == np.ndarray
True
>>> get_dtype(datatype) == np.ndarray
False
```

Using `get_dtype` as suggested by @coretl wouldn't work in this case, where I am just trying to see if the datatype is an array, not what that array contains.

ophyd/v2/core.py
RAYemelyanova: I've decided I'd like to keep simple conversions; the reason being, we can have a SimSignalBackend that fails with a 'Disconnected' message if `connect()` hasn't been called, in the same way the other backends do.

The nice thing about these simulated converter classes is we can keep logic that is separate to each 'type' in its own class rather than munging them together in the SimSignalBackend.

ophyd/v2/core.py
```diff
@@ -402,96 +405,150 @@ def monitor_reading_value(self, callback: ReadingValueCallback[T]) -> Monitor:
...
-        self._listeners = listeners
-        self._listeners.append(self)
+    task: Awaitable[asyncio.Task[None]]
+
+    def __init__(self, task: asyncio.Task[None]):
+        self.task = cast(Awaitable[asyncio.Task[None]], task)
 
     def close(self):
-        self._listeners.remove(self)
+        pass
```
coretl, line 417: We need to make sure that `close` actually stops updates.

However, thinking about this, a `SignalBackend` shouldn't have to handle multiple subscribers, as there is only one, the `_SignalCache`.

How about changing `monitor_reading_value(cb: Callback[...]) -> Monitor` to `set_callback(cb: Optional[Callback[...]]) -> None` which tells the backend of a possible single subscriber to be notified when the reading and value changes. It should still have the same behaviour of immediately calling it on the current value, but means there is no need to keep a list of listeners in the backend.

coretl (changes requested):

ophyd/v2/core.py
RAYemelyanova: So should the stuff using `Monitor`, mainly SimSignalBackend, now use `_SignalCache` instead?

RAYemelyanova:
just investigating why the CI job is failing... it shouldn't be. I'm unsure as to how this is happening as I can't recreate this on my machine.

RAYemelyanova:
should be fixed now... :crossed_fingers:

ophyd/v2/core.py
```diff
@@ -403,86 +408,151 @@ def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
...
+    def reading(self, value: T, timestamp: float, severity: int) -> Reading:
+        return Reading(
+            value=value,
+            timestamp=timestamp,
+            alarm_severity=-1 if severity > 2 else severity,
+        )
+
+    def descriptor(self, source: str, value) -> Descriptor:
+        assert (
+            type(value) in primitive_dtypes.keys()
```
coretl, line 427: Nit
```suggestion
            type(value) in primitive_dtypes
```

ophyd/v2/core.py
```diff
@@ -403,86 +408,151 @@ def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
...
-        self._timestamp = time.monotonic()
-        if self.callback:
-            self.callback(self._reading, self._value)
-
 
-def set_sim_value(signal: Signal[T], value: T):
+async def set_sim_value(signal: Signal[T], value: T):
     """Set the value of a signal that is in sim mode"""
-    _sim_backends[signal].set_value(value)
+    await _sim_backends[signal].put(value)
```
coretl, line 555: Please can you turn this back to a regular function rather than an async one, it's important that we can set the simulated value of a signal without yielding to asyncio so we can do some of the more tricky sequencing of tests

coretl (changes requested):
Almost there

ophyd/v2/core.py
RAYemelyanova: All the tests are passing; what kind of sequencing are we talking about?

I think it makes the SimSignalBackend messier and less like it's non-simulated counterparts.

If you'd like I can write a test that means I have to use a non-async function. Although there are no such existing tests and you'd have to describe a situation which would warrant this.

RAYemelyanova:
I've made the change you requested; I don't like it, because I don't understand why SimSignalbackend can't behave exactly like the other backends, and why it needs a non-synchronous method equivalent to a put method. my issue with this, is people may use the simulated backend in a way that they wouldn't do with an aioca/p4p one, so if they ever swap it out things fail.

coretl:
> I've made the change you requested; I don't like it, because I don't understand why SimSignalbackend can't behave exactly like the other backends, and why it needs a non-synchronous method equivalent to a put method. my issue with this, is people may use the simulated backend in a way that they wouldn't do with an aioca/p4p one, so if they ever swap it out things fail.

The issue is that sometimes (rarely) we want to test the case where 2 signals update on the backend and fire callbacks before our coroutine fires. This is only predictable if we can set 2 sim values in quick succession without another coroutine firing. This is only guaranteed if `set_sim_value` is sync. Using the sim backend is about being able to make predictable things that could happen unpredictably with the aioca/p4p ones (because of the inherent race conditions of multiple PVs updating over network sockets).

ophyd/v2/core.py
```diff
@@ -403,86 +408,160 @@ def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
...
         if value is None:
-            self.set_value(self._initial_value)
+            write_value = self._initial_value
         else:
-            self.set_value(value)
+            write_value = self.converter.write_value(value)
+
+        self._value = write_value
+        self._timestamp = time.monotonic()
+        self._reading = await self.get_reading()
```
coretl, line 530: Duplicated code, can these use `_set_value` instead?

ophyd/v2/core.py
```diff
@@ -403,86 +408,160 @@ def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
...
-        return dict(value=self._value, timestamp=self._timestamp)
+        if self.callback:
+            self.callback(self._reading, self._value)
+
+    async def get_descriptor(self) -> Descriptor:
+        return self.converter.descriptor(self.source, self._value)
 
     async def get_reading(self) -> Reading:
-        return self._reading
+        return self.converter.reading(self._value, self._timestamp, self._severity)
```
coretl, line 550: Why do we stash `self._reading` sometimes but not here? Maybe we should never stash it, and create it on demand?

tacaswell:
Very superficial question: do we want to put simulation related code in its own module rather than `core`?

RAYemelyanova:
> Very superficial question: do we want to put simulation related code in its own module rather than `core`?

Yes - although I think that should come as another PR in the form of refactoring anyways. There's a lot of stuff in core.py that should be separated out in my opinion. For example all the signals should probably go into a signals.py file instead of core.py, and there are plenty of what most people would consider 'utility' functions like wait_for_connection that should probably go in a utils.py, although I know there are mixed opinions about the code-cleanliness of the existence of a utils.py.

Either way I think this needs a broader discussion.

ophyd/v2/core.py
```diff
@@ -403,86 +408,156 @@ def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
...
     """An simulated backend to a Signal, created with ``Signal.connect(sim=True)``"""
 
-    _initial_value: T
     _value: T
+    _initial_value: T
     _timestamp: float
+    _severity: int
 
     def __init__(self, datatype: Optional[Type[T]], source: str) -> None:
+        pv = re.split(r"://", source)[-1]
```
coretl, line 501: I'm curious, did you have a reason to use `re.split` rather than `str.split`? Either is fine by me.

coretl (approved):

ophyd/v2/core.py
RAYemelyanova: hm... not particularly. No idea why that's re.split!

PR #1119, opened 2023-05-18
TernaryDevice and ArrayDevice

gwbischof:


ophyd/tests/test_array.py
```diff
@@ -0,0 +1,175 @@
...
+            self._state = None
+            await asyncio.sleep(self._delay)
+            self._state = False
+
+    @property
+    def state(self):
+        return self._state
+
+
+class TernaryArrayIOC(PVGroup):
```
gwbischof, line 52: I plan to update this to use `type` to create the class.

ophyd/array.py
```diff
@@ -0,0 +1,61 @@
+from functools import reduce
+from ophyd import Device, DeviceStatus
+
+
+def ArrayDevice(devices, *args, **kwargs):
```
gwbischof, line 5: Using a function here makes it so that it can't be subclassed.  This should be updated to fix that.

PR #1123, opened 2023-06-09
v2: Make stage/unstage async

coretl:
This fixes the linting issues on master, and improves the StandardReadable interface

### DIFF
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -21,7 +21,6 @@ from typing import (
     Iterable,
     List,
     Optional,
-    Protocol,
     Sequence,
     Set,
     Tuple,
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -8,7 +8,6 @@ import sys
 import time
 from abc import abstractmethod
 from contextlib import suppress
-from dataclasses import dataclass
 from enum import Enum
 from typing import (
     Any,
@@ -30,7 +29,6 @@ from typing import (
     TypeVar,
     Union,
     cast,
-    runtime_checkable,
 )
 
 import numpy as np
@@ -145,6 +143,9 @@ class Device(HasName):
     #: The parent Device if it exists
     parent: Optional[Device] = None
 
+    def __init__(self, name: str = "") -> None:
+        self.set_name(name)
+
     @property
     def name(self) -> str:
         """Return the name of the Device"""
@@ -231,8 +232,9 @@ async def connect_children(device: Device, sim: bool):
 
 def name_children(device: Device, name: str):
     """Call ``child.set_name(child_name)`` on all child devices in series."""
-    for child_name, child in get_device_children(device):
-        child.set_name(f"{name}-{child_name.rstrip('_')}")
+    for attr_name, child in get_device_children(device):
+        child_name = f"{name}-{attr_name.rstrip('_')}" if name else ""
+        child.set_name(child_name)
         child.parent = device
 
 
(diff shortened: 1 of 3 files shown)


ophyd/v2/core.py
```diff
@@ -145,6 +143,9 @@ class Device(HasName):
     #: The parent Device if it exists
     parent: Optional[Device] = None
 
+    def __init__(self, name: str = "") -> None:
+        self.set_name(name)
```
RAYemelyanova, line 147: why is this no longer needed? Does HasName have a constructor similar to this anyway?

ophyd/v2/core.py
```diff
@@ -788,57 +787,6 @@ async def _wait_for_value():
...
-class AsyncReadable(Protocol):
-    """Readable narrowed to only be async"""
-
-    @abstractmethod
-    async def read(self) -> Dict[str, Reading]:
-        ...
-
-    @abstractmethod
-    async def describe(self) -> Dict[str, Descriptor]:
-        ...
```
RAYemelyanova, line 801: so is the idea that now, all readables are async in v2?

ophyd/v2/core.py
```diff
@@ -788,57 +787,6 @@ async def _wait_for_value():
...
-
-@dataclass
-class _UncachedSignal(AsyncReadable):
-    signal: SignalR
-
-    async def read(self) -> Dict[str, Reading]:
-        return await self.signal.read(cached=False)
-
-    async def describe(self) -> Dict[str, Descriptor]:
-        return await self.signal.describe()
```
RAYemelyanova, line 839: presuming this has been deleted because all signals are cached?

RAYemelyanova (approved):
All makes sense, I have just a few questions for my own understanding.

ophyd/v2/core.py
coretl: The constructor doesn't appear in the HasName protocol. I'm adding this here rather than in StandardReadable because the `epicsdemo.SampleStage` isn't actually `Readable`, so adding this allows it to inherit from `Device` rather than `StandardReadable`

ophyd/v2/core.py
coretl: I simplified `StandardReadable` so it only reads `SignalR`s, which meant I no longer needed this intermediate protocol for "things that `StandardReadable` can read"

ophyd/v2/core.py
coretl: Deleted as I found a better way to do this in `StandardReadable` that didn't need this intermediate class

RAYemelyanova: ah; this makes a lot of sense. StandardReadable now has cached and uncached signals as part of its definition, making this bit unnecessary. understood.

ophyd/v2/core.py
```diff
@@ -854,74 +802,68 @@ async def merge_gathered_dicts(
...
+    async def describe(self) -> Dict[str, Descriptor]:
+        return await merge_gathered_dicts(
+            [sig.describe() for sig in self._read_signals + self._read_uncached_signals]
+        )
+
+    async def read(self) -> Dict[str, Reading]:
+        return await merge_gathered_dicts(
+            [sig.read() for sig in self._read_signals]
+            + [sig.read(cached=False) for sig in self._read_uncached_signals]
+        )
```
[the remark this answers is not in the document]
coretl: Let's assume you need to set pv1 to a value, and when that completes you can read pv2. If you use cached signals this will be equivalent to:

```python
values = []
camonitor(pv1, values.append)
caput(pv2, wait=True)
return values[-1]
```

However there is a race condition between the caput callback coming in, and the monitor update coming in. To be safe, we should use uncached signals for these, which is less timely but guaranteed to be correct:

```python
caput(pv2, wait=True)
return caget(pv1)
```

PR #1124, opened 2023-06-16
v2: Add set_and_wait_for_value()

coretl:
Useful for busy record, or other Signals with pattern:
- Set Signal with wait=True and stash the Status
- Read the same Signal to check the operation has started
- Return the Status so calling code can wait for operation to complete

Merge after #1123


RAYemelyanova (approved):
looks good to me

PR #1125, opened 2023-06-16
v2: remove monitor

RAYemelyanova:
simple PR to remove `Monitor` and `SimMonitor` from ophyd/v2/core.py.

It has been agreed that it makes sense if the backend only keeps track of one single callback. Therefore, we no longer need a `Monitor` to keep track of potentially several callbacks.

### DIFF
--- a/ophyd/v2/_aioca.py
+++ b/ophyd/v2/_aioca.py
@@ -4,7 +4,15 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Dict, Optional, Sequence, Type, Union
 
-from aioca import FORMAT_CTRL, FORMAT_RAW, FORMAT_TIME, Subscription, caget, camonitor, caput
+from aioca import (
+    FORMAT_CTRL,
+    FORMAT_RAW,
+    FORMAT_TIME,
+    Subscription,
+    caget,
+    camonitor,
+    caput,
+)
 from aioca.types import AugmentedValue, Dbr, Format
 from bluesky.protocols import Descriptor, Dtype, Reading
 from epicscorelibs.ca import dbr
--- a/ophyd/v2/_aioca.py
+++ b/ophyd/v2/_aioca.py
@@ -230,3 +230,5 @@ class CaSignalBackend(SignalBackend[T]):
                 datatype=self.converter.read_dbr,
                 format=FORMAT_TIME,
             )
+        else:
+            self.subscription = None
--- a/ophyd/v2/_aioca.py
+++ b/ophyd/v2/_aioca.py
@@ -220,10 +220,10 @@ class CaSignalBackend(SignalBackend[T]):
         return self.converter.value(value)
 
     def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
-        if self.subscription:
-            self.subscription.close()
-            self.subscription = None
         if callback:
+            assert (
+                not self.subscription
+            ), "Cannot set a callback when one is already set"
             self.subscription = camonitor(
                 self.read_pv,
                 lambda v: callback(self.converter.reading(v), self.converter.value(v)),
--- a/ophyd/v2/_aioca.py
+++ b/ophyd/v2/_aioca.py
@@ -231,4 +231,6 @@ class CaSignalBackend(SignalBackend[T]):
                 format=FORMAT_TIME,
             )
         else:
+            if self.subscription:
+                self.subscription.close()
             self.subscription = None
--- a/ophyd/v2/_aioca.py
+++ b/ophyd/v2/_aioca.py
@@ -4,13 +4,12 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Dict, Optional, Sequence, Type, Union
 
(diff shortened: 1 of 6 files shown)


RAYemelyanova:
I'm getting the following, seemingly unrelated, mypy errors:

```
(venv) rose@rose-Precision-3560:~/Documents/projects/ophyd$ mypy ophyd/v2/
ophyd/v2/core.py:805: error: Incompatible return value type (got "Union[Status, List[Any]]", expected "List[Any]")  [return-value]
ophyd/v2/core.py:811: error: Incompatible return value type (got "Union[Status, List[Any]]", expected "List[Any]")  [return-value]
ophyd/v2/core.py:888: error: Argument 1 to "__iadd__" of "list" has incompatible type "Union[Status, List[Any]]"; expected "Iterable[StandardReadable]"  [arg-type]
ophyd/v2/core.py:896: error: Argument 1 to "__iadd__" of "list" has incompatible type "Union[Status, List[Any]]"; expected "Iterable[StandardReadable]"  [arg-type]
Found 4 errors in 1 file (checked 9 source files)
```

RAYemelyanova:
^ can confirm that I get these errors even on the master branch... I will  ` # type: ignore` them for now.

coretl:
> ^ can confirm that I get these errors even on the master branch... I will ` # type: ignore` them for now.

They will be fixed by #1123

ophyd/v2/core.py
```diff
@@ -510,9 +493,9 @@ def set_sim_put_proceeds(signal: Signal[T], proceeds: bool):
         event.clear()
 
 
-def monitor_sim_value(signal: Signal[T], callback: ReadingValueCallback[T]) -> Monitor:
+def set_sim_callback(signal: Signal[T], callback: ReadingValueCallback[T]) -> None:
     """Monitor the value of a signal that is in sim mode"""
-    return _sim_backends[signal].monitor_reading_value(callback)
+    return _sim_backends[signal].set_callback(callback)
```
coretl, line 498: Hmm, I forgot we did this, this leads to a hard-to-debug failure mode:
```python
set_sim_callback(print)
sig.put(value)  # prints value
sig.stage()  # replaces the callback
sig.put(value)  # doesn't print value anymore
```
I think we should probably make the `set_callback` methods more strict so it errors if you try and set a callback and one is already set

ophyd/v2/core.py
```diff
@@ -484,16 +465,18 @@ async def get_reading(self) -> Reading:
...
         return self._value
 
-    def monitor_reading_value(self, callback: ReadingValueCallback[T]) -> Monitor:
-        callback(self._reading, self._value)
-        return SimMonitor(callback, self.listeners)
+    def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
+        if callback:
+            callback(self._reading, self._value)
+
+        self.callback = callback
```
coretl, line 472: E.g.
```suggestion
    def set_callback(self, callback: Optional[ReadingValueCallback[T]]) -> None:
        if callback:
            assert not self.callback, "Cannot set a callback when one is already set"
            callback(self._reading, self._value)
        self.callback = callback
```

coretl (changes requested):

ophyd/v2/core.py
RAYemelyanova: Could you expand this example? Am struggling to recreate it.

Should I make an unset_callback method to a SignalBackend?

ophyd/v2/core.py
RAYemelyanova: should I also apply similar logic for the aioca/p4p backends?

ophyd/v2/core.py
RAYemelyanova: I think I've done this now for all use cases.

ophyd/v2/_aioca.py
```diff
@@ -211,10 +219,16 @@ async def get_value(self) -> T:
...
+                not self.subscription
+            ), "Cannot set a callback when one is already set"
+            self.subscription = camonitor(
+                self.read_pv,
+                lambda v: callback(self.converter.reading(v), self.converter.value(v)),
+                datatype=self.converter.read_dbr,
+                format=FORMAT_TIME,
+            )
+        else:
+            self.subscription = None
```
coretl, line 234: Need to close the subscription first
```suggestion
            if self.subscription:
                self.subscription.close()
            self.subscription = None
```

ophyd/v2/_p4p.py
```diff
@@ -223,10 +223,17 @@ async def get_value(self) -> T:
...
-            self.read_pv, async_callback, request="field(value,alarm,timestamp)"
-        )
+            async def async_callback(v):
+                callback(self.converter.reading(v), self.converter.value(v))
+
+            self.subscription = self.ctxt.monitor(
+                self.read_pv, async_callback, request="field(value,alarm,timestamp)"
+            )
+        else:
+            self.subscription = None
```
coretl, line 239: Ditto
```suggestion
            if self.subscription:
                self.subscription.close()
            self.subscription = None
```

coretl (changes requested):

coretl (approved):

Issue #1113, opened 2023-04-27
v2: non-descriptive error for epics_signal factories if aioca or p4p not installed.


RAYemelyanova:
# Summary of Problem
If aioca is not installed, making an epics_signal_r (or rw, or x .. etc) will fail with a non-descriptive error message:

```
AttributeError: 'function' object has no attribute 'value'
```

This is unexpected behaviour and the error is obscure.

## Description, with examples

If aioca or p4p are not installed, ophyd/v2/epics.py has:

```python
try:
    from ._aioca import CaSignalBackend
except ImportError as ca_error:

    def CaSignalBackend(*args, ca_error=ca_error, **kwargs):  # type: ignore
        raise NotImplementedError("CA support not available") from ca_error

```

and equivalent for p4p. ._aioca and ._p4p contain import p4p, and import aioca, which will fail with the import error. So if my venv doesn't have aioca for example, CaSignalBackend will be a function, not a class.

The issue with this, is later on if you want to make backends (as epics_signal_r does, or any epics signal factory)...

```python
def _transport_pv(pv: str) -> Tuple[EpicsTransport, str]:
    split = pv.split("://", 1)
    if len(split) > 1:
        # We got something like pva://mydevice, so use specified comms mode
        transport_str, pv = split
        transport = EpicsTransport[transport_str]
    else:
        # No comms mode specified, use the default
        transport = _default_epics_transport
    return transport, pv

def _make_backend(
    datatype: Optional[Type[T]], read_pv: str, write_pv: str
) -> SignalBackend[T]:
    r_transport, r_pv = _transport_pv(read_pv)
    w_transport, w_pv = _transport_pv(write_pv)
    transport = get_unique({read_pv: r_transport, write_pv: w_transport}, "transports")
    return transport.value(datatype, r_pv, w_pv)
```

_transport_pv should return a tuple of an enum and string. However,if I don't have aioca installed, CaSignalBackend is replaced with a function, which apparently isn't stored properly in the enum. This means that the call to `transport.value` fails, because `transport` is then just a function, instead of an Enum. In ipython...

```python
In [19]: EpicsTransport.pva
Out[19]: <EpicsTransport.pva: <class 'ophyd.v2._p4p.PvaSignalBackend'>>

In [20]: EpicsTransport.ca
Out[20]: <function ophyd.v2.epics.CaSignalBackend(*args, ca_error=ModuleNotFoundError("No module named 'aioca'"), **kwargs)>

[...]

PR #1131, opened 2023-07-04
Use classes for fallback epics signal backends

tpoliaw:
Using functions means that the enums referencing them do not behave as
expected. eg

    def foo(): pass

    class Transport(Enum):
        a = foo

In this setup, `Transport` will have no member named `a` (only a class
method), and `Transport.a.value()` will raise an AttributeError.

### DIFF
--- a/ophyd/v2/epics.py
+++ b/ophyd/v2/epics.py
@@ -11,16 +11,18 @@ try:
     from ._aioca import CaSignalBackend
 except ImportError as ca_error:
 
-    def CaSignalBackend(*args, ca_error=ca_error, **kwargs):  # type: ignore
-        raise NotImplementedError("CA support not available") from ca_error
+    class CaSignalBackend:  # type: ignore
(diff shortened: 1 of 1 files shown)


ophyd/v2/epics.py
```diff
@@ -11,16 +11,18 @@
...
 
 
 try:
     from ._p4p import PvaSignalBackend
 except ImportError as pva_error:
 
-    def PvaSignalBackend(*args, pva_error=pva_error, **kwargs):  # type: ignore
-        raise NotImplementedError("PVA support not available") from pva_error
+    class PvaSignalBackend:
+        def __init__(*args, pva_error=pva_error, **kwargs):  # type: ignore
```
coretl, line 24: ignore needs to go on the line above

coretl (commented):
Thanks for this, fixes #1113

coretl:
Fixes #1113

coretl (approved):

PR #1137, opened 2023-07-21
ENH: pass the count parameter along, as already documented.

clintonroy:
The lowest layer of pv.get accepts a count, which specifies how many array values to return. This count argument was not percolated all the way up the api stack, but was documented.

I require this to be able to retrieve values from a data acquisition device that can have a very large array of results, and the default value (2048) is way too small.

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1283,7 +1283,7 @@ class EpicsSignalBase(Signal):
         return (self._metadata["lower_ctrl_limit"], self._metadata["upper_ctrl_limit"])
 
     def _get_with_timeout(
-        self, pv, timeout, connection_timeout, as_string, form, use_monitor
+        self, pv, timeout, connection_timeout, count, as_string, form, use_monitor
     ):
         """
         Utility method implementing a retry loop for get and get_setpoint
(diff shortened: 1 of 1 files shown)


ophyd/signal.py
```diff
@@ -1283,7 +1283,14 @@ def limits(self):
         return (self._metadata["lower_ctrl_limit"], self._metadata["upper_ctrl_limit"])
 
     def _get_with_timeout(
-        self, pv, timeout, connection_timeout, as_string, form, use_monitor
+        self,
+        pv,
+        timeout,
+        connection_timeout,
+        count,
```
tacaswell, line 1290: I'm a bit torn, on one hand adding new positionals in the middle is a source of latent bugs, but on the other hand all of the positionals are required so any usage is fully broken (`TypeError: missing parameter`) rather than the subtle "all your inputs are shifted".

On net I think it is 👍🏻 , just set off alarms in my head.

mrakitin:
Power-cycled the CI to check they still succeed against the latest version.

prjemian:
Consider adding an [option to the workflow](https://github.com/bluesky/hklpy/blob/dd2cf5a0d32e73f04d8ec5a223faff7e6f9593a2/.github/workflows/conda_unit_test.yml#L11) to run the unit tests manually.

image

mrakitin:
There was a recent fix to the tests (https://github.com/bluesky/ophyd/pull/1167), I hope it will help resolve the failing runs after my rebase on `master`.

PR #1140, opened 2023-07-31
ENH: Add __contains__ to AndStatus

### LINKED ISSUE #1139: Add a __contains__ to AndStatus
For our unit tests it would be useful to know if the status we're getting back from a signal contains a supplied signal. I propose creating a `__contain__` method on the `AndStatus` that recursively checks the status tree for a given object.

DominicOram:
Fixes #1139

Mainly useful for tests but also could be generally useful.


ophyd/status.py
```diff
@@ -569,6 +569,16 @@ def __str__(self):
             "".format(self.__class__.__name__, self)
         )
 
+    def __contains__(self, status: StatusBase):
```
callumforrester, line 572: Nit: use `bluesky.protocols.Status` rather than `StatusBase`, provides a little more flexibility.

callumforrester (approved):
Looks good to me!

One interesting use case is if someone in future were to write, say, an `OrStatus` and we might want `__contains__` to be interoperable. It might make sense to then make `__contains__` part of the `Status` API... Definitely out-of-scope for this PR but I could make an issue if there's appetite. @coretl any thoughts WRT ophyd-async?

callumforrester:
Yeah, I agree actually, stick to `asyncio.gather` rather than reinvent the wheel. @DominicOram sorry for coopting your PR to talk about this

DominicOram:
> Looks good to me!
> 
> One interesting use case is if someone in future were to write, say, an `OrStatus` and we might want `__contains__` to be interoperable. It might make sense to then make `__contains__` part of the `Status` API... Definitely out-of-scope for this PR but I could make an issue if there's appetite. @coretl any thoughts WRT ophyd-async?

Yes, I thought about spreading `__contains__` more widely and we thought we had a usecase for `OrStatus` at one point but it fell out after more discussion of the problem. For now I'm happy with this PR and don't see a strong need to extend it currently.

> Yeah, I agree actually, stick to asyncio.gather rather than reinvent the wheel. @DominicOram sorry for coopting your PR to talk about this

That's ok, I agree sticking with `gather` makes sense.

ophyd/status.py
DominicOram: Does this work? Seems to fail the tests as it looks like we have no current dependency on `bluesky`?

danielballan:
Regarding `OrStatus`, I implemented this in 2016 but we ripped it out before merging because it was unclear what to do with errors, and there were no particularly compelling use cases.

https://github.com/bluesky/ophyd/pull/241#issuecomment-188914204

ophyd/status.py
callumforrester: Ah, I didn't realise the bluesky dependency is dev-only, my bad! Can either ignore this comment and keep `StatusBase` or couch the import in `if TYPE_CHECKING`, I don't mind

DominicOram: Given the lack of use of `bluesky.protocols.Status` generally in v1 code I will revert to `StatusBase`

PR #1147, opened 2023-08-15
Allow write path template on FileStoreBase to include .

### LINKED ISSUE #1146: FileStoreBase write_path_template cannot be  "." if read_path_template is None
Resolving read_path_template uses self.write_path_template, which is the already resolved self._write_path_template

"." throws an exception while formatting the read_path_template, despite already having been checked for being a valid relative path and allowed by write_path_template
https://github.com/bluesky/ophyd/blob/5c413cc0848ef035b9839de328064d6af47b589b/ophyd/areadetector/filestore_mixins.py#L247

`@property
    def read_path_template(self):
        "Returns write_path_template if read_path_template is not set"
        rootp = self.reg_root

        if self._read_path_template is None:
            return self.write_path_template

        ret = PurePath(self._read_path_template)

        if rootp not in ret.parents:
            if not ret.is_absolute():
                ret = rootp / ret
            else:
                raise ValueError(
                    (
                        "root: {!r} in not consistent with " "read_path_template: {!r}"
                    ).format(rootp, ret)
                )
        ret = os.path.join(ret, "")
        return str(ret)`

callumforrester:
Fixes #1146 

`FileStoreBase` "validates" its write path template (unless a read path template is also supplied). In this context that means it checks if the store's root is an ancestor of the template, but does not allow `.` as the write path.

The use case for this is making a file store write to its own root. 

Changes in this PR to address this:

* Make `FileStoreBase` consider the value of root (or simply `.`) a valid write path under root.
* Refactor common behaviour out of read and write path properties
* Add suite of tests to validate that read and write paths are processed properly in a variety of use cases
* Check that behaviour holds whether root/write/read paths are passed in the constructor or set after object creation
* Add a minimal dummy implementation that is used in the tests without needing to run an IOC

Note: If we decide that `.` should be explicitly disallowed, then the first point is easily reversible and the others are hopefully still useful.

### DIFF
--- a/ophyd/areadetector/filestore_mixins.py
+++ b/ophyd/areadetector/filestore_mixins.py
@@ -26,7 +26,7 @@ import warnings
 from collections import defaultdict, deque
 from datetime import datetime
 from itertools import count
-from pathlib import PurePath, PurePosixPath, PureWindowsPath
+from pathlib import Path, PurePath, PurePosixPath, PureWindowsPath
 
 from ..device import BlueskyInterface, GenerateDatumInterface, Staged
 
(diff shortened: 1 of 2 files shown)


callumforrester:
Paging @tacaswell and @DiamondJoseph

tacaswell:
The refactoring seems good and I think I'm sold on this.  The requirement to include the root twice was never great (if I recall correctly, there was a path dependency is that we had read_path and write_path for a while and then we added the notion of root so added it in a way that would not break stuff on the floor).

----

A different API that we talked about at the time was a root level integer to say how many levels to consider as part of the root but we (incorrectly) judged that was more likely to end up with incorrect values than including the part of the path that was the root directly.

tacaswell (approved):

tacaswell:
We do not seem to check that `self.reg_root` is actually absolute which leaves the thing we are actually trying to prevent (which is relative paths leaking out into the documents).

PR #1151, opened 2023-08-21
v2: Refactor and Test DeviceCollector Error Handling

callumforrester:
Changes:

* Add tests to verify that `DeviceCollector` handles exceptions generated by `connect()`
* Add tests to verify that exceptions from devices failing to connect are logged
* Change error logging so that timeouts and exceptions from devices are logged in a uniform way
* Add comments to error handling code
* Make pre-existing tests that check the logs consistent with new tests

### DIFF
--- a/ophyd/v2/core.py
+++ b/ophyd/v2/core.py
@@ -327,25 +327,24 @@ class DeviceCollector:
 
     async def _wait_for_tasks(self, tasks: Dict[asyncio.Task, str]):
         done, pending = await asyncio.wait(tasks, timeout=self._timeout)
+
+        # Handle all devices where connection has timed out
         if pending:
-            msg = f"{len(pending)} Devices did not connect:"
+            logging.error(f"{len(pending)} Devices did not connect:")
             for t in pending:
                 t.cancel()
                 with suppress(Exception):
                     await t
-                e = t.exception()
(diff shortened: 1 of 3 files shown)


callumforrester:
Some additional issues have come out of this work
https://github.com/bluesky/ophyd/issues/1152
https://github.com/bluesky/ophyd/issues/1153

RAYemelyanova (approved):
It looks like the main thing you're changing here is how the error gets logged when a device fails to connect, either because of an exception or because of a timeout error.

Seems sensible

coretl:
My aim would be the following:
- If `Device.connect()` raises, then report its traceback in full
- If `Device.connect()` doesn't complete within timeout, cancel it, if it raises `NotConnected` then report the message text (which says which child signals are not connected) without traceback, otherwise report its traceback in full

This PR seems to report traceback for NotConnected errors too, do we want that?

callumforrester:
@coretl To get to where you want, I think we would first have to fix https://github.com/bluesky/ophyd/issues/1153
This PR doesn't change the current behaviour except to make the log formatting more consistent. The information logged is the same.
I suggest that sorting out exactly what is logged under which circumstances should be a separate PR. Possibly also a separate issue pulling together your comment and https://github.com/bluesky/ophyd/issues/1153

coretl:
Superceded in ophyd-async

PR #1155, opened 2023-09-02
ENH: Add a StableSubscriptionStatus

### LINKED ISSUE #1142: SubscriptionStatus that waits for state to be stable
We have a usecase where we would like to wait on a PV being a certain state and staying at that state for a given length of time. If in that time it changes to a different state we want to restart the timer for it to be `stable`.

It was suggested on slack that we could implement this with a `SubscriptionStatus` like object that rather than marking itself as done when the user function is True, starts a cancelable timer to do it and if it gets a False back from the user function cancels the timer.

We would also need to make sure that we do not spawn multiple timers.

DominicOram:
Fixes #1142.

Adds a stable subscription status that easily allows for checking that a component stays stable for some time. Also, adds docs and tests for it.

### DIFF
--- a/docs/user_v1/explanations/status.rst
+++ b/docs/user_v1/explanations/status.rst
@@ -184,6 +184,48 @@ Note that ``set_finished``, ``subscribe`` and ``clear_sub`` are gone; they are
 handled automatically, internally. See
 :class:`~ophyd.status.SubscriptionStatus` for additional options.
 
+StableSubscriptionStatus
+------------------------
+
+The :class:`~ophyd.status.StableSubscriptionStatus` is a Status object that is
+similar to the :class:`~ophyd.status.SubscriptionStatus` but is only marked
+finished based on an ophyd event remaining stable for some given time. For
+example, this could be used to ensure a temperature remains in a given range
+for a set amount of time:
+
+.. code:: python
+
+   from ophyd import Device, Component, StableSubscriptionStatus
+
+   class MyTempSensor(Device):
+       ...
+       # The set point and readback of a temperature that
+       # may fluctuate for a second before it can be considered set
+       temp_sp = Component(...)
+       temp_rbv = Component(...)
+       def set(self, set_value):
+           def check_value(*, old_value, value, **kwargs):
+               "Return True when the temperature is in a valid range."
+               return set_value - 0.01 < value < set_value + 0.01
+
+           status = StableSubscriptionStatus(self.temp_rbv, check_value, stability_time=1)
+           self.temp_sp.set(set_value)
+           return status
+
+The timer for ``stability_time`` is started when the callback condition first
+becomes true and stopped if it becomes false. It is then restarted if the
+condition becomes true again. This will continue until either the condition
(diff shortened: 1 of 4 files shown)


docs/user_v1/explanations/status.rst
```diff
@@ -184,6 +184,35 @@ Note that ``set_finished``, ``subscribe`` and ``clear_sub`` are gone; they are
...
 :class:`~ophyd.status.SubscriptionStatus` for additional options.
 
+StableSubscriptionStatus
+------------------------
+
+The :class:`~ophyd.status.StableSubscriptionStatus` is a Status object that is
+similar to the :class:`~ophyd.status.SubscriptionStatus` but is only marked
+finished based on an ophyd event remaining stable for some given time. For
+example, this could be used to ensure a temperature remains in a given range
+for a set amount of time:
```
callumforrester, line 194: There's an interesting subtlety here. Reading the docs, I would like to know whether the timer starts from the creation of the object or from the first response from the device. Also whether it starts again when the callback returns `False` or only after a `False` then a `True`. See options 1-4 below:


image

DominicOram: Thanks, the answer is 4. I think this is sort of implied but you're right it's not poperly stated, I'll fix the docs

ophyd/status.py
```diff
@@ -749,6 +750,110 @@ def _handle_failure(self):
...
+
+    def __init__(
+        self,
+        device,
+        callback,
+        stability_time,
+        event_type=None,
+        timeout=None,
+        settle_time=None,
+        run=True,
```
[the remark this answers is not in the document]
DominicOram: This seems hard given that I suspect it's type depends on what kind of signal it is attached to? Given this and that the signature remains the same as `SubscriptionStatus` I will respectively ignore the nit-pick

danielballan:
I have only skimmed the code but I will offer high-level comments. We have wanted something like this for awhile, and I think this is a quite elegant way to do it.

I think it is accurate to describe this as a kind of client-side deadband.

danielballan:
Note previous comment was extensible edited: I pressed submit early by accident.

DominicOram:
> Is it worth putting something to this effect in the docs? In summary, I would advice some caution in reaching for this feature (think: should I do this in the IOC?) but it comes up frequently enough that I absolutely agree this is worth having.
> 
> @ZLLentz @tangkong @klauer Have you ever developed something like this? Any thoughts?

Yes, I agree that it's cleaner to be server side but I think it's useful as you've said. I will add in some docs to this affect.

DominicOram:
Docs updated, feel free to add additional comments.

danielballan (approved):
The docs look great. Thanks!

I conceptually approve; I leave line review to Diamond.

callumforrester:
I think the docs build breaking may be legitimate, happy to approve once that's fixed

DominicOram:
> I think the docs build breaking may be legitimate, happy to approve once that's fixed

Docs now passing, are you happy to approve @callumforrester?

callumforrester (approved):

PR #1159, opened 2023-09-22
Remove (obsolete) internal records from FileBase

xiaoqiangwang:
These two were internal implementation details in NDFile.template to sync FileNumber. There is no use to expose them.

Moreover they have been removed since areaDetector 2.2, https://github.com/areaDetector/ADCore/commit/6fb231f5269adcbea01a93247a45255057577c07


tacaswell:
For the plugins we have AD version specific sub-classes.  You can mask out a removed PV by setting it to `Non` in the subclass.

tacaswell:
This is already done in https://github.com/bluesky/ophyd/blob/a0a79e2e99c3ed10890f42666fdccdbd5507537a/ophyd/areadetector/plugins.py#L1134-L1141

so you use the `XXX_V34` version of the plugins you should not need this change.

PR #1160, opened 2023-09-22
Improve fake epicsmotor

dperl-dls:
Following yesterday's discussion of Hyperion with @danielballan @DominicOram, here is a half-baked PR for improving FakeEpicsMotor, which just makes it what we do over and over for our unit tests in Hyperion. 

After writing this, I thought that it might be better/cleaner/something to handle this at a lower level, though. In particular, I had the idea of giving FakeSignal the ability to propagate sets and puts to it to another FakeSignal and then using that in the setup of the Motor instead.

Welcome any opinions.


danielballan:
Thanks @dperl-dls! We will take a look.

dperl-dls:
Closing as we now have a nice sim motor in ophyd-async

PR #1162, opened 2023-10-05
Remove user_v2 from the documentation

rosesyrett:
Now that ophyd.v2 has been ported over to ophyd_async, to prevent confusing users we should remove the ophyd user_v2 documentation.

ophyd-async has github pages running, and the bluesky.project.io repo already makes reference to it in the homepage and dropdown. the docs are identical, except for a few objects which are no longer references as they are stored in (the python equivalent of) private modules, which I've spoken to @coretl about, and which he seems to approve of.


rosesyrett:
Currently, the docs job will fail. This is because of the following warning:

```
/home/rose/Documents/projects/ophyd/ophyd/pv_positioner.py:docstring of ophyd.pv_positioner.PVPositionerComparator:12: WARNING: 'any' reference target not found: done
```

I'm not sure why this is happening and am investigating.

docs/index.rst
```diff
@@ -8,26 +8,17 @@ How the documentation is structured
...
+.. grid:: 1
 
     .. grid-item-card:: :material-regular:`person;4em`
-        :link: user_v1/index
+        :link: user/index
         :link-type: doc
 
         The User Guide v1 contains documentation on how to install and use ophyd's original v1 API.
 
         New users should use this API until v2 has stabilized.
```
coretl, line 19: I think we should remove all v1 references here too

docs/index.rst
```diff
@@ -8,26 +8,17 @@ How the documentation is structured
...
-    .. grid-item-card:: :material-regular:`person_add;4em`
-        :link: user_v2/index
-        :link-type: doc
-
-        The User Guide v2 contains documentation on how to use ophyd's provisional v2 API.
-
-        Developers may use this to develop v2 Devices against, but the API is subject to change
-
 .. toctree::
     :hidden:
```
coretl, line 23: Should we put a note at the bottom that points to `ophyd-async` somewhere here?

rosesyrett: Yep, good idea

rosesyrett:
tests are failing because of this issue: https://github.com/bluesky/ophyd/issues/1163

mrakitin (approved):
Looks good to me!

Issue #1163, opened 2023-10-06
setuptools issue


rosesyrett:
Seems like CI on master is failing, because of a problem with setuptools and how the repository is configured.

This seems like an urgent issue. below is a snippet of the error message being reported:
```
../../../../../opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/setuptools_scm/_integration/setuptools.py:30
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/setuptools_scm/_integration/setuptools.py:30: RuntimeWarning: 
  ERROR: setuptools==56.0.0 is used in combination with setuptools_scm>=8.x
  
  Your build configuration is incomplete and previously worked by accident!
  setuptools_scm requires setuptools>=61
  
  Suggested workaround if applicable:
   - migrating from the deprecated setup_requires mechanism to pep517/518
     and using a pyproject.toml to declare build dependencies
     which are reliably pre-installed before running the build tools
```

PR #1167, opened 2023-10-12
Temporary fix for #1163

coretl:
The long term fix is to upgrade to pyproject.toml, but this will quell the warnings for now


coretl:
Fixes #1163

prjemian (approved):
Looks good, as stated before upgrading to latest packaging recommendations.

PR #1174, opened 2024-01-09
MNT: enforce that reg_root is absolute

tacaswell:


tacaswell:
The tests should check the message on the exception, but working on a train that is almost at my station so pushing up as draft.

danielballan (approved):

PR #1177, opened 2024-02-15
Add two new kwargs to ADTriggerStatus __init__, to allow for using different signals than default AD

jwlodek:
Currently `ADTriggerStatus` essentially hard codes the two signals that are watched/compared when reporting the progress of a trigger execution. I know of at least a few cases where this isn't desirable due to either non-standard AD driver behavior, or because the underlying detector relies on a different progress metric than completed/readout frames.

I'd like to get some feedback on whether or not this approach is reasonable, and if yes I can write up a unit test or two to check behavior when either or both signals are overridden.

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -27,18 +27,29 @@ class ADTriggerStatus(DeviceStatus):
     based on comparing device.cam.array_counter to  device.cam.num_images.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, counter_signal = None, target_signal = None, **kwargs):
         super().__init__(*args, **kwargs)
         self.start_ts = ttime.time()
+
+        # Provide option to override which signals are watched to specify a "done" condition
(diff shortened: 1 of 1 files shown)


tacaswell:
Seems reasonable, but needs updates to docstring please.

jwlodek:
Hm. Not sure why those commits are now included here since they are already merged with master. Let me clean this up.

tacaswell:
I think you did a rebase, merge from your remote, push cycle.  I rebased and force-pushed for you.

ophyd/areadetector/trigger_mixins.py
```diff
@@ -24,29 +24,45 @@ class ADTriggerStatus(DeviceStatus):
     A Status for AreaDetector triggers
 
     A special status object that notifies watches (progress bars)
-    based on comparing device.cam.array_counter to  device.cam.num_images.
+    based on comparing the two signal values counter_signal and
```
tacaswell, line 27: Can this be a numpydoc style docstring?

Issue #844, opened 2020-04-16
Discuss: Issue a warning if a Status' default timeout is None?


danielballan:
There is no universal good default value for timeout. If a timeout is too large, it delays notification of failure and [leaves resources dangling](https://github.com/bluesky/ophyd/pull/837#discussion_r409007458). If a timeout is too short, it could needlessly kill a long task and force a costly restart.

I believe that we used to use default finite timeouts in some places, but after a couple incidents of long operations wrongly interrupted, we have (mostly?) stripped them out of the ophyd codebase, defaulting instead to `timeout=None`, which we take to mean "infinite". Pervasive use of infinite timeouts is not ideal. Perhaps we should discourage it by issuing a warning to the effect:

```
The Status {self!r} has been created with timeout=None.
This means that it will wait forever to finish.
It is better to set a specific maximum time that it is allowed to complete
so that it will fail with a clear error beyond that time.

If you really want to wait forever,
use math.inf instead of None to suppress this warning.
```

When can users choose a timeout?

1. Users can choose it in the class definition, as in

```py
class EpicsMotor(Device):
    def set(self, ...):
        status = DeviceStatus(self, timeout=SOME_HARD_CODED_TIMEOUT)
        ...
        return status
```

For classes like `EpicsMotor` that can correspond to a wide range of actual hardware configured in various ways, this is too blunt an approach. Some actions with an `EpicsMotor` are much longer than others.

2. We can choose it at instantiation time, as in

```py
class EpicsMotor(Device):
    def __init__(self, ..., timeout, ...):
        self.timeout = timeout

    def set(self, ...):
        status = DeviceStatus(self, timeout=self.timeout)
        ...
        return status

a_particular_motor = EpicsMotor('...', timeout=A_PARTICULAR_TIMEOUT)
```

That may be good enough if the motor in question performs movements with a well-defined maximum possible time to complete.

3. We can choose a timeout per a particular action. This is when we have the most information about how long we should expect it to take. For example, for a motor, we can take into account the current velocity setting, the current position, and the distance to the target position. We don't _have_ to take this into account, but we could.

```py
class ParticularEpicsMotor(EpicsMotor):
    def set(self, ...):
        timeout = decision_based_on_how_far_we_have_to_go(...)
        status = DeviceStatus(self, timeout=timeout)
        ...
        return status
```

[...]

PR #1182, opened 2024-03-19
status: do not create thread to run callbacks if timeout is None

mguijarr:
Here is a merge request that mitigates concern expressed here: https://github.com/bluesky/ophyd/pull/837#discussion_r409007458 (also linked with issue #844).

I have the impression, a simple optimization is to not create a thread if it is not needed (in particular when timeout is set to `None`, which seems to be most of the time).

Otherwise some background about the problem `Status` objects want to solve with the callbacks thread would be insightful.

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -114,10 +114,13 @@ class StatusBase:
                 DeprecationWarning,
             )
 
-        self._callback_thread = threading.Thread(
-            target=self._run_callbacks, daemon=True, name=self._tname
-        )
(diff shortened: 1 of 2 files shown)


tacaswell:
Thank you for this,  this seems sensible to me and looks correct to me.

Can you add a test in `test_status.py` to cover the case with no timeout but a settle time?

mguijarr:
@tacaswell Done

tacaswell (approved):

PR #1185, opened 2024-04-08
Fix for issue 1184

mguijarr:
Following comment by @cappel89 , this MR proposes to define `PositionerBase` as an abstract class.

This way, it would not be needed to mention `move()` method needs to be overwritten in a specific way, while still
having a default implementation that can be called in subclasses (must be done explicitely). Same with `.egu` property,
no need to ask developer to overwrite it -> it would be done automatically by Python...

What do you think ?


tacaswell:
If I recall correctly, when we started working on ophyd we were using a metaclass for Device and multi-inherentance involving multiple-metaclasses results in Python telling you it is too complicated and refusing to do it.  I think you could do it by making a MI meta-class of the meta-classes and then using that as the meta-class on the normally-MI'd class or something like that....so we chose to not use the second meta-class!  When the subclass init hook became available we switch to using that (in the spirit of using the simplest language features we needed to get what we wanted) so I think that is why this now works.

tacaswell (approved):

ophyd/positioner.py
```diff
@@ -152,7 +147,8 @@ def low_limit(self):
     def high_limit(self):
         return self.limits[1]
 
-    def move(self, position, moved_cb=None, timeout=None):
+    @abstractmethod
+    def move(self, position, wait=False, moved_cb=None, timeout=None):
```
tacaswell, line 151: I'm not sure what to do about the argument order here.

Adding `wait` as the second positional arguement matches the order of `EpicmMotor.move` and `PseudoPositioner.move` but doing so is a major API breakage on this method (as if anyone was passing `moved_cb` positionally then they would both not get there callback run but wait for the motion to finish!

mguijarr: I had the same dilemma. I thought already implemented `move` show how everybody is supposed to have it.

What about adding `*` after `position` to force use of keyword args ? 

I agree it is an API change. On the other hand current situation is not ideal neither. 

Another possibility is to move `wait` argument at the end...

Issue #1193, opened 2024-05-26
Allow to specify dtype and shape without type inference for Signal description


mguijarr:
(similar issue to #1178)

The signal description (`.describe()`) method guesses the `dtype` and `shape` of a `Signal` value,
from the default value passed to `Signal` constructor.

This leads to a type mismatch, rendering the implementation insufficiently robust. Initializing a
value with 0 does not inherently indicate that the signal is an integer; more frequently, the signal
is a float.
The `describe()` function would be better useful to get real information about the signal,
when considering who or what is interpreting the data—whether it be a human or a process
recording Ophyd data for example. The differentiation between numpy.uint32, numpy.int64, and
a generic integer is significant and must be maintained to ensure data validation.

mguijarr:
Done! (feature merged https://github.com/bluesky/ophyd/pull/1194 )

Issue #1178, opened 2024-02-16
Option to force "number" rather than "integer" in Signal


cjtitus:
Summary
--------------

There are many signals where the dtype is known a-priori to be floats. However, Ophyd always looks at an example value when `describe` is called, and shoves that into a `data_type` function.

If the first value happens to be an integer, the `describe` method then returns integer, and all further data will be cast to an integer, even if everything else is floats. I think the true data is saved, somehow, but have never really managed to get it back out of Tiled.

There should be a way to tell a signal to treat numeric dtypes as `number` and not `integer`, so that this can be set properly on a signal-by-signal basis.

mguijarr:
Thanks to !1194, it is now possible to do add `dtype=float` to `Signal` arguments to force `number` instead of `integer`:

```
>>> from ophyd.signal import Signal
>>> s=Signal(name="mysignal", value=0, dtype=float)
>>> s.describe()
{'mysignal': {'source': 'SIM:mysignal', 'dtype': 'number', 'shape': [], 'dtype_numpy': 'float64'}}

```

PR #1194, opened 2024-05-26
signal: init value can be a ValueInfo structure with dtype, shape and…

mguijarr:
… default value to characterize the value

Solves issue #1193 (and #1178 ?)

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -38,8 +38,16 @@ DEFAULT_CONNECTION_TIMEOUT = object()
 DEFAULT_TIMEOUT = object()
 DEFAULT_WRITE_TIMEOUT = object()
 
+
 # Sentinel to identify if we have never turned the crank on updating a PV
-DEFAULT_EPICSSIGNAL_VALUE = object()
+class _unset_value_sentinel:
+    def __repr__(self):
+        return "UNSET_VALUE"
+
+
+UNSET_VALUE = _unset_value_sentinel()
+# for backward compatibility: before it was called DEFAULT_EPICSSIGNAL_VALUE
+DEFAULT_EPICSSIGNAL_VALUE = UNSET_VALUE
 
 
 class ReadTimeoutError(TimeoutError):
@@ -50,13 +58,31 @@ class ConnectionTimeoutError(TimeoutError):
     ...
 
 
+def check_dtype(value_array, dtype):
+    try:
+        value_array.astype(dtype, casting="same_kind")
+    except TypeError:
+        # check if conversion between signed int to unsigned int would be fine
+        if value_array.dtype.kind == "i" and np.dtype(dtype).kind == "u":
+            bounds = np.iinfo(dtype)
+            if np.all(value_array >= bounds.min) and np.all(value_array <= bounds.max):
+                return True
+        return False
+    else:
+        return True
+
+
+class _DefaultFloat(float):
+    pass
+
+
 class Signal(OphydObject):
     r"""A signal, which can have a read-write or read-only value.
 
     Parameters
     ----------
     name : string, keyword only
-    value : any, optional
+    value : any acceptable value, optional (default: UNSET_VALUE)
         The initial value
     kind : a member the Kind IntEnum (or equivalent integer), optional
         Default is Kind.normal. See Kind for options.
@@ -85,6 +111,7 @@ class Signal(OphydObject):
     rtolerance : any, optional
         The relative tolerance associated with the value
     """
+
     SUB_VALUE = "value"
     SUB_META = "meta"
     _default_sub = SUB_VALUE
@@ -95,7 +122,9 @@ class Signal(OphydObject):
         self,
         *,
         name,
-        value=0.0,
+        value=_DefaultFloat(0.0),
+        dtype=None,
+        shape=None,
         timestamp=None,
         parent=None,
         labels=None,
@@ -116,6 +145,41 @@ class Signal(OphydObject):
         self.cl = cl
         self._dispatcher = cl.get_dispatcher()
         self._metadata_thread_ctx = self._dispatcher.get_thread_context("monitor")
+
+        # check if value corresponds to specified info (or if it is compatible)
+        if dtype == "string":
+            self._value_dtype_str = ""
+            self._value_shape = ()
+            if isinstance(value, _DefaultFloat):
+                # a specific default value was not passed to constructor
+                value = UNSET_VALUE
+        elif dtype is None:
+            self._value_dtype_str = ""
+            self._value_shape = shape
+        else:
+            self._value_dtype_str = np.dtype(dtype).name
+            self._value_shape = shape
+
+            if value is not UNSET_VALUE:
+                value_array = np.asanyarray(value)
+
+                if not check_dtype(value_array, dtype):
+                    if isinstance(value, _DefaultFloat):
+                        # a specific default value was not passed to constructor ;
+                        # change to UNSET_VALUE
+                        # (value will be read via .get() - it does not mean it will correspond
+                        # to the desired dtype, but at least it is more coherent like this...)
+                        value = UNSET_VALUE
+                    else:
+                        raise TypeError(
+                            f"The value {value} does not match the required dtype {dtype}."
+                        )
+        if shape is not None and value is not UNSET_VALUE:
+            if np.asanyarray(value).shape != shape:
+                raise TypeError(
+                    f"The value {value} does not have the required shape {shape}."
+                )
+
         self._readback = value
 
         if timestamp is None:
@@ -152,6 +216,10 @@ class Signal(OphydObject):
 
             self._metadata.update(**unset_metadata)
 
+    @property
+    def source_name(self):
+        return "SIM:{}".format(self.name)
+
     def trigger(self):
         """Call that is used by bluesky prior to read()"""
         # NOTE: this is a no-op that exists here for bluesky purposes
@@ -186,12 +254,9 @@ class Signal(OphydObject):
     def _repr_info(self):
         "Yields pairs of (key, value) to generate the Signal repr"
         yield from super()._repr_info()
-        try:
-            value = self._readback
-        except Exception:
-            value = None
 
-        if value is not DEFAULT_EPICSSIGNAL_VALUE:
+        value = self._readback
+        if value is not UNSET_VALUE:
             yield ("value", value)
 
         yield ("timestamp", self._metadata["timestamp"])
@@ -206,6 +271,8 @@ class Signal(OphydObject):
 
     def get(self, **kwargs):
         """The readback value"""
+        if self._readback is UNSET_VALUE:
+            raise RuntimeError("Signal value has never been read yet")
         return self._readback
 
     def put(
@@ -408,7 +475,7 @@ class Signal(OphydObject):
             "This behavior will likely change in the future."
         )
 
-        if self._readback is DEFAULT_EPICSSIGNAL_VALUE:
+        if self._readback is UNSET_VALUE:
             # If we are here, then we have never turned the crank on this Signal.  The current
             # behavior is to fallback to poking the control system to get the value, however this
             # is problematic and we may want to change in the future so warn verbosely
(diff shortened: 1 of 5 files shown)


mguijarr:
Please comment , if you agree with the proposed change I will add a test.

ophyd/signal.py
```diff
@@ -95,7 +102,7 @@ def __init__(
         self,
         *,
         name,
-        value=0.0,
+        value=DEFAULT_EPICSSIGNAL_VALUE,
```
tacaswell, line 105: If we move this here I suspect we will have to hoist the logic to handle this from EpicsSignal up to the base class.

tacaswell:
Interesting idea!

I have some concerns that this will box us out from ever fixing `EpicsSignal` to trust the metadata that comes across on the channel rather than infering the dtype from the values, but I think that if the rules go 1. user passed 2. Channel provided 3. inferred would be ok.

We have been using `dtype_str` for providing higher-fidelitiy data type information (see discussion in https://github.com/bluesky/event-model/pull/215 despite not getting that PR merged, we have started to use it in-practice in tiled) so we probably should use a different name internally.

I lean +.75 on this being a good idea .

ophyd/signal.py
mguijarr: well, it looks like `DEFAULT_EPICSSIGNAL_VALUE` is more used as a sentinel value to indicate if a default has been provided or not - it is currently already used in `Signal` class, see line 475:

``` 
if self._readback is DEFAULT_EPICSSIGNAL_VALUE:
            val = self.get()
```

or derived classes which do not have an explicit relation with EPICS (i.e not deriving from `EpicsSignalBase`).

I propose to rename it, then ? Like `UNSET_VALUE` or `UNSPECIFIED_VALUE` ?

It seems `None` is considered as a potential ok value. Is it the case ? Otherwise we could initialize with `None`.

In `describe()`, I think the key point is to really read the value if it is not specified. Any way to do it like this is fine for me.

What do you think ?

mguijarr:
> I have some concerns that this will box us out from ever fixing `EpicsSignal` to trust the metadata that comes across on the channel rather than infering the dtype from the values, but I think that if the rules go 1. user passed 2. Channel provided 3. inferred would be ok.

I agree with this.

> We have been using `dtype_str` for providing higher-fidelitiy data type information (see discussion in [bluesky/event-model#215](https://github.com/bluesky/event-model/pull/215) despite not getting that PR merged, we have started to use it in-practice in tiled) so we probably should use a different name internally.

Ok I renamed shape and dtype vars to `_value_shape` and `_value_dtype_str`.

ophyd/signal.py
tacaswell: Oh, if we are already using it is the base class than I'm 50/50 on adding an alias (the hassle of anyone downstream using the old name needing to update is not worth it to remove the old name) vs just using it.

The concern I still have is that in `Signal.get` we just return `self._readback` which means that there is a chance of returning `DEFAULT_EPICSSIGNAL_VALUE` to the user (which we definitely should not do).  The minimal fix is probably to raise in `Signal.get` if the value has never been set.

mguijarr: Wouldn't it be better to initialize to `None` and to return `None` if it has not been set? I am very new to Ophyd, so I do not know the logic/philosophy behind reading signals and why `DEFAULT_EPICSSIGNAL_VALUE` was there at first place

mguijarr: Otherwise I can make an alias and raise in `get()` as you suggest, no problem 👍🏻 

Tell me what you prefer

tacaswell: tl;dr:  I would prefer the alias + raise in `get` path as the minimal change.  

pyepics (the default wrapping we use to talk channel access over in Python) returns `None` when it fails so defaulting to `None` means we could not tell the difference between "failed to read" vs "never read".  The logic was to have a unique sentinel that would never come to use from pyepics and would never be passed to us by the user so we could tell if we had ever been told a value for this signal. Basically detect the start up transient and allows lazy connection (each PV is its own connection dance with a server and if we never read it then it is better to never try to connect).

Going with a custom default object also helps makes clear that it is "default but invalid" where as `None` is frequently "default and valid" or "placeholder to tell the system to do the default thing" and I think that is an important distinction here.

I'm a bit fuzzy remember choices we made probably 8 years ago now, but I am biased to trusted my previous self 🤣

prjemian:
The value of `DEFAULT_EPICSSIGNAL_VALUE`  should be set such that (lazy) programmatic use does not rely on the actual value of the default.  Random value should be assigned to require use of the symbol, such as one of these:

```py
DEFAULT_EPICSSIGNAL_VALUE = object()
DEFAULT_EPICSSIGNAL_VALUE = uuid.uuid4()
```

mguijarr:
Hi @tacaswell , @prjemian 😄 This is ready for me. I don't know why doc would not build... ?

I added specific tests for the feature. I renamed `ValueInfo` to `NumericValueInfo` because it is not adapted to strings
for example. I added `source_name` as a property, in order to make `describe()` more generic. I made that `shape` in `describe()` is a tuple rather than sometimes a list, sometimes a tuple (because, for numpy arrays, it is a tuple normally). 

I would be happy to work on having `EpicsSignalBase` trying to specify dtype and shape from metadata but I have no idea how to do it since I am very new to EPICS, PyEpics in general. Normally it is easy to do it like @tacaswell proposes, 1. user defined , 2. EPICS metadata or 3. inference if there is nothing else... Maybe in another MR ?

mguijarr:
No news ; is there anything wrong with the proposal?

prjemian:
I'm OK with the part on which I commented.  I'll leave it to Tom for the PR review.

wakonig:
Hi @tacaswell! What's the status on this one? It would help us a lot if we could get this one merged soon. Thanks!

mguijarr:
Hi again... I am reviving this discussion, I really need the added functionality so can we find an agreement on this?

tacaswell:
Sorry I never found you again at NOBUGs!

ophyd/signal.py
```diff
@@ -457,6 +481,30 @@ def read(self):
         value = self.get()
         return {self.name: {"value": value, "timestamp": self.timestamp}}
 
+    def _infer_value_kind(
+        self, kind, inference_funcs={"dtype": data_type, "shape": data_shape}
```
tacaswell, line 485: rather than a string + a mutable default + a dict lookup would could we just pass the inference function?

ophyd/signal.py
```diff
@@ -95,7 +104,7 @@ def __init__(
         self,
         *,
         name,
-        value=0.0,
+        value=NumericValueInfo(float, (), 0.0),
```
tacaswell, line 107: Rather than merging the expected info and the default value, can we add a new key(s) of `expected_dtype_and_shape` or `dtype` and `shape`?  In `__init__` it seems reasonable to do a cast, but in `describe` I wonder if we should be checking if the value we have in had is actually the correct dtype/shape or not?

ophyd/tests/test_signal.py
```diff
@@ -695,3 +695,71 @@ def test_import_ro_signal_class():
...
+        name="motor",
+        value=NumericValueInfo(float),
+    )
+    sig_desc = sig.describe()["motor"]
+    assert sig_desc["dtype"] == "float64"
+    assert sig_desc["shape"] == ()
+
+    original = Signal(name="original")
+    original_desc = original.describe()["original"]
+    assert original_desc["dtype"] == "float64"
```
tacaswell, line 719: This is out of the schema for things that can be in descriptors: https://github.com/bluesky/event-model/blob/a1fdaec0c45b89e513d0be9dba0bfc453982e5bb/src/event_model/schemas/event_descriptor.json#L51-L62

tacaswell:
The more-complete dtype information should be put in `dtype_numpy`.  This going in everywhere at the bottom would probably be a big help (as we have been stuffing it in ad-hoc at NSLS-II).

https://github.com/bluesky/event-model/blob/a1fdaec0c45b89e513d0be9dba0bfc453982e5bb/src/event_model/schemas/event_descriptor.json#L63-L91

mguijarr:
I just pushed some code, to comply with your comments. Let's see if tests pass

ophyd/signal.py
mguijarr: 😆 sure ; done in latest version

ophyd/signal.py
mguijarr: In my original proposal I didn't want to add more keyword args, but yeah sure - I introduced `dtype` and `shape` in latest version.

For the check in describe: I didn't change anything yet

ophyd/tests/test_signal.py
mguijarr: Ok so newest version has `dtype` and `dtype_numpy`

tacaswell:
I don't have a good theory as to why we are seeing the pytest issues on this PR.  Can you try rebasing on current main?

mguijarr:
@tacaswell sorry for the delay... I struggled a bit with the changes (this is why test didn't pass last time).

In fact I asked me a lot of questions, finally I think I get it all right.

For the review: maybe the easiest is to have a look to the new test (`test_signal_dtype_shape_info`), because it shows the expected behavior. So if you agree with this, then you can maybe look at changes in details ?

What I find relevant to know:
- I added keyword arguments `dtype` and `shape` which defaults to `None`
- The default value is `0.0`, but I identify it with a `DefaultFloat` class
    - this is to distinguish if user provided a value, or left default
    - indeed, I need to know this in some circumstances like a string signal without specific default passed by the user
        - in this case, I do not initialize with `0.0` but with `UNSET_VALUE` (previously `DEFAULT_EPICSSIGNAL_VALUE`, which is still there for compatibility)
- the UNSET_VALUE has a `__repr__`, so we can see it while debugging more easily
- if a default value is passed, there is a check to validate if it corresponds to the dtype (or shape)
    - it was not so obvious to allow `numpy.int64` (default for ints) to an unsigned int for example, numpy can be tricky!
    - there is no cast, but `TypeError` is raised if it does not correspond
- I removed a weird `__copy__` dunder
    - I could not see why it was like this, no specific test neither
    - with the original `__copy__` one test was not passing, because it recreates the object with different constructor arguments
- in `describe()`, if the dtype and shape are specified it is used for determining infos instead of inferring from data
    - otherwise it is like before
- in `describe()`, if numpy dtype is specified it goes to `dtype_numpy` as you required
    - however, if it is not available the key is also not set in the dictionary (allows old tests to pass)
-  I added `.source_name` as a property, in order to make `describe()` more generic
- I changed `data_shape` to return a tuple (because, it is really a shape), however it is returned as a list by `describe()` (like now) to make it JSON compliant

mguijarr:
Other remark:

- would it be possible to get rid of `integer` for simple dtype in `describe()` ?
    - now we have `dtype_numpy` I don't see why we would need this
    - it would simplify some code

- would it make more sense to not initialize `value` to 0.0 ?
    - I would prefer `UNSET_VALUE`

tacaswell:
> would it be possible to get rid of integer for simple dtype in describe() ?

Maybe we could get the RE to look at `numpy_dtype` and infer the correct `dtype`, however in the document model `dtype` is required but `numpy_dytpe` is optional (https://github.com/bluesky/event-model/blob/a1fdaec0c45b89e513d0be9dba0bfc453982e5bb/src/event_model/schemas/event_descriptor.json#L145-L149) so I think it makes sense to always provide both from ophyd.

> would it make more sense to not initialize value to 0.0 ?

That seems fine as `0.0` is not a valid value for a great many dtypes!

ophyd/signal.py
```diff
@@ -1474,18 +1573,14 @@ def describe(self):
         dict
             Dictionary of name and formatted description string
         """
-        if self._readback is DEFAULT_EPICSSIGNAL_VALUE:
-            val = self.get()
-        else:
-            val = self._readback
+        desc = super().describe()[self.name]
```
tacaswell, line 1576: If `super().describe()` happens to return more than one key this will silently drop the other keys on the ground.

I don't think it _should_ do that (and if it were we would currently not even be asking the parent class to source a describe method), but it seems better to be defensive and either do

```python
ret = super().describe()
desc = ret[self.name]
...
return ret
```

or

```python
(name, desc), = super().describe().items()
if name != self.name: raise RuntimeError
```

mguijarr: Changed (using first proposal), both in `Signal.describe` and `DerivedSignal.describe`

ophyd/ophydobj.py
```diff
@@ -588,14 +588,14 @@ def _repr_info(self):
         if self._parent is not None:
             yield ("parent", self.parent.name)
 
-    def __copy__(self):
```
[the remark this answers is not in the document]
mguijarr: I said in the comment above: 

- I removed a weird __copy__ dunder
    - with the original __copy__ one test was not passing, because it recreates the object with different constructor arguments
    - I could not see why it was like this, no specific test neither
    
But, with my newest code the failing test works again. So I re-enabled `__copy__`, no prob.

(however, I still don't understand what it is for)

mguijarr:
@tacaswell Here we are. Should be ready to merge now ?

mguijarr:
> I think it makes sense to always provide both from ophyd.

Just to be precise: current code I sent, do not _always_ provide `dtype_numpy` ; indeed, the document model check do not understand empty string as a valid `dtype_numpy`. So, in this case (if dtype_numpy is unknown), it is not in the dict returned by `describe()` at all (otherwise the verification logic has to be changed)

About replacing 0.0 default value with `UNSET_VALUE`:

> That seems fine as 0.0 is not a valid value for a great many dtypes! 

Ok. I will make another MR, then. It is not urgent for me.

tacaswell (approved):

Issue #1095, opened 2023-02-24
SoftPositioner is not identified as device or signal using walk_components()


prjemian:
While exploring a resolution to https://github.com/BCDA-APS/apstools/issues/641, tested the proposed solution there to a diffractometer instance from [hklpy](https://blueskyproject.io/hklpy/ready_to_use.html#simulatede4cv), such as:

```py
import hkl

sim4c = hkl.SimulatedE4CV("", name="sim4c")
```

Using the `walk_components()` method, it is possible to print a small table showing is a Component is either a device or a signal:

```py
import pyRestTable

table = pyRestTable.Table()
table.labels = "dotted_name is_device  is_signal".split()
for item in sim4c.walk_components():
    table.addRow((item.dotted_name,item.item.is_device, item.item.is_signal))    
print(table.reST(fmt="md"))
```

This table is printed:

| dotted_name             | is_device | is_signal |
| ----------------------- | --------- | --------- |
| energy                  | False     | True      |
| energy_units            | False     | True      |
| energy_offset           | False     | True      |
| energy_update_calc_flag | False     | True      |
| geometry_name           | False     | True      |
| class_name              | False     | True      |
| sample_name             | False     | True      |
| lattice                 | False     | True      |
| lattice_reciprocal      | False     | True      |
| U                       | False     | True      |
| UB                      | False     | True      |
| reflections             | False     | True      |
| reflections_details     | False     | True      |
| ux                      | False     | True      |
| uy                      | False     | True      |
| uz                      | False     | True      |
| diffractometer_name     | False     | True      |
| _hklpy_version          | False     | True      |
| _pseudos                | False     | True      |
| _reals                  | False     | True      |
| _constraints            | False     | True      |
| _mode                   | False     | True      |
| orientation_attrs       | False     | True      |
| max_forward_iterations  | False     | True      |
| h                       | True      | False     |
| h.readback              | False     | True      |
| h.setpoint              | False     | True      |
| k                       | True      | False     |
| k.readback              | False     | True      |
| k.setpoint              | False     | True      |
| l                       | True      | False     |
| l.readback              | False     | True      |
| l.setpoint              | False     | True      |
| omega                   | False     | False     |
| chi                     | False     | False     |
| phi                     | False     | False     |
| tth                     | False     | False     |

Note that omega, chi, phi, and tth that both columns are `False`.  It is expected that each of these `SoftPositioner`s would be identified as `is_device=True`.

prjemian:
[...]

PR #1198, opened 2024-07-02
Missing destroyed attribute on OphydObject

wakonig:
## Description

While `OphydObject` implements the `destroy` method, it does not provide the `_destroyed` attribute. The latter is only implemented on `Device`. Given the "_", I assume it is not meant to be used directly, yet I'm not aware of another way to check whether an object was destroyed. In the absence of a more elegant solution, I'd suggest to add the `_destroyed` attribute to the `OphydObject` class to make available e.g. to a `SoftPositioner`. 

## Related Issues

https://github.com/bluesky/ophyd/issues/1095

## Type of Change

- Added `_destroyed` attribute to `OphydObject`
- Set `_destroyed` to True after a call to `destroyed`

### DIFF
--- a/ophyd/ophydobj.py
+++ b/ophyd/ophydobj.py
@@ -166,6 +166,7 @@ class OphydObject:
             raise ValueError("name must be a string.")
         self._name = name
         self._parent = parent
+        self._destroyed = False
 
         # dictionary of wrapped callbacks
         self._callbacks = {k: {} for k in self.subscriptions}
@@ -345,6 +346,7 @@ class OphydObject:
     def destroy(self):
         """Disconnect the object from the underlying control layer"""
         self.unsubscribe_all()
+        self._destroyed = True
 
     @property
     def parent(self):


tacaswell:
We are tracking if destroy has been called in `Device` and `Signal` to prevent resuscitating them out of an abundance of caution that we will be able to reliably get any epics callbacks and connections configured correctly again.  For soft signals / positions it is not clear to me that we need to prevent their re-use the same way we do with things talking to an underlying control system.

Why do you need to check if an object has been destroyed?

wakonig:
It is more a question of consistency: This caught me very much by surprise that although it implements the `destroy` method, it lacks the ability to check if the destroy method was called. If this should only be available for classes that truly need it (e.g. Epics-based classes), I'd guess `Device` and `Signal` shouldn't have them either and instead they should move to e.g `PVPositioner` and `EpicsSignalBase`?

tacaswell:
I think nothing has public API to check if `destroy` has been called (other than implicitly via operations failing in some cases).   It is an implementation detail that some classes keep track of this as internal state and keep them selves from being un-destroyed.

If we want to lift "have I been destroyed" up to public state and add it to the protocols, we can talk about that.  However, I am still not clear why you would check if something has been destroyed or not (it appears the only place we call it in ophyd is in the test suite where we create and discard many objects, but when we call `obj.destroy` it is on the path to dropping all references so there is nothing left to check) and adding unused private state does not seem useful to me.

Issue #1199, opened 2024-07-03
CI fails due to old container images


dperl-dls:
E.g.

```
Run source ${GITHUB_WORKSPACE}/scripts/epics_exports.sh
latest: Pulling from nsls2/epics-docker
[DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of docker.io/nsls2/epics-docker:latest to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/
Error: Process completed with exit code 1.
```

This can be temporarily overridden until docker 27.0 is used in GH actions

dperl-dls:
closed by merging #1200

Issue #1197, opened 2024-06-27
Update intersphinx inventory to use https://blueskyproject.io/bluesky/main/objects.inv


mrakitin:
Need to update the intersphinx inventory link to be https://blueskyproject.io/bluesky/main/objects.inv.

_Originally posted by @mrakitin in https://github.com/bluesky/ophyd/pull/1189#discussion_r1587863524_

PR #1200, opened 2024-07-03
Override docker version deprecation

dperl-dls:
Temporary patch for #1199 to allow CI to run
Docs failure is because of https://github.com/bluesky/ophyd/issues/1197


mrakitin (commented):
Thanks for the temporary fix, @dperl-dls!
The docs CI fails with:
```
intersphinx inventory 'https://blueskyproject.io/bluesky/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: https://blueskyproject.io/bluesky/objects.inv
```

I think we need to use https://blueskyproject.io/bluesky/main/objects.inv now. Do you mind changing the link [here](https://github.com/bluesky/ophyd/blob/1f8b64f52fcf99add0a50d67abdc69adc80b036f/docs/conf.py#L107) while you are here? Thanks!

setup.cfg
```diff
@@ -20,7 +20,7 @@ setup_requires =
 # Specify any package dependencies below.
 install_requires =
     networkx>=2.0
-    numpy
+    numpy<2.0 # See https://github.com/mdavidsaver/p4p/issues/145
```
tacaswell, line 23: Can we put this pinning in the pva section?

Now that we have split ophyd async out do we need p4p as a dep anymore?

dperl-dls: Yeah that makes sense, I'll move it there. I don't know whether you need the pva section at all anymore but do feel like it's a bit of a separate issue to this.

mrakitin (approved):
Looks good to me, thanks for the efforts, @dperl-dls!

setup.cfg
mrakitin: Opened a separate issue for removing the PVA section: https://github.com/bluesky/ophyd/issues/1201.

PR #1211, opened 2024-09-20
TYPO should be lower case to be consistent with others in this module

prjemian:
Somehow, this missed testing.  A long time ago.

```py
    tetramm = TetrAMM("8idTetra:QUAD1:", name="tetramm")
    for axis in "x y".split():
        for attr_name in "offset offset_calc scale".split():
            getattr(tetramm, f"position_{attr_name}_{axis}").kind = "config"
AttributeError: position_scale_y
```

### DIFF
--- a/ophyd/quadem.py
+++ b/ophyd/quadem.py
@@ -98,7 +98,7 @@ class QuadEM(SingleTrigger, DetectorBase):
     position_offset_calc_y = Cpt(EpicsSignal, "ComputePosOffsetY")
 
     position_scale_x = Cpt(EpicsSignal, "PositionScaleX")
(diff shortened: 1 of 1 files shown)


prjemian:
We can get by without this change (which might break existing usage) with:

```py
    for attr_name in tetramm.component_names:
        attr = getattr(tetramm, attr_name)
        if attr_name.startswith("current_"):
            for ch_name in attr.component_names:
                getattr(attr, ch_name).kind = "config"
        elif attr_name.startswith("position_"):
            attr.kind = "config"
```

But for sure, all these components need to be `kind="config"`.

danielballan (approved):

PR #1221, opened 2024-11-21
CI: do not try to start docker images that no longer exist

tacaswell:
For reasons not yet understood, these images now have 0 size on dockerhub


tacaswell:
I expect the test suite to still run, but skip a bunch of tests.

jwlodek (approved):

PR #1222, opened 2024-11-25
Enh/control child name mangling

tacaswell:
The Python name (via attributes) of a child device is `foo.bar.baz`, however
for a number of historical reasons (allowed keys in mongo and for attribute
access in Pandas data frames with columns named from the data keys in an event
stream) ophyd-sync has replaced '.' with '_' in the names.  This leads to
ambiguity when looking at the (default) name to which component on which device
the data came from as there may be `_` in the attribute names.  Ophyd-async
used '-' which does allow for round tripping.

This makes it controllable on per-instance basis in ophyd-sync so that users
can either make a different set of trade offs.

xref bluesky/ophyd-async#666

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -234,7 +234,7 @@ class Component(typing.Generic[K]):
             suffix from.
 
         kw : str
-            The key of associated with the suffix.  If this key is
+            The key of associated with the suffix.  If this key is in
             self.add_prefix than prepend the prefix to the suffix and
             return, else just return the suffix.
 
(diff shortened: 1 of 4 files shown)


tacaswell:
I probably should move the doc commit to its own PR...

ophyd/device.py
```diff
@@ -839,10 +841,11 @@ def __init__(
         read_attrs=None,
         configuration_attrs=None,
         parent=None,
+        child_name_seperator="_",
```
coretl, line 844:
```suggestion
        child_name_separator="_",
```

mrakitin (approved):
Looks like a good approach to configure the separator.

PR #1228, opened 2025-01-13
Redis obj

tacaswell:
This is a rebase of #904 but because that was opened from the master branch of the source repo I do not seem to be allowed to push to it.

### DIFF
--- a/ophyd/redis_signal.py
+++ b/ophyd/redis_signal.py
@@ -17,13 +17,13 @@ class NoEventNotifications(EnvironmentError):
 
 
 class RedisSignal(OphydObject):
-    '''Redis backed Ophyd Signal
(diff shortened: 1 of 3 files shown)


ophyd/redis_signal.py
```diff
@@ -0,0 +1,306 @@
...
+        return {}
+
+    @property
+    def hints(self):
+        if self.kind == Kind.hinted:
+            return {"fields": [self.name]}
+        else:
+            return {}
+
+    def subscribe(self, *args, **kwargs):
```
tacaswell, line 175: This should probably only work on `SUB_VALUE` ?

tacaswell: This also needs to thread the user's callback through so the users code gets called when the values update.

PR #1231, opened 2025-01-28
FIX: do not leak the book-keeping float sub-class out of __init__

tacaswell:
Not all serializers (e.g. orjson) will fallback to the float behavior.

### DIFF
--- a/ophyd/tests/test_signal.py
+++ b/ophyd/tests/test_signal.py
@@ -803,3 +803,8 @@ def test_signal_dtype_shape_info(fake_motor_ioc, cleanup):
         name="ok_value", dtype="int64", shape=(2, 2), value=[[1, 2], [3, 4]]
     )
     assert ok_default_shape._value_shape == (2, 2)
+
+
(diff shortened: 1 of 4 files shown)


ophyd/tests/test_signal.py
```diff
@@ -803,3 +803,8 @@ def inverse(self, value):
         name="ok_value", dtype="int64", shape=(2, 2), value=[[1, 2], [3, 4]]
     )
     assert ok_default_shape._value_shape == (2, 2)
+
+
+def test_signal_default_type():
+    s = Signal('aardvark')
```
mrakitin, line 809:
```suggestion
    s = Signal(name="aardvark")
```

mrakitin: That should fix the failing tests.

mrakitin (approved):

tacaswell:
I squashed this all to 2 commits.

mguijarr (approved):
Looks good to me.

PR #1235, opened 2025-03-07
Update status.rst SubscriptionStatus example

whs92:
When you use subscription status you almost never want the subscription to run immediately when you instantiate it. The docs have caused us a number of problems with people copying the example and seeing multiple values in their readings when a trigger executes at a time they don't expect. 

I changed the example to fit what is commonly done (in my experience)


jwlodek:
Maybe the `run` kwarg should default to `False`?

whs92:
> Maybe the `run` kwarg should default to `False`?

I figured that might break people's deployed code

PR #1236, opened 2025-03-28
Added ability to configure wait_for_connection timeout

thomashopkins32:
In trying to set up integration tests for a beamline profile with many connections, I ran into a timeout error waiting for a device to connect. The default of 2 seconds is too low so I bumped it up to 10 seconds.

I also added the ability to configure this for individual `Component`s that may have many sub-`Component`s that try to connect at the same time.

The timeout occurred [here](https://github.com/bluesky/ophyd/blob/main/ophyd/quadem.py#L119-L121). It was running on a single locally hosted IOC that [accepts all connections and returns 0 for everything](https://github.com/NSLS2/gha-beamline-integration-test/blob/2025-1.0-test/spoof_beamline.py).

### DIFF
--- a/ophyd/device.py
+++ b/ophyd/device.py
@@ -172,6 +172,8 @@ class Component(typing.Generic[K]):
     add_prefix: Tuple[str, ...]
     #: Subscription name -> subscriptions marked by decorator.
     _subscriptions: DefaultDict[str, List[Callable]]
+    # timeout for wait_for_connection
+    connection_timeout: float = 10.0
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


mrakitin:
Can you please rebase on `main` to pick up the CI changes from #1233?

prjemian:
Me, too. We tend to set an overall default connection timeout on the
slowest devices which are too be connected on a routine basis. The ability
to fine-tune the connection timeout for devices with many children (area
detectors, scalers, multiple instances of EPICS calc records and their
variants, ...) is an improvement.

JunAishima:
been having the same issue, even when running on my laptop. This PR and correctly configured EPICS_CA environment variables fixed it!

mrakitin (approved):

PR #1240, opened 2025-04-04
Continuous acquisition trigger mixin with circular buffer

thopkins32:
## Overview

Adds a continuous acquisition trigger mixin class that allows for easy integration with area detectors cameras.

## Motivation

Area detectors typically can be run in a few modes: single, multiple, or continuous acquisition. In single acquisition mode, detectors capture a single image. In multiple mode, detectors capture multiple images. Acquisition stops once the desired number of images is reached. This behavior is covered in Ophyd by `TriggerBase` (and relevant subclasses). Continuous acquisition mode captures images continuously, never terminating the underlying control system's acquisition thread.

There are a few benefits to running the camera in continuous acquisition mode:
- Camera temperature remains stable
- Added flexibility for software-based triggers
- Avoids noise in the first few captured frame in some cameras
- Reduces noise in between frame timestamps introduced by acquisition start/stop
- The camera captures frames asynchronously from downstream tasks (like computing statistics, writing images to disk, etc.)

## How it works

This class utilizes a subset of the functionality provided by the [`NDPluginCircularBuff`](https://millenia.cars.aps.anl.gov/software/epics/NDPluginCircularBuff.html) EPICS area detector plugin to buffer frames for a software trigger. The size of the buffer,  the "pre-count", is set to `0` in our implementation to ensure that the most recent frame is the one that is always captured. Triggering the plugin releases from the buffer a number of frames specified by the "post count".

The `NDCircularBuffTriggerStatus` tracks the "post-count quantity" and compares it to the target "post-count".

## How it was tested

This was tested using ADSim. I set up the plugins as:
```
DET -> CB -> HDF5
```
so that we can test file writing from triggering the circular buffer.

- [x] Unit tests added

## Notes

I think there may be a bug with the `NDPluginCircularBuff` plugin implementation when setting the "pre-count" > 0. When used in tandem with the HDF5 plugin, it was always writing one more frame than desired which was strange to me.

## Other changes

Refactors the `ADTriggerStatus` to utilize a base class `TriggerStatus` as to not duplicate code.

### DIFF
--- a/ophyd/areadetector/trigger_mixins.py
+++ b/ophyd/areadetector/trigger_mixins.py
@@ -365,14 +365,10 @@ class ContinuousAcquisitionTrigger(BlueskyInterface):
         # Order of operations is important here.
         self.stage_sigs.update(
             [
-                ("cam.acquire", 1),  # Start acquiring
                 ("cam.image_mode", self.cam.ImageMode.CONTINUOUS),  # 'Continuous' mode
+                ("cam.acquire", 1),  # Start acquiring
                 ("cb.flush_on_soft_trigger", 0),  # Flush the buffer on new image
                 ("cb.preset_trigger_count", 0),  # Keep the buffer capturing forever
-                # TODO: Figure out why this leaks an extra frame
-                # Tested this with the HDF5 plugin and it writes an extra frame to
-                # the file when `pre_count` is non-zero.
-                # Possibly a bug in the NDCircularBuff plugin?
                 ("cb.pre_count", 0),  # The number of frames to take before the trigger
                 ("cb.capture", 1),  # Start filling the buffer
             ]


setup.cfg
```diff
@@ -31,7 +31,7 @@ dev =
     attrs>=19.3.0
     black==22.3.0
     bluesky>=1.11.0
-    caproto[standard] >=0.4.2rc1
+    caproto[standard] >=0.4.2rc1,<1.2.0
```
tacaswell, line 34:
```suggestion
    caproto[standard] >=0.4.2rc1,!=1.2.0
```

excluding a specific version to avoid a bug is safer as if we forget to remove it won't prevent fixed versions from working (and we may want to avoid 1.12.0 forever).

thomashopkins32: PR with this change here so we can merge sooner: https://github.com/bluesky/ophyd/pull/1241

thomashopkins32:
The ADSim tests will fail until a new image is released that has the NDCircularBuff plugin configured correctly. This depends on [github.com/epics-containers/ioc-adsimdetector](https://github.com/epics-containers/ioc-adsimdetector) having their ibek-support bumped to https://github.com/epics-containers/ibek-support/commit/5fc52f9185aba48e867acddcb62cd34cc7c1cb5c

thomashopkins32:
See https://github.com/NSLS2/csx-profile-collection/blob/a80629650f595d04d9ff7dae35deb2a873403d21/startup/csx1/devices/areadetector.py#L670-L687 for a version of this integrated for an actual area detector at CSX.

tacaswell:
Merging on the strength of "used on floor".

PR #1246, opened 2025-05-05
TST: use newly agreed on convention of always including num frames

tacaswell:


danielballan:
Fixing something that has been broken for 10 years has risks—as @tacaswell knows better than I.

Let's set this aside during the current batch of changes to Tiled / `TiledWriter` / `databroker` / `bluesky-tiled-plugins`. Could revisit.

Issue #1232, opened 2025-02-03
Correct return types to match current bluesky protocols


evalott100:
Ophyd still uses older and less specific return types for some of the bluesky protocols.

``` python
class SomeDevice(Device):
    readable_sig = Component(SignalRO, "prefix")


def read_stuff(signal: Readable):
    pass


my_device = SomeDevice("prefix", name="name")

read_stuff(my_device.readable_sig)
```

gives type errors

```
Argument of type "SignalRO" cannot be assigned to parameter "signal" of type "Readable" in function "read_stuff"
  "SignalRO" is incompatible with protocol "Readable"
    "read" is an incompatible type
      Type "() -> dict[str | Unknown, dict[str, float | Unknown]]" cannot be assigned to type "() -> SyncOrAsync[Dict[str, Reading[Unknown]]]"
        Function return type "dict[str | Unknown, dict[str, float | Unknown]]" is incompatible with type "SyncOrAsync[Dict[str, Reading[Unknown]]]"
          Type "dict[str | Unknown, dict[str, float | Unknown]]" cannot be assigned to type "SyncOrAsync[Dict[str, Reading[Unknown]]]"
    "describe" is an incompatible type
      Type "() -> dict[str | Unknown, dict[str, Unknown]]" cannot be assigned to type "() -> SyncOrAsync[Dict[str, DataKey]]"
        Function return type "dict[str | Unknown, dict[str, Unknown]]" is incompatible with type "SyncOrAsync[Dict[str, DataKey]]"
```

Correcting 

https://github.com/bluesky/ophyd/blob/1e5f03752f6a6b36ae5738ebbdf4f7645129b20d/ophyd/device.py#L563

and

https://github.com/bluesky/ophyd/blob/1e5f03752f6a6b36ae5738ebbdf4f7645129b20d/ophyd/device.py#L590

to match

https://github.com/bluesky/bluesky/blob/88c403ffbfb539edfb74240703caca6c228480b7/src/bluesky/protocols.py#L269

and

https://github.com/bluesky/bluesky/blob/88c403ffbfb539edfb74240703caca6c228480b7/src/bluesky/protocols.py#L287

would fix the above.

There may be other places, e.g

https://github.com/bluesky/ophyd/blob/1e5f03752f6a6b36ae5738ebbdf4f7645129b20d/ophyd/flyers.py#L77

and

[...]

PR #1248, opened 2025-05-14
Correct return types to match current bluesky protocols

NoxHarmonium:
As per https://github.com/bluesky/ophyd/issues/1232 we have noticed that a lot of Ophyd objects such as `Signal` flag a lot of issues with type checkers when you use them with Bluesky plans, but they work fine at runtime.

This is due to a mismatch between the Ophyd types and the Bluesky protocols (https://github.com/bluesky/bluesky/blob/5f5d2c1db8ee5c52fbc9a5f9e4fb3513b2d8625e/src/bluesky/protocols.py).

This is my attempt at updating the types in Ophyd to keep the type checker happy. Sadly in a lot of the cases the fix was to make the types looser, rather than stricter. This is because we can't bring the Bluesky/event-model types into Ophyd (at least as far as I know, maybe we can?) so we would have to copy/paste heaps of types around and they might get out of sync. The bluesky types are only available at the moment as dev dependencies.

There were tests that were designed to make sure that the Ophyd objects remained compatible with the Bluesky protocols, but somewhere along the line they stopped working so I've given them an uplift and added pyright as a typechecker option since it is pretty popular now being part of VSCode via pylance.

There was one protocol (`Subscribable`) that I couldn't get Ophyd to conform too and that I'll need some extra guidance on (see the PR comments)

I have tested it with one of our beamline libraries, and it reduced the number of type errors we were flagging from 133 -> 0 :tada: . However, I am still a bit worried about changing the public API of base Ophyd objects like Device and Signal as people may be overriding them in different ways. I'd love your opinion on that.

I'll try and annotate the PR to help with the review.

Let me know what you think!

Before:
Before Screenshot
After:
After Screenshot


ophyd/__init__.py
```diff
@@ -1,5 +1,3 @@
-# type: ignore
```
NoxHarmonium, line 1: I'm not sure why this type ignore statement was here, as well as the one at the top of `device.py`.
It was preventing mypy from traversing the submodules properly so I removed it. It didn't seem to break anything in the build, or when I imported it in to our beamline library so maybe its OK?

ophyd/device.py
```diff
@@ -86,10 +86,6 @@
 DEFAULT_CONNECTION_TIMEOUT = object()
 
 
-class OrderedDictType(Dict[A, B]):
```
NoxHarmonium, line 89: I _think_ this was here for compatibility with Python < 3.8
IIRC, it was the source of most of the type issues.
Since 3.8 is the minimum target now, I think we can just replace it with `TypedDict`

ophyd/tests/test_bluesky_protocols.py
```diff
@@ -1,93 +1,157 @@
...
 
 
-def test_subscribable():
+# TODO: Ophyd signature is incompatible with bluesky protocol
+# (extra parameters, returns int instead of None, different parameter names)
+# Pyright is stricter and picks this up. Disabled for now
+@pytest.mark.skip()
+def test_subscribable(run):
     cmd = imports + "from bluesky.protocols import Subscribable;"
-    run(cmd + "foo: Subscribable = hw.signal1")
```
NoxHarmonium, line 76: `signal1` doesn't seem to exist. It may have been removed at some point (or maybe it never existed) but the type checker wasn't picking it up because `--disallow-any-unimported` wasn't on

ophyd/tests/test_bluesky_protocols.py
```diff
@@ -1,93 +1,157 @@
...
+    run(cmd + "foo: HasHints = ophyd.Signal(name='test')")
+
+
+def test_flyable(run):
     cmd = imports + "from bluesky.protocols import Flyable;"
     run(cmd + "foo: Flyable = hw.flyer1")
     run(cmd + "foo: Flyable = sim.TrivialFlyer()")
 
 
-def test_hinted():
```
NoxHarmonium, line 31: Replaced with `test_hashints` because `Hinted` doesn't exist anymore

ophyd/tests/test_bluesky_protocols.py
```diff
@@ -1,93 +1,157 @@
...
 
-def test_stoppable():
+def test_stoppable(run):
     cmd = imports + "from bluesky.protocols import Stoppable;"
     run(cmd + "foo: Stoppable = hw.motor1")
     run(cmd + "foo: Stoppable = hw.flyer1")
 
 
-def test_subscribable():
+# TODO: Ophyd signature is incompatible with bluesky protocol
```
NoxHarmonium, line 131: I have a feeling that Ophyd devices shouldn't conform to the `Subscribable` protocol as the concept of subscription is different between the two libraries.

The ophyd signature looks like this:
`def subscribe(self, callback, event_type=None, run=True) -> int`
and the Bluesky protocol looks like this:
`def subscribe(self, function: Callback[T]) -> None`

They seem to work differently in that Ophyd returns a `cid` when you subscribe which you can use as a token to unsubscribe. Bluesky doesn't give you a `cid` and unsubscribing works by passing the callback.

ophyd/device.py
```diff
@@ -1667,6 +1668,11 @@ def _repr_info(self):
         yield ("read_attrs", self.read_attrs)
         yield ("configuration_attrs", self.configuration_attrs)
 
+    # Placeholder required for devices to conform to the Bluesky Moveable protocol
+    # The actual implementation will be provided by sub classes
+    def set(self, *args, **kwargs) -> Any:
+        return super().set(*args, **kwargs)
```
NoxHarmonium, line 1674: These placeholder `set()` functions that just forward on to `super()` don't seem to break any tests so they seem OK but might need some extra scrutiny because they change the public API I suppose

evvaaaa: Why is this needed? If `set` is just calling `super`'s set, and it hasn't refined the return type at all, then what is it for? If we leave this out then`super()`'s `set` would be used instead anyway.

ophyd/sim.py
```diff
@@ -524,15 +527,15 @@ def position(self):
...
+    def hints(self) -> Hints:
         return {}
 
 
 class SynAxisNoHints(SynAxis):
     readback = Cpt(_ReadbackSignal, value=0.0, kind="omitted")
 
     @property
-    def hints(self):
+    def hints(self) -> Hints:
```
evvaaaa, line 538: If this raises an `AttributeError` then this should be type hinted `None`, or given no return type.

ophyd/signal.py
```diff
@@ -516,7 +519,7 @@ def value(self, value):
         self.put(value)
 
     @raise_if_disconnected
-    def read(self):
+    def read(self) -> OrderedDict[str, Any]:
```
evvaaaa, line 522: As before, we should use the bluesky protocols and `dict` here.

ophyd/sim.py
```diff
@@ -761,7 +764,7 @@ class TrivialFlyer:
     name = "trivial_flyer"
     parent = None
 
-    def kickoff(self):
+    def kickoff(self) -> Any:
```
evvaaaa, line 767:
```suggestion
    def kickoff(self) -> NullStatus:
```
?

ophyd/sim.py
```diff
@@ -902,9 +905,13 @@ def _scan(self):
...
                 self._data.append(event)
 
-    def stop(self, *, success=False):
+    def stop(self, success=False):
         pass
 
+    # Placeholder required for components to conform to the Bluesky Moveable protocol
+    def set(self, *args, **kwargs) -> Any:
+        return super().set(*args, **kwargs)
+
```
evvaaaa, line 914: Perhaps this should just pass as well? This is a sim so wouldn't have a `super()` anyway right?

evvaaaa: Correction, it should still return a `NullStatus`.

ophyd/status.py
```diff
@@ -416,7 +417,7 @@ def _finished(self, success=True, **kwargs):
             )
             self.set_exception(exc)
 
-    def exception(self, timeout=None):
+    def exception(self, timeout=None) -> Any:
```
evvaaaa, line 420:
```suggestion
    def exception(self, timeout=None) -> None | Exception:
```

ophyd/tests/test_bluesky_protocols.py
```diff
@@ -1,93 +1,156 @@
...
-    assert exit_status == 0
-
-
-def test_checkable():
+SupportedTypeCheckers = Literal["pyright", "mypy"]
+enabled_type_checkers: List[SupportedTypeCheckers] = ["pyright", "mypy"]
+
+
+@pytest.fixture(params=enabled_type_checkers, scope="session")
+def run(request) -> Callable[[str], None]:
```
evvaaaa, line 18: Not relevant to this PR, but it feels a little strange to have this test at all when we could just have a file which imports and defines these that a mypy CI run would pick up on.

ophyd/tests/test_bluesky_protocols.py
evvaaaa: Hmmm I don't know if the cid is useful here?

It's not like it would be accessible from the private [`_temp_callback_ids`](https://github.com/bluesky/bluesky/blob/5f5d2c1db8ee5c52fbc9a5f9e4fb3513b2d8625e/src/bluesky/run_engine.py#L753).

I'm not really sure what we need `event_type` for here either:

https://github.com/bluesky/bluesky/blob/5f5d2c1db8ee5c52fbc9a5f9e4fb3513b2d8625e/src/bluesky/suspenders.py#L67-L68

@tacaswell

ophyd/tests/test_bluesky_protocols.py
```diff
@@ -1,93 +1,156 @@
...
     cmd = imports + "from bluesky.protocols import Subscribable;"
-    run(cmd + "foo: Subscribable = hw.signal1")
+    run(cmd + "foo: Subscribable = hw.signal")
     run(cmd + "foo: Subscribable = ophyd.Signal(name='test')")
     run(cmd + "foo: Subscribable = ophyd.Device(name='test')")
     run(cmd + "foo: Subscribable = hw.motor1")
     run(cmd + "foo: Subscribable = hw.flyer1")
 
 
 if __name__ == "__main__":
```
evvaaaa, line 144: I think we can remove this `if __name__ ==  "__main__"` section.

ophyd/tests/test_bluesky_protocols.py
```diff
@@ -1,93 +1,156 @@
...
     cmd = imports + "from bluesky.protocols import Stoppable;"
     run(cmd + "foo: Stoppable = hw.motor1")
     run(cmd + "foo: Stoppable = hw.flyer1")
 
 
-def test_subscribable():
+# TODO: Ophyd signature is incompatible with bluesky protocol
+# (extra parameters, returns int instead of None, different parameter names)
+# Pyright is stricter and picks this up. Disabled for now
+@pytest.mark.skip()
```
evvaaaa, line 134:
```suggestion
# TODO: ...
@pytest.mark.skip(reason="Ophyd signature is incompatible with bluesky protocol")
```

ophyd/utils/types.py
```diff
@@ -0,0 +1,14 @@
+from typing import List, Literal, Tuple, TypedDict
+
+
+class Hints(TypedDict, total=False):
```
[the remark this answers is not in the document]
evvaaaa: Can we not use bluesky as a dependency and use the protocols like ophyd-async does?

@danielballan thoughts?

setup.cfg
```diff
@@ -57,6 +57,7 @@ dev =
     pytest-faulthandler
     pytest-rerunfailures
     pytest-timeout
+    pyright[nodejs]
```
evvaaaa, line 60: Why the `[nodejs]` version?

evvaaaa (changes requested):
I'm fairly sure we have to use `Dict[]` from typing instead of my suggested `dict[]` in python 3.8.

Also my suggestions on using the protocols themselves are dependent on if we decide to include bluesky as a dependency.

ophyd/tests/test_bluesky_protocols.py
NoxHarmonium: Now that I think about it, are the concepts even the same between Bluesky and Ophyd? In Ophyd a subscription is for watching for changes in device attributes, and in Bluesky a subscription is for handling events (e.g. with BestEffortCallback)

FWIW I know in our code base we have instances where we use `event_type` when subscribing to an Ophyd device to filter what events we want to receive and we store the CID to unsubscribe later.

ophyd/device.py
NoxHarmonium: Good point. It made the type checker happy, but now that I look again, I don't think the base classes even have an implementation of `set`. `Component` doesn't even have a base class.

I'll have to have another look at this. Maybe adding `set()` to `BlueskyInterface` would make more sense.

ophyd/device.py
```diff
@@ -562,7 +563,7 @@ def trigger(self) -> StatusBase:
         """
         pass
 
-    def read(self) -> OrderedDictType[str, Dict[str, Any]]:
+    def read(self) -> OrderedDict[str, Any]:
```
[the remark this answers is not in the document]
NoxHarmonium: It would be great if we could just make `bluesky` a dependency like `ophyd-async`. Then we could just use all the Bluesky types. I was assuming that we were trying to avoid coupling Ophyd to Bluesky but if not I'll re-do this PR. I suppose it might not be a breaking change as long as the version constraints were not too strict.

ophyd/sim.py
NoxHarmonium: I guess we could also use `NoReturn`. If the type checker is happy with that I'll make the change.

ophyd/sim.py
NoxHarmonium: I thought I had to change that to `Any` because the type checker didn't think that Ophyd's version of `BaseStatus` conformed to Bluesky's `Status` protocol but I'll double check

ophyd/tests/test_bluesky_protocols.py
NoxHarmonium: Very good point. I assumed there was a reason it was done that way and didn't want to change too much. Happy to replace this with a type checking step in the CI pipeline if that's the way people want to go.

setup.cfg
NoxHarmonium: Strangely enough, `pyright` is written in Javascript, I suppose because it's closely tied to pylance and VSCode.
Adding the `[nodejs]` feature just makes it install nodejs automatically using a wheel and is recommended (see https://pypi.org/project/pyright/)

> It's highly recommended to install pyright with the nodejs extra which uses [nodejs-wheel](https://pypi.org/project/nodejs-wheel-binaries/) to download Node.js binaries as it is more reliable than the default [nodeenv](https://pypi.org/project/nodeenv/) solution.

ophyd/tests/test_bluesky_protocols.py
evvaaaa: I suppose it's to mimick type checking downstream which could be using either.

ophyd/device.py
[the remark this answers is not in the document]
evvaaaa: We can wait and see. I suppose the difference is that bluesky doesn't require ophyd-async :thinking:

Issue #1251, opened 2025-06-12
BUG: pyepics 3.5.8 adds `monitor_delta` kwarg to `PV`, which `PyepicsShimPV` does not expect


tangkong:
Recently pyepics 3.5.8 added a `monitor_delta` kwarg to its `PV` class, which in and of itself is not a breaking change.
https://github.com/pyepics/pyepics/blob/4a4caebd92b75ee21c774cfc33c05ae77f4d1f97/epics/pv.py#L224

https://github.com/pyepics/pyepics/blob/4a4caebd92b75ee21c774cfc33c05ae77f4d1f97/epics/pv.py#L124-L129

However, ophyd patches over the `default_pv_class` and only expects some subset of the kwargs
https://github.com/bluesky/ophyd/blob/85e8d50f47235162e7035762e7d9859094f92e60/ophyd/_pyepics_shim.py#L53-L65

Error from our recent CI
```
>           thispv = default_pv_class(pvname, form=form, callback=callback,
                                      connection_callback=connection_callback,
                                      access_callback=access_callback,
                                      connection_timeout=timeout, count=count,
                                      verbose=verbose, auto_monitor=auto_monitor,
                                      monitor_delta=monitor_delta)
E           TypeError: __init__() got an unexpected keyword argument 'monitor_delta'
```

mrakitin:
Closed via #1253.

PR #1252, opened 2025-06-12
BUG: consume and ignore additional kwargs to PyepicsShimPV

tangkong:
# Summary
* New pyepics (3.5.8) adds support for a new kwarg in `PV`: `monitor_delta`
* ophyd patches this with a subclass `PyepicsShimPV` that consumes mostly the same kwargs, but very specifically those kwargs

Approach: consume any extra kwargs and drop them on the floor

# Context
See #1251 

It's possible we want to use this information, but I see that ophyd doesn't explicitly list pyepics as a core dependency.  The dev dependency itself pins pyepics to a lower version than this, but downstream consumers of ophyd likely don't install ophyd[dev] so they don't pick up this pin

### DIFF
--- a/ophyd/_pyepics_shim.py
+++ b/ophyd/_pyepics_shim.py
@@ -62,6 +62,7 @@ class PyepicsShimPV(epics.PV):
         connection_callback=None,
         connection_timeout=None,
         access_callback=None,
+        default_monitor=None,
     ):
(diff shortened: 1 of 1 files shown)


tangkong:
erm, apparently my brain broke and I used the wrong kwarg here.  I'll submit another PR

PR #1257, opened 2025-06-20
EpicsSignalNoValidation does not accept 'write_pv' as kwarg

### LINKED ISSUE #1256: EpicsSignalNoValidation does not accept 'write_pv' as kwarg
While evaluating the [`EpicsSignalNoValidation`](https://github.com/bluesky/ophyd/blob/f29fb219fa18fa8f5a8214d65302714d97c74a7f/ophyd/signal.py#L1679C7-L1679C30) class to solve a problem, it was found that the `EpicsSignalNoValidation` does not accept a `write_pv` kwarg, despite this comment in the source code: https://github.com/bluesky/ophyd/blob/f29fb219fa18fa8f5a8214d65302714d97c74a7f/ophyd/signal.py#L1686-L1689

Here's the error with ophyd 1.10.7

```bash
$ python -c "from ophyd.signal import EpicsSignalNoValidation; signal=EpicsSignalNoValidation('gp:gp:int1', name='signal', write_pv='gp:gp:int2'); signal.wait_for_connection(); print(f'{signal.read()=}')"
  File "<string>", line 1, in <module>
  File "/home/prjemian/.conda/envs/bluesky_2025_1/lib/python3.11/site-packages/ophyd/signal.py", line 1051, in __init__
    super().__init__(name=name, metadata=metadata, **kwargs)
TypeError: Signal.__init__() got an unexpected keyword argument 'write_pv'
```

@QZhang234 - Thanks for the contributions here!

prjemian:
- closes #1256
- #1044

@tangkong : Adding you as reviewer since you were the one committing EpicsSignalNoValidation to the repo ([as `EpicsSignalWO` in the commit message](https://github.com/bluesky/ophyd/commit/4fcc53cc3a8af3ebbdcb012f4f612b6eeb22e9c6)).
@mrakitin, @ZLLentz : Adding y'all since you reviewed #1044.

attn: @qzhang234

### DIFF
--- a/ophyd/signal.py
+++ b/ophyd/signal.py
@@ -1676,62 +1676,6 @@ class EpicsSignalRO(EpicsSignalBase):
             self._run_metadata_callbacks()
 
 
-class EpicsSignalNoValidation(EpicsSignal):
-    """An EpicsSignal that does not verify values on set.
-    This signal does support readback, but does not guarantee that
-    the readback will match the set value.
-
-    Keyword arguments are passed on to the base class (Signal) initializer
-
-    Parameters
-    ----------
-    read_pv : str
-        The PV to read from
-    write_pv : str, optional
-        The PV to write to if different from the read PV
-    name : str, optional
-        Name of signal.  If not given defaults to write_pv
-    write_timeout : float or None, optional
-        The timeout for a reply when put completion is used. This is
-        only applied if the PV is connected within connection_timeout (below).
-
-        This is very different than the connection and read timeouts
-        above. It relates to how long an action takes to complete, such motor
-        motion or data acquisition. Any default value we choose here is likely
-        to cause problems---either by being too short and giving up too early
-        on a lengthy action or being too long and delaying the report of a
-        failure. A finite value can be injected here or, perhaps more usefully,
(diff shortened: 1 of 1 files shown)


ZLLentz:
I'm a bit confused because, prior to this PR, `EpicsSignalNoValidation` seemingly doesn't work at all. Because it uses `EpicsSignalBase` as the parent class, `put` is actually `Signal.put`, e.g. no epics putting is happening.

Maybe we don't have any unit tests with real PVs for this?

The change looks strictly positive/correct to me.

I wonder if this class needs more work though, should the docstring clarify that this only skips creating a proper status object on `set`? The text in the class docstring implies it does more (I assumed we were skipping `check_value`, for example), so it should either do more or be clarified to note exactly what it does.

ZLLentz (approved):
Approving strictly positive improvement :+1:

tangkong:
We did have unit tests for this that attempted puts to a test ioc, but I guess we never verified the values were correct on the IOC?

tangkong (approved):
Good catch! I independently verified that `.set()` properly changes PVs.

mrakitin:
No objections from me, sorry for the delay!

PR #1263, opened 2025-09-25
Add custom timeout exception to be raised upon timeout in status.wait(timeout=2...)

cappel89:
# Summary
This PR adds the option to specify a custom exception to the .wait() method. 

## Use Case
If you want to raise a specific Exception upon running into a Timeout during the `.wait(timeout=...)` method, a try/except statement is needed. This PR adds the option to in addition specify a 'timeout_exception' to the `.wait(timeout=5, timeout_exception=...)` method of StatusBase. This allows the simpler syntax below, without try/except.

``` python
status = my_device.trigger()
status.wait(timeout=5, timeout_exception=TimeoutError(f"Device {self.name} timed out in trigger method.")
```

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -439,9 +439,7 @@ class StatusBase:
         return self._exception
 
     @tracer.start_as_current_span(f"{_TRACE_PREFIX} wait")
-    def wait(
-        self, timeout: float | None = None, timeout_exception: Exception | None = None
-    ):
+    def wait(self, timeout=None, timeout_exception=None):
         """
         Block until the action completes.
 


ophyd/status.py
```diff
@@ -465,9 +470,14 @@ def wait(self, timeout=None):
             indicates that the action itself raised ``TimeoutError``, distinct
             from ``WaitTimeoutError`` above.
         """
+        exception_to_raise = (
```
tacaswell, line 473: Creating the exception grabs things about the current stack, we should not create it until we need it.

tacaswell (changes requested):
I have concerns about passing in an instantiated exception object.

tacaswell:
I'm also not convinced by the use-case of avoiding try-except.  I would be much more sold on the ability to raise a custom sub-class that let you do better filtering in try-except.

I would also expect this to be set at Status init time not at wait time.  Devices have control over how the the status objects are created but no control over how they are waited on.

If the goal is to get better text in the error messages, then a better option would be to let `StatusBase` take a callable with the signature `Callable[(Self, float), str]` to format the error message, that way in trigger you could do:

```python
    return DeviceStatus(self, wait_msg_func=lambda sts, timeout: f'{sts.device.name} did not complete trigger in {timeout}s')
```

cappel89:
> I'm also not convinced by the use-case of avoiding try-except. I would be much more sold on the ability to raise a custom sub-class that let you do better filtering in try-except.
> 
> I would also expect this to be set at Status init time not at wait time. Devices have control over how the the status objects are created but no control over how they are waited on.
> 
> If the goal is to get better text in the error messages, then a better option would be to let `StatusBase` take a callable with the signature `Callable[(Self, float), str]` to format the error message, that way in trigger you could do:
> 
> ```python
>     return DeviceStatus(self, wait_msg_func=lambda sts, timeout: f'{sts.device.name} did not complete trigger in {timeout}s')
> ```

I agree with your statement that this should be moved to when the Status is created, and apply to any TimoutException raised by the status object. My initial thought was to make the PR as simple as possible, and therefore only adapt ``wait``, but I agree that this would have only partially covered the use-case that was targeted. 

The use-case in particular is that I create a status object with a pending timeout, or which will be awaited for later on in a different section of the code. If the status runs into a timeout exception, I would like to more control about what is raised. So in short (A) a descriptive error, and (B) a custom error message.
I can prepare a refactoring that allows for both in the initialization of StatusBase (i.e. timeout_exc_cls, timeout_exc_msg), but would like to confirm first that this would be a use case that you are in favor of merging.

PR #1264, opened 2025-09-25
Add option to DeviceStatus to skip "stop" call on failure

cappel89:
# Summary
This PR adds a kwarg (`call_stop_on_failure`) to DeviceStatus that allows to specify whether 'stop' should be called on the device upon failure of the DeviceStatus.

## Use Case
A DeviceStatus is useful as it carries an instance of the device object itself. This is convenient to simplify access to signals and methods of the device in callback methods. However, it would be nice to be able to have fine-grained control of whether the DeviceStatus calls 'stop' on failure as there may be additional logic governing whether this is desired or not. This PR simply adds an optional kwarg to prohibit 'stop' from being called upon failure

### DIFF
--- a/ophyd/status.py
+++ b/ophyd/status.py
@@ -681,6 +681,9 @@ class DeviceStatus(StatusBase):
 
     Parameters
     ----------
+    call_stop_on_failure: bool, optional
+        This option determines whether the device's ``stop()`` method is called
+        if the status fails. Default is True.
     timeout: float, optional
         The amount of time to wait before marking the Status as failed.  If
         ``None`` (default) wait forever. It is strongly encouraged to set a
@@ -691,12 +694,10 @@ class DeviceStatus(StatusBase):
         status has completed to running callbacks. Default is 0.
     """
 
-    def __init__(self, device, call_stop_on_failure: bool = True, **kwargs):
+    def __init__(self, device, *, call_stop_on_failure=True, **kwargs):
         self.device = device
         self._watchers = []
-        self._call_stop_on_failure = (
-            call_stop_on_failure if isinstance(call_stop_on_failure, bool) else True
-        )
+        self._call_stop_on_failure = bool(call_stop_on_failure)
         super().__init__(**kwargs)
         self._trace_attributes.update(
             {"device_name": device.name, "device_type": device.__class__.__name__}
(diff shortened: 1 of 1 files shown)


ophyd/status.py
```diff
@@ -691,9 +691,12 @@ class DeviceStatus(StatusBase):
         status has completed to running callbacks. Default is 0.
     """
 
-    def __init__(self, device, **kwargs):
+    def __init__(self, device, call_stop_on_failure: bool = True, **kwargs):
         self.device = device
         self._watchers = []
+        self._call_stop_on_failure = (
+            call_stop_on_failure if isinstance(call_stop_on_failure, bool) else True
```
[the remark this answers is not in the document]
tacaswell: I agree, either do no validation or at most cast through bool.

ophyd/status.py
```diff
@@ -691,9 +691,12 @@ class DeviceStatus(StatusBase):
         status has completed to running callbacks. Default is 0.
     """
 
-    def __init__(self, device, **kwargs):
+    def __init__(self, device, call_stop_on_failure: bool = True, **kwargs):
```
tacaswell, line 694: I don't think we have type annotations anywhere else in ophyd(-sync), I am not against adding them, but I think adding them one-at-a-time is not a good practice.

ophyd/status.py
```diff
@@ -705,7 +708,7 @@ def __init__(self, device, **kwargs):
     def _handle_failure(self):
         super()._handle_failure()
         self.log.debug("Trying to stop %s", repr(self.device))
-        if hasattr(self.device, "stop"):
+        if hasattr(self.device, "stop") and self._call_stop_on_failure:
```
tacaswell, line 711: micro-optimization:

```suggestion
        if self._call_stop_on_failure and hasattr(self.device, "stop"):
```

if we don't plan to use `stop`, don't bother doing the lookup.

ophyd/status.py
ZLLentz: I'm worried that if we don't add them one-at-a-time, we'll never get around to adding them at all.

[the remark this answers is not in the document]
cappel89: done

PR #1266, opened 2025-09-26
Add OrStatus, improve error propagation for AndStatus

cappel89:
# Summary
This PR adds an OrStatus objects and slightly improves error handling in the AndStatus. It was separated and refactored based on comments in #1265 (https://github.com/bluesky/ophyd/pull/1265#issuecomment-3340155606). 

## Use Case
In some cases, multiple status objects are created from methods or `set` calls. In some cases, it can be preferable to combine these status objects to hand them over to a different section in the code. This was already possible with the AndStatus object, however, the error handling in case of a failure was not optimal as the actual Exception leading to the status to fail was not properly propagated. This PR fixes this. In addition, a second option was added to combine status objects with a logical ``or``. Here, the idea is that it will resolve in success if any of the status objects succeeds, and only fails if a single status object fails.

## AndStatus
``` python
st1 = StatusBase()
st2 = StatusBase()
st3 = DeviceStatus(dev)
and_status = st1 & st2 & st3

and_status.done # is False
st1.set_exception(Exception("Test exception"))
and_status.done # is True
and_status.exception() # is Exception("Test exception")
```

## OrStatus
``` python
st1 = StatusBase()
st2 = StatusBase()
st3 = DeviceStatus(dev)
or_status = st1 | st2 | st3

st1.set_exception(Exception("Test exception"))
or_status.done # is False
st2.set_exception(TimeoutError("Test exception"))
or_status.done # is False
st3.set_exception(RuntimeError("Test exception"))
or_status.done # is True
or_status.exception() # is RuntimeError('RuntimeError: Exception: Test exception; TimeoutError: Test exception; RuntimeError: Test exception')
```


cappel89:
@tacaswell 
Thanks for the insides on the ``AndStatus``. By itself it 'almost' serves the purpose that we wanted to have. 
I used your input from the ``AndAllStatus``, and improved that the exception from sub-statuses is propagated to the AndStatus, which for me was missing in the current implementation. 
Besides that, I added in the same fashion the ``OrStatus``, and also added an __or__ method to StatusBase, which should now work with the same syntax. Please let me know your thoughts.

PR #1281, opened 2026-06-04
Handle case where data_type field is disabled for make_data_key

thopkins32:
Fixes problem with Pilatus IOC and others that disable `data_type` even though you can still read a value from this PV (the wrong value).

- [x] unit test
- [x] test with ADSim

### DIFF
--- a/ophyd/areadetector/detectors.py
+++ b/ophyd/areadetector/detectors.py
@@ -118,7 +118,7 @@ class DetectorBase(ADBase):
             external="FILESTORE:",
         )
 
-        # Some IOCs disable the data type, so we can't rely on its value
+        # If the data type isn't disabled, we assume it is accurate
         if self.cam.data_type_disabled.get() == 0:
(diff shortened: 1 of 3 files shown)


ophyd/areadetector/detectors.py
```diff
@@ -110,7 +110,12 @@ def make_data_key(self):
             self.cam.array_size.array_size_y.get(),
             self.cam.array_size.array_size_x.get(),
         )
-        dtype_numpy = np.dtype(self.cam.data_type.get(as_string=True).lower()).str
+        # Some IOCs disable the data type, so we can't rely on its value
```
tacaswell: We should not provide the key at all if the PV is not enabled.

thopkins32:
Works with ADSim: 
image

thopkins32:
One more: 

image

tacaswell (approved):

jwlodek:
LGTM

jwlodek (approved):

PR #1282, opened 2026-06-04
dtype_numpy property on DetectorBase, Fix Pilatus to int32 dtype

thopkins32:
Does two things:
- Enables easy override of `dtype_numpy` by simply overriding the property when inheriting from `DetectorBase`
- Forces Pilatus to use int32 dtype since this is always fixed at the IOC (and not reported correctly through `DataType_RBV`)

### DIFF
--- a/ophyd/areadetector/detectors.py
+++ b/ophyd/areadetector/detectors.py
@@ -6,6 +6,7 @@
 .. _areaDetector: https://areadetector.github.io/master/index.html
 """
 import warnings
+from typing import Optional
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


ophyd/areadetector/detectors.py
```diff
@@ -118,11 +128,9 @@ def make_data_key(self):
             external="FILESTORE:",
         )
 
-        # If the data type isn't disabled, we assume it is accurate
-        if self.cam.data_type_disabled.get() == 0:
-            ret["dtype_numpy"] = np.dtype(
-                self.cam.data_type.get(as_string=True).lower()
-            ).str
+        dtype_numpy = self.dtype_numpy
+        if dtype_numpy:
```
tacaswell:
```suggestion
        if dtype_numpy is not None:
```

tacaswell (approved):

jwlodek (approved):

PR #1288, opened 2026-08-20
Delay creation of temp directories until required

tpoliaw:
When the `SynSignalWithRegistry` is created without a save_path, it creates a temp directory instead. As this directory was eagerly created, and the `ophyd.sim` module creates an instance as part of the demo devices in the module (via the `globals().update`), simply importing `ophyd.sim` was enough to create a directory that would never be deleted.

As it's in `/tmp` by default, it's not a huge issue, but in some environments, eg running pytest across multiple processes using xdist, it can leave 20+ of these empty directories per test run which quickly add up.

The temp directory is now created lazily when the `save_path` is first used. It is still not deleted but this was presumably the original intention given there may be data in it.

### DIFF
--- a/ophyd/sim.py
+++ b/ophyd/sim.py
@@ -952,10 +952,7 @@ class SynSignalWithRegistry(SynSignal):
         self._resource_uid = None
         self._datum_counter = None
         self._asset_docs_cache = deque()
-        if save_path is None:
-            self.save_path = mkdtemp()
-        else:
-            self.save_path = save_path
+        self._save_path = save_path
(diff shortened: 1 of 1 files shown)


tacaswell (approved):

tacaswell:
Pushing the use of pytests temp directory fixture on the `hw` fixture is probably worth doing as a follow on PR.

tpoliaw:
The `hw` fixture in `conftest`? It looks like that already uses pytest's temp directory for the devices it creates, it's the `globals().update(hw().__dict__)` in the root of the `sim` module that creates its own that was the problem.

PR #1290, opened 2026-09-16
Make HDF5 plugin warmup signals configurable

### LINKED ISSUE #1280: HDF5Plugin.warmup() sets PVs that are disabled on some detectors downstream
On the Xspress3 community IOC, the corresponding PVs for `image_mode` and `acquire_period` are disabled. This can be fixed by overwriting the entire warmup() method in a class that inherits from `HDF5Plugin`, but a cleaner solution would move the signals and their warmup values to a property, so that downstream all that is needed is to set the property, and the warmup() method otherwise works as written.

CAOShurong:
## Summary

- expose the ordered camera-signal mapping used by `HDF5Plugin.warmup()`
- let downstream plugin subclasses omit unsupported or disabled camera PVs without replacing the complete warmup procedure
- retain the existing default signal order, values, acquisition delay, and restoration behavior

Fixes #1280.

## Verification

- pre-fix regression failed for both caproto and pyepics because warmup signals were not overridable; exact head passes 2/2 focused cases
- `ophyd/tests/test_areadetector.py` excluding four Windows-only path test groups: 3,554 passed / 320 skipped; the 44 excluded failures reproduce unchanged on exact base `114a924`
- changed-file pre-commit: large-file, merge-conflict, Black, and Flake8 hooks passed
- sdist and wheel build passed at exact head `881879a`
- isolated install of the built wheel passed the downstream-subclass warmup mapping probe

## AI assistance

OpenAI Codex assisted with implementation and testing. The submitted diff and exact-head verification results were reviewed in the contribution workflow before submission.


ophyd/areadetector/plugins.py
```diff
@@ -973,26 +973,31 @@ class HDF5Plugin(FilePlugin, version=(1, 9, 1), version_type="ADCore"):
...
+    def warmup_signals(self):
+        """Ordered mapping of signals to values used while warming up."""
+        return OrderedDict(
             [
                 (self.parent.cam.array_callbacks, 1),
                 (self.parent.cam.image_mode, "Single"),
                 (self.parent.cam.trigger_mode, "Internal"),
-                # just in case tha acquisition time is set very long...
+                # just in case the acquisition time is set very long...
                 (self.parent.cam.acquire_time, 1),
```
prjemian: wrong line for this: Could this common step be moved into the `warmup()` method?

ophyd/areadetector/plugins.py
```diff
@@ -973,26 +973,31 @@ class HDF5Plugin(FilePlugin, version=(1, 9, 1), version_type="ADCore"):
...
+    def warmup(self):
+        """
+        A convenience method for 'priming' the plugin.
+
+        The plugin has to 'see' one acquisition before it is ready to capture.
+        This sets the array size, etc.
+        """
+        self.enable.set(1).wait()
+        sigs = self.warmup_signals
+
```
prjemian: common step (start acquire) could move here

ophyd/areadetector/plugins.py
prjemian: This line was intended:

```python
                (self.parent.cam.acquire_time, 1),
```

